PR #79, opened 2014-11-11
Un-advertise "dx import" and "dx new gtable"

godotgildor:
...vertised.  They still work, just won't be advertised through help or command line completion.


psung:
Thanks @godotgildor!

Note that I took the liberty of rewording your squashed commit message in d8408eba (in general when you use `git merge --squash` you'll want to synthesize a new commit message for the change as a whole, rather than be left with all the constituent commit messages).

As a side effect of my rewriting away your original commit, you might need to `git fetch; git reset --hard origin/master` the next time you return to your dx-toolkit checkout.

PR #83, opened 2014-12-17
Adding close option to new record

evanworley:
- Updating new record documentation with --close option
- Adding test for creating a new record with and without --close

### DIFF
--- a/src/python/dxpy/bindings/dxrecord.py
+++ b/src/python/dxpy/bindings/dxrecord.py
@@ -82,6 +82,8 @@ class DXRecord(DXDataObject):
         :type dx_hash: dict
         :param init_from: Record from which to initialize the metadata
         :type init_from: :class:`DXRecord`
+        :param close: Whether or not to close the record immediately after creating it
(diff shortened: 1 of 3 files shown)


src/python/dxpy/bindings/dxrecord.py
```diff
@@ -82,6 +82,8 @@ def _new(self, dx_hash, **kwargs):
         :type dx_hash: dict
         :param init_from: Record from which to initialize the metadata
         :type init_from: :class:`DXRecord`
+        :param close: Whether or not to close the record immediately after creating it
```
psung: I think it's more appropriate for this to be explicitly named as a keyword arg to the `_new` method. The `media_type` argument for `DXFile` object might be a useful precedent here, since it's an option specific to files, just as `close` is an option specific to records.

evanworley: Sure, fixed. Is there anything else?

psung:
LGTM, please squash merge to master. Thanks!

PR #84, opened 2014-12-24
DXHTTPRequest retry loop: obey timeout

kislyuk:
@psung, please take a look.

### DIFF
--- a/src/python/dxpy/__init__.py
+++ b/src/python/dxpy/__init__.py
@@ -287,9 +287,8 @@ def DXHTTPRequest(resource, data, method='POST', headers=None, auth=True, timeou
     if hasattr(data, 'seek') and hasattr(data, 'tell'):
         rewind_input_buffer_offset = data.tell()
 
-    time_started, last_exc_type, last_error, last_traceback = None, None, None, None
-    if timeout:
-        time_started = time.time()
+    last_exc_type, last_error, last_traceback = None, None, None
+    time_started = time.time() if timeout else None
     try_index = 0
     while True:
         success, streaming_response_truncated = True, False
--- a/src/python/dxpy/__init__.py
+++ b/src/python/dxpy/__init__.py
@@ -188,7 +188,7 @@ def _process_method_url_headers(method, url, headers):
     else:
         return method, _url, _headers
 
-def DXHTTPRequest(resource, data, method='POST', headers=None, auth=True, timeout=600,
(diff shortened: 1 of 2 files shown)


src/python/dxpy/__init__.py
```diff
@@ -288,13 +288,15 @@ def DXHTTPRequest(resource, data, method='POST', headers=None, auth=True, timeou
...
     last_exc_type, last_error, last_traceback = None, None, None
+    time_started = time.time() if timeout else None
     try_index = 0
     while True:
         success, streaming_response_truncated = True, False
         response = None
         try:
             _method, _url, _headers = _process_method_url_headers(method, url, headers)
-            response = session_handler.request(_method, _url, headers=_headers, data=data, timeout=timeout, auth=auth,
+            _timeout = timeout or 600
```
kislyuk: IMO it's better to preserve the simpler semantics of passing the unmodified timeout value to requests (which then sets it on the socket) than it is to change it to ensure that we spend no more than `timeout` before returning (which is not a guarantee that the caller can fully rely on anyway, since it's not anything like a real-time system). The scenario of the first few requests returning quickly and the last one hanging (which is necessary to trigger the scenario of significantly missing the timeout deadline before returning) seems far-fetched to me. I agree there is a minor correctness issue, but the two applications of the timeout value really deal with different issues: the one passed to requests and set on the socket takes care of stalled out requests, while the one checked in the retry loop takes care of breaking out of the 503 retry-later mode.

(We may also want to investigate using urllib3's retry support, which was introduced over the course of 2014. Maybe it better deals with this socket/total timeout duality?)

psung: Ok, that sounds reasonable to me. Thanks!

psung:
LGTM

kislyuk:
@psung, when preparing to merge this with your comment in mind, I did realize that if the server specifies a retry after interval much greater than our timeout, we can oversleep by a lot. I've added a condition for that, please take another look.

psung:
Thanks. As you said before, maybe it's better just to represent `timeout` to the caller as the socket timeout that will be applied to each attempt and not as any kind of bound on the total time. So I could go either way on this last change, but I think it's fine.

kislyuk:
Thanks for the review, merged in https://github.com/dnanexus/dx-toolkit/commit/5931eb4.

PR #86, opened 2014-12-31
PTFM-13951 Retry upon connection error even for non-idempotent requests

psung:
@kislyuk, care to review?

### DIFF
--- a/src/python/dxpy/__init__.py
+++ b/src/python/dxpy/__init__.py
@@ -126,6 +126,7 @@ environment variables:
 from __future__ import (print_function, unicode_literals)
 
 import os, sys, json, time, logging, platform, collections, ssl, traceback
+import errno
 import requests
 import socket
 
@@ -191,6 +192,9 @@ def _process_method_url_headers(method, url, headers):
         return method, _url, _headers
 
 
+_RETRYABLE_SOCKET_ERRORS = set([errno.ENETDOWN, errno.ENETUNREACH, errno.ECONNREFUSED])
+
+
 def _is_retryable_exception(e):
     """Returns True if the exception is always safe to retry.
 
@@ -202,16 +206,16 @@ def _is_retryable_exception(e):
     have been established, we return False.
 
     """
-    # TODO: what other errors can be safely retried?
     try:
-        # TODO: there seems to be some issue with checking isinstance(e,
-        # requests.exceptions.ConnectionError) -- the type of e reports
-        # ConnectionError but it doesn't match; could there be two
-        # classes of that name??
-        #
-        # Unfortunately requests doesn't seem to provide a sensible API
-        # to retrieve the cause
-        return isinstance(e.args[0].args[1], (socket.gaierror, socket.herror))
+        if isinstance(e, requests.exceptions.ConnectionError):
+            # Unfortunately requests doesn't seem to provide a sensible
+            # API to retrieve the cause
+            cause = e.args[0].args[1]
+            if isinstance(cause, (socket.gaierror, socket.herror)):
+                return True
+            if isinstance(cause, socket.error) and cause.errno in _RETRYABLE_SOCKET_ERRORS:
+                return True
+        return False
     except (AttributeError, TypeError, IndexError):
         return False
 
--- a/src/python/dxpy/__init__.py
+++ b/src/python/dxpy/__init__.py
@@ -192,7 +192,14 @@ def _process_method_url_headers(method, url, headers):
         return method, _url, _headers
 
 
-_RETRYABLE_SOCKET_ERRORS = set([errno.ENETDOWN, errno.ENETUNREACH, errno.ECONNREFUSED])
+# When any of the following errors are indicated, we are sure that the
+# server never received our request and therefore the request can be
+# retried (even if the request is not idempotent).
+_RETRYABLE_SOCKET_ERRORS = {
+    errno.ENETDOWN,     # The network was down
+    errno.ENETUNREACH,  # The subnet containing the remote host was unreachable
+    errno.ECONNREFUSED  # A remote host refused to allow the network connection
+}
 
 
 def _is_retryable_exception(e):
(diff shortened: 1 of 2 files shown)


src/python/dxpy/__init__.py
```diff
@@ -188,6 +191,35 @@ def _process_method_url_headers(method, url, headers):
...
+    This is True if the client was never able to establish a connection
+    to the server (for example, name resolution failed or the connection
+    could otherwise not be initialized).
+
+    Conservatively, if we can't tell whether a network connection could
+    have been established, we return False.
+
+    """
+    try:
+        if isinstance(e, requests.exceptions.ConnectionError):
```
kislyuk: The name `ConnectionError` is already imported, perhaps use it here?

src/python/dxpy/__init__.py
```diff
@@ -188,6 +191,35 @@ def _process_method_url_headers(method, url, headers):
...
+            # Unfortunately requests doesn't seem to provide a sensible
+            # API to retrieve the cause
+            cause = e.args[0].args[1]
```
kislyuk: In Python 3, you can chain exceptions with `raise e from original`, providing a natural interface for this. Python 2 sadness :(

src/python/test/test_dxpy.py
```diff
@@ -1877,6 +1877,23 @@ def test_retry_after_without_header_set(self):
         self.assertTrue(50000 <= time_elapsed)
         self.assertTrue(time_elapsed <= 70000)
 
+    @unittest.skip("Example of a request that should be retried")
```
kislyuk: I assume this is under a skip clause because it takes too long? I wonder if we could instrument the code and tests to control the sleep delay between retries and assert on the number of retries made.

src/python/test/test_dxpy.py
```diff
@@ -1877,6 +1877,23 @@ def test_retry_after_without_header_set(self):
...
+        dxpy.DXHTTPRequest('http://doesnotresolve.dnanexus.com/', {}, prepend_srv=False, always_retry=False)
+
+    @unittest.skip("Example of a request that should be retried")
+    def test_connection_refused(self):
+        # This call shouldn't return immediately, rather it should be
+        # retried with the entire retry loop taking approximately 1
+        # minute.
+        #
+        # Connecting to a port on which there is no server running
+        dxpy.DXHTTPRequest('http://localhost:20406', {}, prepend_srv=False, always_retry=False)
```
kislyuk: Something to consider, but not necessarily do anything about, is that the new behavior of retrying on name resolution or connection refused errors may be annoying in interactive use when dx is misconfigured. Perhaps it's worth it to make it configurable or even auto-configure it not to retry if sys.stdin.isatty() (but I'm not convinced of that need yet).

src/python/dxpy/__init__.py
```diff
@@ -188,6 +191,35 @@ def _process_method_url_headers(method, url, headers):
     else:
         return method, _url, _headers
 
+
+_RETRYABLE_SOCKET_ERRORS = set([errno.ENETDOWN, errno.ENETUNREACH, errno.ECONNREFUSED])
```
psung: Thanks, done. I documented each class of error, too.

src/python/dxpy/__init__.py
psung: Good point, done.

src/python/dxpy/__init__.py
psung: Indeed. But even without the special syntax, is there no Python convention to retrieve the original exception? I guess what I had in mind was like this in Java: http://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#getCause()

src/python/test/test_dxpy.py
psung: Yes, it does take too long, but I also had no way to assert that retries were being made (apart from checking the time taken, which seemed lame to me).

Perhaps part of the way to test this is to generate the exceptions and then add a unit test on `_is_retryable_exception(e)`. I like your idea of controlling the sleep delay in tests too-- actually sleeping to wait for time to elapse in tests doesn't really sit well with me.

src/python/test/test_dxpy.py
psung: Indeed, one way out may also be to keep the retry behavior the same but to print name resolution failures to stderr.

src/python/test/test_dxpy.py
kislyuk: We could try adding the number of retries attempted as an attribute of the exception object thrown. That lets us assert on that value in tests.

src/python/dxpy/__init__.py
kislyuk: There is no convention on how to introspect exception details :( I usually cast exceptions to strings and do `if pattern in str(e): ...` but it's no improvement over what you did.

src/python/test/test_dxpy.py
psung: I realized I can't modify the exception being thrown without resetting the call stack to that of the new throw site, and I didn't really want to go down that road (this relates to both the recent change to preserve the call stack properly, and also our other discussion about if there were a more useful way to chain exceptions).

In order to fix this test I've started catching the exception that's raised and checking that it passes `dxpy._is_retryable_exception` (but leaving the exception untouched in `DXHTTPRequest`).

src/python/test/test_dxpy.py
psung: I'll keep this in mind but I'm not planning to change the behavior at this time to handle this scenario differently.

src/python/test/test_dxpy.py
kislyuk: Nice!

kislyuk:
LGTM

psung:
Squash merged as e5b3f7a

psung:
Thanks @kislyuk for the review!

PR #87, opened 2015-01-28
PTFM-13654 dxpy config manager

kislyuk:
@psung, please review.

* Unify dxpy configuration management
* Save and load entire configuration at once, never piecemeal
* Simplify and deduplicate code in some places
* Begin test suite


src/python/dxpy/utils/config.py
```diff
@@ -0,0 +1,297 @@
...
+                dxpy.set_workspace_id(environ["DX_PROJECT_CONTEXT_ID"])
+                dxpy.set_project_context(environ["DX_PROJECT_CONTEXT_ID"])
+
+    def get_global_conf_dir(self):
+        return self._global_conf_dir
+
+    def get_user_conf_dir(self):
+        return self._user_conf_dir
+
+    def get_session_conf_dir(self, cleanup=True):
```
psung: Perhaps `cleanup` should default to False and only be set to True when initializing. It seems superfluous to attempt the cleanup when saving and clearing the config, too.

src/python/dxpy/utils/config.py
```diff
@@ -0,0 +1,297 @@
...
+        if key in self.CORE_VAR_NAMES:
+            self._sync_dxpy_state()
+
+    def __iter__(self):
+        for item in self.VAR_NAMES:
+            if item in environ:
+                yield item
+
+    def __len__(self):
+        return len(self.VAR_NAMES)
```
psung: The length reported here ought to be consistent with the number of items returned by the iterator.

src/python/dxpy/utils/config.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def _write_conf_dir(self, conf_dir):
+        try:
+            os.makedirs(conf_dir, 0o700)
+        except OSError:
+            os.chmod(conf_dir, 0o700)
+
+        env_jsonfile_path = os.path.join(conf_dir, "environment.json")
+        # Make sure the file has 600 permissions
+        with _open_for_writing_with_permissions(env_jsonfile_path, 0o600) as fd:
```
psung: An alternative strategy might be to write to a separate file and (atomically) rename it to `environment.json`, but maybe that's overkill for now.

src/python/dxpy/utils/config.py
```diff
@@ -0,0 +1,297 @@
...
+            json.dump({k: self.get(k, self.defaults.get(k)) for k in self}, fd, indent=4)
+            fd.write("\n")
+
+        for var in self.STANDALONE_VAR_NAMES:
+            # Make sure the file has 600 permissions
+            with _open_for_writing_with_permissions(os.path.join(conf_dir, var), 0o600) as fd:
+                value = self.get(var, self.defaults.get(var, ""))
+                fd.write(value.encode(sys_encoding) if USING_PYTHON2 else value)
+
+        if not os.path.exists(os.path.join(self._user_conf_dir, "unsetenv")):
```
psung: I feel that this should be factored out into a separate function, perhaps to be called just once in `save`. It's independent of the values in the configuration, and it shouldn't be written twice (once for each call to `_write_conf_dir`.

src/python/dxpy/scripts/dx.py
```diff
@@ -3446,7 +3438,7 @@ def register_subparser(subparser, subparsers_action=None, categories=('other', )
                                         description='Clears all environment variables set by dx.  More specifically, it removes local state stored in ~/.dnanexus_config/environment.  Does not affect the environment variables currently set in your shell.', prog='dx clearenv')
 parser_clearenv.add_argument('--reset', help='Reset dx environment variables to empty values. Use this to avoid interference between multiple dx sessions when using shell environment variables.',
                              action='store_true')
-parser_clearenv.set_defaults(func=clearenv, interactive=True)
+parser_clearenv.set_defaults(func=clearenv, interactive=False)
```
psung: Why is this needed?

src/python/dxpy/utils/config.py
```diff
@@ -0,0 +1,297 @@
+# Copyright (C) 2013-2014 DNAnexus, Inc.
```
psung: 2014 => 2015

psung:
Thanks for taking care of this! Mostly it looks good to me, please see the minor suggestions in the line comments.

I merged latest master and resolved the merge conflicts, and I'm running the integration tests against this branch now.

src/python/dxpy/scripts/dx.py
kislyuk: Previously, `dx clearenv` would invoke a hardcoded shortcut that would avoid initializing dxpy from the environment. The invocation through the argparse parser would only happen when running `dx sh`. Now we can use the parser for both.

src/python/dxpy/scripts/dx.py
```diff
@@ -427,27 +414,24 @@ def logout(args):
...
         os.environ['DX_PROJECT_CONTEXT_ID'] = project
         if name is not None:
-            set_env_var("DX_PROJECT_CONTEXT_NAME", name)
+            dxpy.config["DX_PROJECT_CONTEXT_NAME"] = name
         if write:
-            write_env_var("DX_PROJECT_CONTEXT_ID", project)
+            dxpy.config.write("DX_PROJECT_CONTEXT_ID", project)
             if name is not None:
-                write_env_var("DX_PROJECT_CONTEXT_NAME", name)
+                dxpy.config.write("DX_PROJECT_CONTEXT_NAME", name)
```
psung: Previously these write operations were mostly writing different files, but now they write all the files (resulting in each config file being written multiple times). Consider batching the writes with a single "save" call at the end.

src/python/dxpy/utils/config.py
kislyuk: Good point, fixed.

src/python/dxpy/utils/config.py
kislyuk: Fixed, thanks.

src/python/dxpy/utils/config.py
kislyuk: I'd rather add that later along with lazy loading.

src/python/dxpy/utils/config.py
kislyuk: Sounds good to me. Refactored.

src/python/dxpy/utils/config.py
kislyuk: Fixed.

src/python/dxpy/utils/config.py
psung: SGTM.

src/python/dxpy/scripts/dx.py
kislyuk: Thanks for pointing this out. I ended up refactoring this block since the logic is simpler now.

src/python/dxpy/scripts/dx.py
```diff
@@ -427,42 +414,29 @@ def logout(args):
...
-        os.environ['DX_PROJECT_CONTEXT_ID'] = project
-        if name is not None:
-            set_env_var("DX_PROJECT_CONTEXT_NAME", name)
-        if write:
-            write_env_var("DX_PROJECT_CONTEXT_ID", project)
-            if name is not None:
-                write_env_var("DX_PROJECT_CONTEXT_NAME", name)
-            else:
-                try:
-                    os.remove(os.path.expanduser('~/.dnanexus_config/DX_PROJECT_CONTEXT_NAME'))
```
psung: Thanks for this refactoring, this is much better. Possibly minor point, is it important to distinguish between `DX_PROJECT_CONTEXT_NAME` being None and ""? Before you would remove the corresponding file if the value was set to None, but this is no longer done (with good reason, I think).

psung:
Thanks. Did you ever figure out why the branch tests were failing? Does it have to do with 2669032?

kislyuk:
No. I'm still trying to figure that out. The tests in question pass in isolation. It seems a previous test messes up the environment, or maybe even worse, environment leaks in from another concurrent Jenkins job.

kislyuk:
FYI, https://github.com/dnanexus/dx-toolkit/commit/26690325522b23f1171a5aa313cdd344c9c45c98 is unrelated - it's a fix for a funny problem that I observed on a customer's shell... `LC_CTYPE=UTF-8 dx`

kislyuk:
@psung, after digging around, I found the issue (dxpy.JOB_ID was not being reset). The branchtest passes now. I'm ready to merge.

src/python/dxpy/scripts/dx.py
kislyuk: They should be treated the same, but you can't set an environment variable to None. Are there any practical differences that you see needed with that distinction in mind?

psung: None that I can foresee; I looked through its callers just now too.

src/python/dxpy/utils/config.py
```diff
@@ -0,0 +1,297 @@
...
+            dxpy.set_security_context(json.loads(environ["DX_SECURITY_CONTEXT"]))
+
+        if "DX_JOB_ID" in environ:
+            dxpy.set_job_id(environ["DX_JOB_ID"])
+            dxpy.set_workspace_id(environ.get("DX_WORKSPACE_ID"))
+            dxpy.set_project_context(environ.get("DX_PROJECT_CONTEXT_ID"))
+        elif "DX_PROJECT_CONTEXT_ID" in environ:
+            dxpy.set_job_id(None)
+            dxpy.set_workspace_id(environ["DX_PROJECT_CONTEXT_ID"])
+            dxpy.set_project_context(environ["DX_PROJECT_CONTEXT_ID"])
```
psung: Should we clear these variables if neither DX_JOB_ID nor DX_PROJECT_CONTEXT_ID is set?

psung:
Thanks, nice work! LGTM but please see the one minor suggestion.

src/python/dxpy/utils/config.py
kislyuk: Good point, how about now?

psung:
LGTM, thanks!

kislyuk:
Merged in https://github.com/dnanexus/dx-toolkit/commit/952bdf0e146bfbd2bce2c2b00acfec47bca0e304, thanks for the review.

PR #88, opened 2015-03-04
PTFM-14387 Move job log message truncation

kislyuk:
The truncation previously resided in the dx-log-stream wrapper, which
ingests stdout and stderr, but data sent directly to the log socket
bypasses it. The wrapper invokes the logger function, which is
responsible for forwarding the message. The logger function is used in
Python-based apps. This change eliminates confusing JSON parsing
errors when Python-based apps send log lines that are too long.

@psung please review

### DIFF
--- a/src/python/dxpy/dxlog.py
+++ b/src/python/dxpy/dxlog.py
@@ -70,8 +70,19 @@ class DXLogHandler(SysLogHandler):
 
     def emit(self, record):
         level = self.encodePriority(record)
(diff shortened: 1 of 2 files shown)


psung:
LGTM

kislyuk:
Pushed to master in https://github.com/dnanexus/dx-toolkit/commit/98d35a368b3632b9ddf75d131acc1df29535ef70, thanks for the review.

PR #91, opened 2015-04-07
PTFM-14841 Invoke download URL refresh logic in every iteration

kislyuk:
The download URL cache and refresh logic is already there, but the
file download request iterator was storing a static copy of the
download URL instead of running the cache/refresh helper.

@psung, please review


psung:
LGTM

kislyuk:
Thanks for the review. In 4724f44

PR #92, opened 2015-04-08
PTFM-14733 Log resource usage through priority socket

kislyuk:
This is a quick and dirty way to make the resource usage log lines bypass the 2 MB job log limit and continue to be sent even when the bulk logs have been muted.

@psung please review


psung:
LGTM

kislyuk:
In https://github.com/dnanexus/dx-toolkit/commit/072c3b57e546d326dc98ed49b4bdee1f3e582fb5

PR #93, opened 2015-04-09
PTFM-13969 reuse bundled depends

psung:
@kislyuk, do you mind taking a look? I changed the hash method as we discussed, and fixed a couple of other miscellaneous bugs in "dx build".

### DIFF
--- a/src/python/dxpy/app_builder.py
+++ b/src/python/dxpy/app_builder.py
@@ -38,6 +38,8 @@ from __future__ import (print_function, unicode_literals)
 
 import os, sys, json, subprocess, tempfile, multiprocessing
 import datetime
+import hashlib
+
 import dxpy
 from . import logger
 from .utils import merge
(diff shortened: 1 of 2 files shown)


src/python/dxpy/app_builder.py
```diff
@@ -150,20 +203,54 @@ def upload_resources(src_dir, project=None, folder='/'):
...
+            project=dest_project,
+            folder=target_folder,
+            properties=dict(resource_bundle_checksum=directory_checksum),
+            visibility='either',
+            zero_ok=True,
+            return_handler=True
+        )
+
+        if existing_resources:
+            logger.debug("Found existing resource bundle that matches local resources directory: " +
```
kislyuk: We now have logging configured to print only INFO and higher priority messages, so for me, these messages don't show up with any of `dx build -a`, `_DX_DEBUG=1 dx build -a`, or `_DX_DEBUG=2 dx build -a`. Is that intentional? It may be best to also have these messages use logger.info, and also to set logging level to debug if `_DX_DEBUG` is set (though that's outside the scope of this PR).

kislyuk:
Thanks, this looks great.

LGTM modulo the logging issue.

psung:
Thanks! I changed that log line to use `logger.info`.

PR #94, opened 2015-04-16
Check chunk complete

godotgildor:
This is my proposed change to ua so that after all upload threads are complete, it queries the status of each chunk that it has uploaded and checks that it is marked as complete.  If it finds a chunk not complete, it attempts to re-upload it.

I have taken a few largeish files (100 MB and 400 MB) and uploaded them using a variety of chunk sizes (5M, 10M, 75M) and the uploads complete without issue and after redownloading the files they have the proper md5sum.  I have not been able to verify if a chunk is not complete after the upload is done that the retry logic does execute since I'm not sure how to get a chunk to be in a not complete state.

### DIFF
--- a/src/ua/main.cpp
+++ b/src/ua/main.cpp
@@ -103,6 +103,11 @@ int NUMTRIES_g; // Number of max tries for a chunk (to be given by user)
 
 string userAgentString; // definition (declared in chunk.h)
 
+// Max number of times to check if a chunk is complete.
+int NUM_CHUNK_CHECKS = 3;
+// Time in milliseconds between checks to see if a chunk is complete.
+int CHUNK_CHECK_SLEEP = 1000;
+
 bool finished() {
   return (chunksFinished.size() + chunksFailed.size() == totalChunks);
 }
@@ -184,6 +189,17 @@ void compressChunks() {
   }
 }
 
+bool is_chunk_complete(Chunk *c) {
+    bool chunk_complete = false;
+    JSON result = fileDescribe(c->fileID);
+    string partIndex = boost::lexical_cast<string>(c->index + 1); // minimum part index is 1
+
+    if (result["parts"].has(partIndex) && result["parts"][partIndex]["state"].get<string>() == "complete")
+        chunk_complete = true;
+
+    return chunk_complete;
+}
+
 void uploadChunks(vector<File> &files) {
   try {
     while (true) {
@@ -194,7 +210,13 @@ void uploadChunks(vector<File> &files) {
       bool uploaded = false;
       try {
         c->upload(opt);
-        uploaded = true;
+
+        for(int currRetry=0; currRetry < NUM_CHUNK_CHECKS; ++currRetry) {
+            if(is_chunk_complete(c))
+                uploaded = true;
+                break;
+            boost::this_thread::sleep(boost::posix_time::milliseconds(CHUNK_CHECK_SLEEP));
+        }
       } catch (runtime_error &e) {
         ostringstream msg;
         msg << "Upload failed: " << e.what();
@@ -679,6 +701,8 @@ int main(int argc, char * argv[]) {
       if (files[i].failed) {
         cerr << "File \""<< files[i].localFile << "\" could not be uploaded." << endl;
       } else {
+        // @TODO Insert code to check if chunks are pending or complete.
+        // @TODO Try to reupload any pending chunks.
         cerr << "File \"" << files[i].localFile << "\" was uploaded successfully. Closing..." << endl;
         if (files[i].isRemoteFileOpen) {
           files[i].close();
--- a/src/ua/main.cpp
+++ b/src/ua/main.cpp
@@ -555,48 +555,31 @@ void check_for_complete_chunks(vector<File> &files) {
       if (fileDescriptions.find(c->fileID) == fileDescriptions.end())
         fileDescriptions[c->fileID] = fileDescribe(c->fileID);
 
-      // Note that currently we are re-reading the chunk data and perhaps
-      // compressing it in the main thread.  Ideally we'd do this in separate
-      // threads like the original loop.  However, this should be a rare
-      // event, so we'll stick to this simple implementation for now.
       if (!is_chunk_complete(c, fileDescriptions[c->fileID])) {
         // After the chunk was uploaded, it was cleared, removing the data
         // from the buffer.  We need to reload if we're going to upload again.
-        c->read();
-        if (c->toCompress)
-          c->compress();
-        chunksToUpload.produce(c);
+        chunksToRead.produce(c);
       }
     }
     // All of the chunks were marked as complete, so let's exit and we
     // should be safeish to close the file.
-    if(chunksToUpload.size() == 0)
+    if(chunksToRead.size() == 0)
       return;
 
     // Set the totalChunks variable to the # of chunks we're going
-    // to retry now.
-    totalChunks = chunksToUpload.size();
-    // Upload the chunks which weren't marked as complete.
-    DXLOG(logINFO) << " upload...";
-    for (int i = 0; i < opt.uploadThreads; ++i) {
-      uploadThreads.push_back(boost::thread(uploadChunks, boost::ref(files)));
-    }
-    // The upload threads don't exit once the upload is complete.  Instead they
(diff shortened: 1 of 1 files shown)


src/ua/main.cpp
```diff
@@ -184,6 +187,17 @@ void compressChunks() {
   }
 }
 
+bool is_chunk_complete(Chunk *c) {
+    bool chunk_complete = false;
+    JSON result = fileDescribe(c->fileID);
+    string partIndex = boost::lexical_cast<string>(c->index + 1); // minimum part index is 1
```
dkillebrew: Are you sure that index+1 is the right thing to do here?

src/ua/main.cpp
```diff
@@ -184,6 +187,17 @@ void compressChunks() {
...
+
+    if (result["parts"].has(partIndex) && result["parts"][partIndex]["state"].get<string>() == "complete")
```
dkillebrew: I think this could be more simply expressed:

`return result["parts"].has(partIndex) && result["parts"][partIndex]["state"].get<string>() == "complete"`

src/ua/main.cpp
```diff
@@ -529,6 +543,39 @@ void printEnvironmentInfo() {
...
+// upload timed-out.  It's possible that a second upload succeeds,
+// has the chunk marked as "complete" and then the first request makes
+// its way through the queue and marks the chunk as pending again.
+// Since we are just about to close the file, we'll check to see if any
+// chunks are marked as pending, and if so, we'll retry them.
+void check_for_complete_chunks(vector<File> &files) {
+  for (int currCheckNum=0; currCheckNum < NUM_CHUNK_CHECKS; ++currCheckNum){
+    while (!chunksFinished.empty()) {
+      Chunk *c = chunksFinished.consume();
+      if (!is_chunk_complete(c)) {
```
dkillebrew: When there are thousands of chunks, isn't this inefficient? `is_chunk_complete` causes a `fileDescribe`. It's more efficient to describe each file once. What about:

```
describe all files
add any pending chunks to chunksToUpload
```

src/ua/main.cpp
```diff
@@ -529,6 +543,39 @@ void printEnvironmentInfo() {
...
+          chunksToUpload.produce(c);
+      }
+    }
+    // All of the chunks were marked as complete, so let's exit and we
+    // should be safeish to close the file.
+    if(chunksToUpload.size() == 0)
```
dkillebrew: Before we loop around and check for pending chunks, shouldn't we wait for `chunksToUpload` to drain?

```
while chunksToUpload is not empty:
  sleep
then loop again and describe all files
etc.
```

src/ua/main.cpp
```diff
@@ -529,6 +543,39 @@ void printEnvironmentInfo() {
...
+    }
+
+    // Wait for the upload threads to complete.
+    DXLOG(logINFO) << " upload...";
+    for (int i = 0; i < (int) uploadThreads.size(); ++i) {
+      uploadThreads[i].join();
+    }
+  }
+}
+
```
dkillebrew: Rather than silently 'give up' after 3 attempts to reupload `pending` chunks, maybe we should print a warning to stderr/stdout and then terminate the program with a non-success exit code?

src/ua/main.cpp
godotgildor: I got this from file.cpp:184 which indicates that chunks are 1-based while the chunk index is 0-based.  (Also see chunk.cpp:470)

src/ua/main.cpp
godotgildor: I concur.

src/ua/main.cpp
godotgildor: I didn't see any quick way to map from Files to chunks, so I have slightly tweaked this to just cache file descriptions once per retry cycle.

src/ua/main.cpp
godotgildor: I don't think so.  The way the code is currently written is let's say we check all chunks and they are all marked as complete (which should happen 99.999% of the time).  Then this break will exit us out of this retry immediately (not slowing anything down from current behavior).

In the rare case we have a chunk that needs to be uploaded, we'll blow by this check and then the next little clip of code will actually push the chunks to be uploaded back on the queue of items to upload, and then the following little clip of code will do the pause to wait for chunksToUpload to drain.  Then we'll re-enter this check loop, check to see if all these new chunks that had to be reuploaded are now marked complete etc.

src/ua/main.cpp
godotgildor: I can print out a little warning.

godotgildor:
Just pushed changes that should address most of your suggestions.  Let me know how it looks!

src/ua/main.cpp
dkillebrew: Ah, I should read more carefully. I missed the creation & join() of uploadThreads.

What if there's a problem in `uploadChunks`, `uploaded==false`, we don't take the branch at 213? Assuming we take the branch at 224, we'll add a chunk to `chunksToRead` at 239. But the read thread(s) aren't running so nothing further will happen. 
We might need to call `createWorkerThreads` and `joinWorkerThreads` to ensure all necessary threads are able to 'do their thing'.

src/ua/main.cpp
```diff
@@ -529,6 +538,63 @@ void printEnvironmentInfo() {
...
+    // Upload the chunks which weren't marked as complete.
+    DXLOG(logINFO) << " upload...";
+    for (int i = 0; i < opt.uploadThreads; ++i) {
+      uploadThreads.push_back(boost::thread(uploadChunks, boost::ref(files)));
+    }
+
+    // Wait for the upload threads to complete.
+    DXLOG(logINFO) << " upload...";
+    for (int i = 0; i < (int) uploadThreads.size(); ++i) {
+      uploadThreads[i].join();
```
dkillebrew: I think we have to interrupt the threads, or else they never exit.

Looks like the current code waits for the monitor thread to finish before doing interrupt & join worker threads. Monitor exits when finished() is true.

src/ua/main.cpp
```diff
@@ -529,6 +538,63 @@ void printEnvironmentInfo() {
...
+    map<string, JSON> fileDescriptions;
+    while (!chunksFinished.empty()) {
+      Chunk *c = chunksFinished.consume();
+
+      // Cache file descriptions so we only have to do once per file,
+      // not once per chunk.
+      if (fileDescriptions.find(c->fileID) == fileDescriptions.end())
+          fileDescriptions[c->fileID] = fileDescribe(c->fileID);
+
+      if (!is_chunk_complete(c, fileDescriptions[c->fileID])) {
```
dkillebrew: Assuming that I've identified some correctness issues below, we should ensure this code is actually tested. Let's force this branch to be taken. I suggest (for testing only) change this line to take the branch 10% of the time, and then upload something that will be many parts. Perhaps make NUM_CHUNK_CHECKS very big to test that we converge to a completely finished state.

src/ua/main.cpp
godotgildor: I think you are right.  I had assumed that each upload thread would exit once the upload is complete.  I wonder why they don't?

Oh well, I'm sure there's a reason.  Seems like following the same pattern as the main portion, creating a monitor thread and letting it interrupt the worker threads should work...

src/ua/main.cpp
godotgildor: I also found another hiccup which should have been obvious to me, but became apparent when running the test you suggested.  Once the chunk is uploaded, the actual data in the chunk is deleted.  This obviously makes sense because otherwise by the end of the upload you'd have all the data you uploaded present in memory!  So rather than simply push the chunks to be retried into the upload queue, we should push them into the read queue so that the data is reloaded, optionally compressed, and then uploaded.  This actually simplifies the code I think.  

Anyway, with these changes, I then ran the test you suggested:
I uploaded a 72 MB file in 5 MB chunks and had each chunk pretend to fail 50% of the time.  I allowed up to 30 retry attempts (only needed 3-4 of course).  I then uploaded the same file using the current version of ua.  I downloaded the two files and ran md5sum to verify that they have the same hash.  They do.  

I ran this test with ua both with and without a compression step.  Both checked out good.

dkillebrew:
LGTM
Good work, thanks!

PR #95, opened 2015-04-27
PTFM-14962 Use devtools to install versioned CRAN execDepends

kislyuk:
@psung please review. @billagee interest

Note: although devtools supports installation of package versions, the R dependency metadata schema does not seem to accommodate it, so dependencies get pulled in versionless. In the case of ggplot2, which fails because the plyr dependency is unavailable, this means the exec depends spec needs to be updated to specify a version for plyr: https://github.com/dnanexus/nucleus/commit/91a75b0


psung:
LGTM

kislyuk:
In https://github.com/dnanexus/dx-toolkit/commit/e5d1b562b03635c89cbf1920d3d1b8baa3f54a16

PR #98, opened 2015-05-28
Fix the build/centos_install_python27.sh script

hammer:
Python doesn't seem to make `bz2` files available anymore.


kislyuk:
Thanks!

PR #101, opened 2015-06-09
PTFM-15319 return true on JBOR dxlink

kislyuk:
@ptn24 please review

### DIFF
--- a/src/python/dxpy/bindings/dxdataobject_functions.py
+++ b/src/python/dxpy/bindings/dxdataobject_functions.py
@@ -63,7 +63,17 @@ def is_dxlink(x):
     Returns whether *x* appears to be a DNAnexus link (is a dict with
     key ``"$dnanexus_link"``) with a referenced data object.
     '''
-    return isinstance(x, dict) and '$dnanexus_link' in x and (isinstance(x['$dnanexus_link'], basestring) or isinstance(x['$dnanexus_link'], dict) and 'id' in x['$dnanexus_link'])
+    if not isinstance(x, dict):
+        return False
+    if '$dnanexus_link' not in x:
+        return False
+    link = x['$dnanexus_link']
+    if isinstance(link, basestring):
+        return True
+    elif isinstance(link, dict):
+        return ('id' in link or 'job' in link)
+    else:
+        return False
 
 def get_dxlink_ids(link):
     '''
--- a/src/python/dxpy/bindings/dxdataobject_functions.py
+++ b/src/python/dxpy/bindings/dxdataobject_functions.py
@@ -63,7 +63,9 @@ def is_dxlink(x):
     Returns whether *x* appears to be a DNAnexus link (is a dict with
     key ``"$dnanexus_link"``) with a referenced data object.
     '''
-    if not isinstance(x, dict) or '$dnanexus_link' not in x:
+    if not isinstance(x, dict):
+        return False
+    if '$dnanexus_link' not in x:
         return False
     link = x['$dnanexus_link']
     if isinstance(link, basestring):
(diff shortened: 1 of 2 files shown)


src/python/dxpy/bindings/dxdataobject_functions.py
```diff
@@ -63,7 +63,17 @@ def is_dxlink(x):
     Returns whether *x* appears to be a DNAnexus link (is a dict with
     key ``"$dnanexus_link"``) with a referenced data object.
     '''
-    return isinstance(x, dict) and '$dnanexus_link' in x and (isinstance(x['$dnanexus_link'], basestring) or isinstance(x['$dnanexus_link'], dict) and 'id' in x['$dnanexus_link'])
+    if not isinstance(x, dict):
+        return False
+    if '$dnanexus_link' not in x:
```
ptn24: Let's combine this with the check in L66.

src/python/dxpy/bindings/dxdataobject_functions.py
```diff
@@ -63,7 +63,17 @@ def is_dxlink(x):
...
+        return False
+    link = x['$dnanexus_link']
+    if isinstance(link, basestring):
+        return True
```
ptn24: This looks like it changes the original logic more so than simply saying that 'job' can also be in the `$dnanexus_link`. Is that intentional?

src/python/dxpy/bindings/dxdataobject_functions.py
kislyuk: Combined.

src/python/dxpy/bindings/dxdataobject_functions.py
kislyuk: The reason I unwrapped the conditionals here and above is that the line was far too long and hard to read (including relying on multiple operator precedence rules without parentheses), and dereferenced the dictionary key multiple times. The logic remains the same.

mckinsel: The assumption that a dxlink will have a key 'id' may be embedded elsewhere in the code. For example, will this change break `_guess_link_target_type`?

ptn24: As we discussed, the logic is indeed changed. Please verify the appropriate logic.

src/python/dxpy/bindings/dxdataobject_functions.py
kislyuk: On second thought, I've reverted that change. Let's keep these clauses separate to avoid any possible confusion from the predicate ordering (leading `not` or the need to use parentheses for grouping).

src/python/dxpy/bindings/dxdataobject_functions.py
kislyuk: @ptn24, I looked over this clause again and tested it with a bunch of inputs, and can confirm that it's logically equivalent. What was the input that you thought was not equivalent?

I've added tests to confirm this.

kislyuk: @mckinsel, this doesn't break _guess_link_target_type, because that function already craps out with an exception when an unexpected link format is encountered. It will continue to throw an exception in the same manner for JBORs.

ptn24: I apologize for the mistake; it is indeed logically equivalent. Thank you :)

ptn24:
LGTM.

kislyuk:
Thanks, merged in f7744271cbcfa3be6a5cc9f890eb986828bf041c

PR #102, opened 2015-06-09
Adding several configuration options which should prevent uploads from getting stuck

evanworley:
@jmdale @asimenos Please review and let me know if there is any reason we shouldn't push this out to prod right now.


jmdale:
LGTM

Seems pretty safe to me.

evanworley:
Thanks @jmdale , merging it in

PR #105, opened 2015-08-07
Added script to slice zipped VCF

johnwallace123:
I've added a script to slice a VCF that users of DNANexus may find useful.  Below is my git commit, which explains the usage of the script, its advantages and limitations.  You are free to use the code with any license you see fit.  While the code works for our needs, it may need modification to suit the needs of a generic DNANexus customer; please feel free to make any modifications you deem necessary.

Thanks!

Adding a script to slice a block gzipped VCF file given the VCF file and
its tabix index.  Both the tabix index and the VCF file may be given as:
- a file name
- a URL
- a DXFile ID (note that project-<>:file-<> doesn't seem to work)

Also needed is a list of intervals which can be either a file, as above
or a comma separated list of intervals in the form chrom:start-stop.
These intervals are CLOSED, 1-based intervals.

The script will write the file to a provided filename (mandatory!), but
it can append to an existing file, and it also has an option to prevent
writing the last "empty" bgzip block signifying end of file.
Additionally, printing of the header is optional as well.

The script will decompress only the part of the VCF file needed to
determine variant inclusion in a file, so is particularly well-suited to
downloading entire chromosomes at a time.  On the other hand, this
script will perform very poorly when downloading a large number of small
regions (<1,000nt) very close together (<16,000nt separation).

Note that this script differs from tabix in that it does NOT look for an
END= tag in the INFO field to determine overlapping.  Additionally, it
does not look at the size of the REF allele to determine if the position
should be retrieved.  A variant is retrieved IF AND ONLY IF the reported
POS is contained in a target region.

### DIFF
--- a/src/python/dxpy/cli/vcf_slicer.py
+++ b/src/python/dxpy/cli/vcf_slicer.py
@@ -1,0 +1,879 @@
+#!/usr/bin/env python
+
+import sys
+import dxpy
+import argparse
+import zlib
+import urllib
+import cStringIO
+import struct
+import gzip
+import bisect
+from collections import deque
+
+_bgzf_magic = b"\x1f\x8b\x08\x04"
+_bgzf_header = b"\x1f\x8b\x08\x04\x00\x00\x00\x00\x00\xff\x06\x00\x42\x43\x02\x00"
+_bgzf_eof = b"\x1f\x8b\x08\x04\x00\x00\x00\x00\x00\xff\x06\x00BC\x02\x00\x1b\x00\x03\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+_bytes_BC = b"BC"
+
+
+def convert_vfp(vfp):
+	'''
+	convert the virtual file pointer into a tuple of 
+	(block offset, within block offset)
+	'''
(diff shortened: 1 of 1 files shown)


kislyuk:
Hi John,

Thank you for submitting this, and apologies for the delay in our response. We very much appreciate your contribution. With that said, we currently treat dx-toolkit as a repository for general purpose SDK tools for interfacing with the DNAnexus API, and we have not established a policy for genomics-specific external contributions like this. While we work on establishing this policy internally, we are sadly unable to accept this script here as is, both because there isn't yet a "contrib" directory where we would want such scripts to go, and also because there isn't a contributor licensing/IP assignment policy worked out.

In the meantime, I suggest that you create a public repo of your own and put this script there. From there you can bundle it into DNAnexus apps, by copy or by reference (via git execDepends), and share it with other users. We also do have a community contributions page on the DNAnexus Wiki (https://wiki.dnanexus.com/Community/Home) that you can edit to add this resource there. @asimenos can also provide further tips for sharing these resources.

Thanks again for your contribution, and sorry we can't accept it at this time.

PR #106, opened 2015-08-20
PTFM-14569 fixing dx help new workflow issue

godotgildor:
Previously when adding a subparser it got the name of the subparser from the last item in the

_choices_actions list from the subparsers_action.  However, if a subparser does not have a help
field, like new gtable, then it won't have an entry in the subparsers_action._choices_actions
list.  Instead, it will grab the name of the previous subparser that contained a help field,
which in this case is dx new workflow.

This fix grabs the name from the prog field of the subparser instead.  I checked
all subparsers to see if their name from the prog field would match the name from
the _choices_actions list, and they all do, EXCEPT: new gtable (which was incorrect
before and the subject of this bug) and make_download_url and upgrade.  The latter two
did not match because they did not explicitly set the prog field when creating the subparser.
I corrected this oversight, and now they all match except new gtable, which is now correct
whereas before it was wrong.

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -3347,19 +3347,27 @@ class DXArgumentParser(argparse.ArgumentParser):
                                                              msg=message))
 
 def register_subparser(subparser, subparsers_action=None, categories=('other', )):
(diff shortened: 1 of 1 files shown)


kislyuk:
LGTM

PR #114, opened 2015-10-01
Make download all inputs availble to dxpy

godotgildor:


src/python/dxpy/bindings/download_all_inputs.py
```diff
@@ -0,0 +1,161 @@
...
+    with concurrent.futures.ThreadPoolExecutor(max_workers=max_num_parallel_downloads) as executor:
+       future_files = {executor.submit(_download_one_file, file_rec, idir): file_rec for file_rec in to_download}
+       for future in concurrent.futures.as_completed(future_files):
+           file_rec = future_files[future]
+           try:
+               future.result()
+           except Exception as exc:
+               sys.stderr.write('%r -> %s generated an exception' % (file_rec['src_file_id'], file_rec['trg_fname']))
+               raise
+           else:
```
kislyuk: Does this clause do anything?

src/python/dxpy/bindings/download_all_inputs.py
```diff
@@ -0,0 +1,161 @@
...
+# Download files in parallel
+#   to_download: list of tuples describing files to download
+def _parallel_file_download(to_download, idir, max_num_parallel_downloads):
+    with concurrent.futures.ThreadPoolExecutor(max_workers=max_num_parallel_downloads) as executor:
+       future_files = {executor.submit(_download_one_file, file_rec, idir): file_rec for file_rec in to_download}
+       for future in concurrent.futures.as_completed(future_files):
+           file_rec = future_files[future]
+           try:
+               future.result()
+           except Exception as exc:
```
kislyuk: To make this intterruptible by Ctrl+C on 2.7, please add a timeout to `as_completed(timeout=sys.maxint)` and an `except KeyboardException: os._exit()` catch. See comments in https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/bindings/dxfile_functions.py#L255-L260

src/python/dxpy/bindings/download_all_inputs.py
godotgildor: Which clause are you asking about?

This code is verbatim from the existing download_all_inputs code.  Basically, this code previously lived in src/python/scripts in a simple python script that was only usable from bash apps/applets.  This PR moves the code more or less as is from the scripts directory to the src/python/dxpy/bindings/ directory, so that the same functionality is available from python.

src/python/dxpy/bindings/download_all_inputs.py
godotgildor: I have updated the PR to include a try/except block wrapping the concurrent.futures block with the same pattern as the example @kislyuk pointed to.  I also added the suggested timeout.

src/python/dxpy/bindings/download_all_inputs.py
kislyuk: Thanks for the explanation. I was just commenting that the "else: pass" snippet seems like it doesn't do anything and could be removed. I understand it's outside the scope of this PR, so it's not blocking the merge.

src/python/dxpy/bindings/download_all_inputs.py
```diff
@@ -0,0 +1,168 @@
...
+        if (exclude is None) or (d not in exclude):
+            dirs_to_create.append(d)
+
+    # Create the directory structure, in preparation for download.
+    # Allows performing the download in parallel.
+    _create_dirs(idir, dirs_to_create)
+
+    # Remove excluded inputs
+    if (exclude is not None) and len(exclude) > 0:
+        inputs = file_load_utils.filter_dict(inputs, args.exclude)
```
kislyuk: I believe `args.exclude` is an invalid reference here. To catch errors like this, you can run `pylint -E <file.py>` and/or `pyflakes <file.py>`.

kislyuk:
LGTM pending the fix for `args.exclude` and any other issues the linters report.

JoeSachmo:
Thanks! Merged in the following commit:
https://github.com/dnanexus/dx-toolkit/commit/20faf4b5f5609d432dd1bfecb1020fb3f67db0af

PR #115, opened 2015-10-09
dxpy.download_dxfile: restore positional args compatibility

kislyuk:
Restore compatibility with function calls that use the dxid positional
argument name as a keyword arg.

@psung, please review

### DIFF
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -106,7 +106,7 @@ def download_dxfile(dxid, filename, chunksize=dxfile.DEFAULT_BUFFER_SIZE, append
     :param append: If True, appends to the local file (default is to truncate local file if it exists)
     :type append: boolean
 
-    Downloads the remote file referenced by *dxid* or *dxfile* and saves it to *filename*.
+    Downloads the remote file referenced by *dxid* and saves it to *filename*.
 
     Example::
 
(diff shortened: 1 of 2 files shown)


src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -96,17 +96,19 @@ def _get_executor():
         _executor = ThreadPoolExecutor(max_workers=cpu_count())
     return _executor
 
-def download_dxfile(dxfile_or_id, filename, chunksize=dxfile.DEFAULT_BUFFER_SIZE, append=False, show_progress=False,
-                    project=None, **kwargs):
+def download_dxfile(dxid, filename, dxfile=None, chunksize=dxfile.DEFAULT_BUFFER_SIZE, append=False,
+                    show_progress=False, project=None, **kwargs):
```
psung: If you really want to guarantee backwards compatibility, retain the same ordering of all the arguments that existed previously and add your new one at the end:

```
                    project=None, dxfile=None, **kwargs):
``

However, both this and the proposed API above are kind of annoying to use. `dxid` and `dxfile` are mutually exclusive, but if you supply `dxfile` as a keyword argument you then still have to override `dxid=None`. One option is to make all the arguments keyword:

```
def download_dxfile(dxid=None, filename=None, ..., dxfile=None, **kwargs):
```

But I find this kind of unsatisfying too since you lose the statically available information that at least two arguments are needed. Frankly I'd be fine with restoring the previous signature exactly and just documenting the fact that you can supply a DXFile or an ID in the `dxid` position. That might be the least bad option, actually.

kislyuk:
OK - I took your last suggestion and restored the original signature. How does this look?

psung:
LGTM, thanks!

psung:
Squash merged as 79b3b4f5

PR #116, opened 2015-10-12
dxpy.download_dxfile: handle files with no part md5 metadata

kislyuk:
@psung, please take a look.


psung:
LGTM, but please add a regression test at some point.

Will the checksumming behavior ever be documented in helpstrings, the wiki, etc.? If and when there is any public expectation that downloads will be verified end-to-end, then we must warn the user whenever this has not been done.

kislyuk:
Thanks. I've updated PTFM-16281 to mention this. Proposed warning API is in https://github.com/dnanexus/dx-toolkit/commit/4e09ed2.

PR #117, opened 2015-10-26
PTFM-15570 dxpy supplies project ID on file/download requests

psung:
@ptn24 Preview of coming attractions, just so you have a better idea of where I'm going with this. I will open up a Revision for this as soon as D601 lands.


psung:
Sent for review as D433, so all further discussion should happen there.

PR #119, opened 2015-12-17
Add omit-resources feature

godotgildor:
Add a simple command line argument to dx get to allow users to fetch
the applet without the resources.


kislyuk:
Thanks! Could you add a test in the test suite?

PR #121, opened 2016-01-12
Allow insecure download of Python on CentOS

mckinsel:
Certificate checking doesn't work for some customers, but they still want to
use dx-toolkit with CentOS or RHEL. The CentOS python installation script
won't work for them unless they add a `-k` to every curl call. This change
creates an option for the whole script so they don't have to do that.


psung:
Merged as e9ced7e

PR #123, opened 2016-02-09
Fix print_progress bug

godotgildor:
There was a small bug when printing the status of a download in that
each time we print the new status, we have a carriage return which
resets the cursor to the beginning of the line so that each new progress
overwrites the previous line.  This works fine as long as the new line
is greater than or equal to the previous line's length, but if it's shorter,
then you'll see part of the old line left.


psung:
LGTM, squash merged as d768118

PR #124, opened 2016-02-11
add the ability to create the source jar for dnanexus-api Java bindings ...

kedarmhaswade:
This commit makes it possible to generate source jar when one does mvn verify. It also adds a trivial entry to .gitignore.


psung:
Thank you, merged to master as 9ded343.

PR #126, opened 2016-02-13
Re-adding --omit-resources flag

godotgildor:
The commit for PTFM-17064 accidentally removed the omit-resources
option from fetching an applet.  This commit puts it back in
and adds it for "get app".

### DIFF
--- a/src/python/test/test_dxclient.py
+++ b/src/python/test/test_dxclient.py
@@ -6491,6 +6491,46 @@ class TestDXGetExecutables(DXTestCase):
             self.assertTrue(os.path.exists("destfile"))
             self.assertTrue(os.path.exists(os.path.join("destfile", "dxapp.json")))
 
+    def test_get_applet_omit_resources(self):
+        # TODO: not sure why self.assertEqual doesn't consider
+        # assertEqual to pass unless the strings here are unicode strings
+        app_spec = {
+            "name": "get_applet",
+            "dxapi": "1.0.0",
+            "runSpec": {"file": "code.py", "interpreter": "python2.7"},
+            "inputSpec": [{"name": "in1", "class": "file"}],
+            "outputSpec": [{"name": "out1", "class": "file"}],
+            "description": "Description\n",
+            "developerNotes": "Developer notes\n",
+            "types": ["Foo"],
+            "tags": ["bar"],
+            "properties": {"sample_id": "123456"},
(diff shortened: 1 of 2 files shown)


psung:
Thanks, and sorry about the brokenness. Let's discuss adding a regression test so this doesn't inadvertently get broken again.

src/python/test/test_dxclient.py
```diff
@@ -6606,6 +6646,59 @@ def test_get_app(self):
             with self.assertSubprocessFailure(stderr_regexp='can only call.*\n'):
                 run("dx get " + new_app_id)
 
+    def test_get_app_omit_resources(self):
+        self.maxDiff = None
```
psung: Is this still needed?

src/python/test/test_dxclient.py
```diff
@@ -6491,6 +6491,46 @@ def test_get_applet(self):
...
+        os.mkdir(os.path.join(app_dir, "resources"))
+        with open(os.path.join(app_dir, "resources", "resources_file"), 'w') as f:
+            f.write('content\n')
+        new_applet_id = json.loads(run("dx build --json " + app_dir))["id"]
+        with chdir(tempfile.mkdtemp()):
+            run("dx get --omit-resources " + new_applet_id)
+            self.assertFalse(os.path.exists(os.path.join("get_applet", "resources")))
+
+            output_json = json.load(open(os.path.join("get_applet", "dxapp.json")))
+            self.assertTrue("bundledDepends" in output_json["runSpec"])
```
psung: Could you an assertion to the vanilla "dx get" test that bundledDepends is **not** propagated into the output_json? (Thanks! I never thought about that before.)

psung: Here and elsewhere, consider using the more specific assertion self.assertIn("bundledDepends", ...)

godotgildor: I have made the suggested edits.  One note: in the standard dx get test, the test creates a dxapp.json for the test applet, builds the applet, and then get's it.  The original dxapp.json did not have a bundledDepends in it, and there is an assertion that the pre-build and post-build dxapp.json's match, and so if the post-build DID have a bundledDepends argument, then presumably this would already be caught.  Nevertheless, probably a good idea to have the check explicitly called out.

psung:
Thanks! Merged as 36feeae

PR #127, opened 2016-02-17
More informative error for dx add member

thanhhahaha:
The CLI will suggest to the user to use 'dx update member' instead if calling dx add member on existing org member. 

@ptn24 please review

### DIFF
--- a/src/python/dxpy/cli/org.py
+++ b/src/python/dxpy/cli/org.py
@@ -74,7 +74,7 @@ def add_membership(args):
     except:
         pass
     else:
-        raise DXCLIError("Cannot add a user who is already a member of the org")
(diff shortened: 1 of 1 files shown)


src/python/dxpy/cli/org.py
```diff
@@ -74,7 +74,7 @@ def add_membership(args):
     except:
         pass
     else:
-        raise DXCLIError("Cannot add a user who is already a member of the org")
+        raise DXCLIError("Cannot add a user who is already a member of the org. To update existing member's permissions use 'dx update member'")
```
ptn24: "existing" -> "an existing".

ptn24: "permissions" -> "permissions,".

ptn24:
LGTM.

ptn24:
47e859f.

PR #131, opened 2016-03-21
Clarifying help string

godotgildor:
Updating the help string for dx terminate to indicate that it can also be
used for analyses.


psung:
Thanks. Merged as c8ccd2dd.

PR #135, opened 2016-03-30
'.sh' suffix has been added to the dx environment autoload file

storozhilov:
`/etc/profile` script on Ubuntu processes `/etc/profile.d/*.sh` files only, so I added a respective suffix to the `/etc/profile.d/dnanexus.environment` file.


psung:
Thank you! Merged to master as 5b0dbe697a3aceecb9f6441732c989562ae4eeb8

PR #136, opened 2016-04-07
Updating dx

godotgildor:
Made the following changes:
1. Bug fix for checking version of readline module.  pyreadline module does not have __doc__ variable, so need to check that it's not NONE before searching for libedit.
2. Adding ability to specify protocol to dx logout, just as you can for dx login.
3. Setting default protocol for authserver communication to https. Can be overriden by explicitly passing in a different protocol, but will no longer make assumptions based upon port.
4. Use the same get_auth_server_name helper function in dx login as we use in dx logout code.
5. Add support for China staging api and corresponding auth server to the get_auth_server_name helper function.


bpr:
LGTM.

PR #137, opened 2016-04-08
AZURE-57 dxpy.DXFile supports parameterized part sizes

terrace-sharib:
Changing bindings and error messaging to reflect upload limitation differences between S3 and Azure.


.gitignore
```diff
@@ -3,6 +3,7 @@
 /build/wrapper_table.json
 /build/doc_env/
 /build/py_env/
+/build/py27_env/
```
psung: Can you try removing this directory in your local checkout and seeing if it comes back? I believe this was renamed in the Makefiles to py_env recently, so you should be able to remove this.

src/python/dxpy/bindings/dxfile.py
```diff
@@ -112,24 +120,29 @@ def _ensure_http_threadpool(cls):
         if cls._http_threadpool is None:
             cls._http_threadpool = dxpy.utils.get_futures_threadpool(max_workers=cls._http_threadpool_size)
 
+    @classmethod
+    def _set_file_limits(cls, file_limits):
+        cls._buffer_size = file_limits['minimumPartSize']
```
psung: These should be instance variables so that clients can use file bindings against files in multiple clouds in the same process.

src/python/dxpy/bindings/dxfile.py
```diff
@@ -112,24 +120,29 @@ def _ensure_http_threadpool(cls):
...
+        cls._maximum_parts = file_limits['maximumNumParts']
+        cls._minimum_part_size = file_limits['minimumPartSize']
+        cls._maximum_part_size = file_limits['maximumPartSize']
+
     def __init__(self, dxid=None, project=None, mode=None,
-                 read_buffer_size=DEFAULT_BUFFER_SIZE, write_buffer_size=DEFAULT_BUFFER_SIZE):
+                 read_buffer_size=DEFAULT_BUFFER_SIZE, write_buffer_size=DEFAULT_BUFFER_SIZE, file_size=1, file_is_mmapd=False):
```
psung: file_size should default to None, or some other special value, since clients are not obligated to supply it. Also, it's not actually binding on the client to conform to this, so consider calling this expected_file_size.

src/python/dxpy/bindings/dxfile.py
```diff
@@ -363,6 +396,16 @@ def write(self, data, multithread=True, **kwargs):
             does not affect where the next :meth:`write` will occur.
 
         '''
+        self.file_limits = dxpy.api.project_describe(self.project, {'fields': {'fileUploadParameters': True}})['fileUploadParameters']
+        self._set_file_limits(self.file_limits)
+
+        self._write_buf_size = self._get_buffer_size_for_file(self.file_size, self.file_is_mmapd)
```
psung: get_buffer_size_for_file could just ensure that we respect the minimum and maximum part sizes.

I also think that instead of throwing an error, we should just adjust the buffer size to comply with the maximum/minimum (selecting the nearest legitimate value, and possibly not using the exact part size the client wanted). Otherwise, a part size that the client sets for Azure will fail on AWS, and vice versa.

src/python/dxpy/bindings/dxfile.py
```diff
@@ -402,6 +445,27 @@ def write_request(data_for_write_req):
...
+            Returns the file size in readable form'
+        """
+        B = float(bytes)
+        KB = float(1024)
+        MB = float(KB * 1024)
+        GB = float(MB * 1024)
+        TB = float(GB * 1024)
+
+        if B < KB:
+            return '{0} {1}'.format(B, 'Bytes' if 0 == B > 1 else 'Byte')
```
psung: This desugars to (0 == B and B > 1) which never holds. I think you meant to write "or".

src/python/dxpy/bindings/dxfile.py
```diff
@@ -701,7 +765,7 @@ def read(self, length=None, use_compression=None, project=None, **kwargs):
         # file exists. Otherwise, it's an error.
         #
         # If project=None, we fall back to the project attached to this handler
-        # (if any). If this is supplied, it's treated as a hint: if it's a
+        # (if any). If this is supplied, it's treated as a hint: if it's a/d
```
psung: Stray typo here?

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -349,12 +329,14 @@ def upload_local_file(filename=None, file=None, media_type=None, keep_open=False
...
     fd = file if filename is None else open(filename, 'rb')
+    file_is_mmapd = hasattr(fd, "fileno")
 
     try:
         file_size = os.fstat(fd.fileno()).st_size
     except:
         file_size = 0
-    buffer_size = _get_buffer_size_for_file(file_size, file_is_mmapd=hasattr(fd, "fileno"))
+
+    file_is_mmapd = hasattr(fd, "fileno")
```
psung: Remove duplicated line

src/python/test/test_dxabs.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2013-2015 DNAnexus, Inc.
```
psung: "2013-2015" => "2016"

src/python/test/test_dxabs.py
```diff
@@ -0,0 +1,100 @@
...
+def setUpTempProjects(thing):
+    thing.old_workspace_id = dxpy.WORKSPACE_ID
+    thing.proj_id = dxpy.api.project_new({'name': 'azure-test-project', 'region': 'azure:westus'})['id']
+    dxpy.set_workspace_id(thing.proj_id)
+
+def tearDownTempProjects(thing):
+    dxpy.api.project_destroy(thing.proj_id, {'terminateJobs': True})
+    dxpy.set_workspace_id(thing.old_workspace_id)
+
+class TestDXFile(unittest.TestCase):
```
psung: I'd like these tests to only run when the user specifically calls for them to be run, since not all users will have Azure regions enabled. I added some infrastructure to support this on master. Please merge it into your branch and then:

* Make this test class conditional on testutil.TEST_AZURE (using unittest.skipUnless)
* The region should be set based on testutil.TEST_AZURE instead of hard-coded here (in case the user has access to some azure region other than westus.

src/python/test/test_dxabs.py
```diff
@@ -0,0 +1,100 @@
...
+        setUpTempProjects(self)
+
+        self.new_file = tempfile.NamedTemporaryFile(delete=False)
+        self.new_file.close()
+
+    def tearDown(self):
+        os.remove(self.new_file.name)
+        tearDownTempProjects(self)
+
+    def test_upload_download_files_dxfile(self):
```
psung: Can you test a file that's larger than 4MB?

src/python/dxpy/bindings/dxfile.py
```diff
@@ -38,7 +38,7 @@
 
 DXFILE_HTTP_THREADS = min(cpu_count(), 8)
 MIN_BUFFER_SIZE = 1024*1024
-DEFAULT_BUFFER_SIZE = 1024*1024*16
+DEFAULT_BUFFER_SIZE = 1024*1024*4
```
psung: Once you've made the changes suggested below, please change this back to 16MB so as to leave the existing behavior unchanged for AWS users.

src/python/dxpy/bindings/dxfile.py
```diff
@@ -147,6 +160,26 @@ def __init__(self, dxid=None, project=None, mode=None,
...
+        """
+        # Raise buffer size (for files exceeding DEFAULT_BUFFER_SIZE * 10k
+        # bytes) in order to prevent us from exceeding 10k parts limit.
+        min_buffer_size = int(math.ceil(float(file_size) / self._maximum_parts))
+        buffer_size = max(self._buffer_size, min_buffer_size)
+        if file_size >= 0 and file_is_mmapd:
+            # For mmap'd uploads the buffer size additionally must be a
+            # multiple of the ALLOCATIONGRANULARITY.
+            buffer_size = int(math.ceil(float(buffer_size) / mmap.ALLOCATIONGRANULARITY)) * mmap.ALLOCATIONGRANULARITY
+        if buffer_size * self._maximum_parts < file_size:
```
psung: It would be nice to assert here that the file is no larger than the maximum file size. This would allow dxpy to fail fast if the client tries to upload a file larger than 200GB on Azure.

psung:
Also, note that "close" has a check to upload an empty part if there have been 0 bytes (and 0 parts) previously written to the file. This should be disabled conditional on emptyLastPartAllowed being false.

You may not be able to test this until backend support is complete though.

src/python/dxpy/bindings/dxfile.py
terrace-sharib: Wouldn't that be a change to the existing functionality?  Does this code adjust the buffer size to the defaults for max and min?

```buffer_size = max(self._buffer_size, min_buffer_size)```

So, wouldn't we basically just be removing the bit below?
```
if buffer_size * self._maximum_parts < expected_file_size:
            raise AssertionError('part size is not large enough to complete upload')
```

psung: Yes, you're correct (on both counts).

I'm fine with changing the existing functionality since this should only convert previously broken code (code that specified a bogus buffer size) to now working code... right?

terrace-sharib: Good point.  I will change.

PR #139, opened 2016-04-14
AZURE-57-dxpy PR modifcations

terrace-sharib:
Includes the following:
* Uses the TEST_AZURE flag to skip tests for non-Azure infrastructure
* Reverted the DEFAULT_BUFFER_SIZE to the original S3 limit
* Added tests for a larger file size
* Removed some typos and duplicate code


psung:
Thanks! Refactored (for testability and to put all the interesting logic in a single method) and added some more tests.

Merged as 46e3e7fc912eaeb2dfcfc721b99fd44e9301a31c

PR #142, opened 2016-04-21
Added ability to dereference symlinks

johnwallace123:
Adding the ability to dereference symbolic links that are pointing to
remote files (i.e. not in the resource directory) of an app.  This patch
will ensure the following cases:

1. Links to local files and directories are kept
2. Links to remote files are dereferenced
3. Broken links cause an AppBuilderError
4. Links to remote directories cause an AppBuilderError
5. Links to local links (regardless of destination) are kept
6. Links to remote links (links to files) are dereferenced
7. Absolute links are treated as remote links (case 2,4)
8. Links that travel outside of the resource directory (i.e. ../resources/foo)
are treated as remote links (case 2,4).  This is pretty evil; please
don't be evil.

In any case, adding the new option "--force-symlinks" to the "dx build"
command will restore the previous behavior and keep symlinks that may or
may not resolve.  The use case for this could be perhaps a link in
resources that will point to a directory known to exist in a DNAnexus
instance (like /home/dnanexus for example).

Note that in the upload_resources_advanced test relies on the "old"
behavior of symlinks, so I also added the "--force-symlinks" tag to
those unit tests.

### DIFF
--- a/src/python/dxpy/app_builder.py
+++ b/src/python/dxpy/app_builder.py
@@ -36,9 +36,13 @@ the effective destination project.
 
 from __future__ import print_function, unicode_literals, division, absolute_import
 
-import os, sys, json, subprocess, tempfile, multiprocessing, gzip, io, tarfile, stat
+import os, sys, json, subprocess, tempfile, multiprocessing
 import datetime
+import gzip
 import hashlib
+import io
+import tarfile
+import stat
 
 import dxpy
 from . import logger
@@ -130,19 +134,19 @@ def get_destination_project(src_dir, project=None):
     if project is not None:
         return project
     return _get_applet_spec(src_dir)['project']
-    
+
 def is_link_local(link_target):
     """
     :param link_target: The target of a symbolic link, as given by os.readlink()
-    :type link_target: string    
+    :type link_target: string
     :returns: A boolean indicating the link is local to the current directory.
               This is defined to mean that os.path.isabs(link_target) == False
-              and the link NEVER references the parent directory, so 
+              and the link NEVER references the parent directory, so
               "./foo/../../curdir/foo" would return False.
     :rtype: boolean
     """
     is_local=(not os.path.isabs(link_target))
-    
+
     if is_local:
         # make sure that the path NEVER extends outside the resources directory!
         d,l = os.path.split(link_target)
@@ -151,13 +155,13 @@ def is_link_local(link_target):
             link_parts.append(l)
             d,l = os.path.split(d)
         curr_path = os.sep
-    
+
         for p in reversed(link_parts):
             is_local = (is_local and not (curr_path == os.sep and p == os.pardir) )
             curr_path = os.path.abspath(os.path.join(curr_path, p))
-            
-    return is_local    
-    
+
+    return is_local
+
 def _fix_perms(perm_obj):
     """
     :param perm_obj: A permissions object, as given by os.stat()
@@ -172,9 +176,9 @@ def _fix_perms(perm_obj):
     ret_perm = perm_obj | stat.S_IROTH | stat.S_IRGRP | stat.S_IRUSR
     if ret_perm & stat.S_IXUSR:
         ret_perm = ret_perm | stat.S_IXGRP | stat.S_IXOTH
-    
+
     return ret_perm
-    
+
 def _fix_perm_filter(tar_obj):
     """
     :param tar_obj: A TarInfo object to be added to a tar file
@@ -183,7 +187,7 @@ def _fix_perm_filter(tar_obj):
     :rtype: tarfile.TarInfo
     """
     tar_obj.mode = _fix_perms(tar_obj.mode)
-    return tar_obj   
+    return tar_obj
 
 
 def upload_resources(src_dir, project=None, folder='/', ensure_upload=False, force_symlinks=False):
@@ -222,15 +226,15 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False, for
     if os.path.exists(resources_dir) and len(os.listdir(resources_dir)) > 0:
         target_folder = applet_spec['folder'] if 'folder' in applet_spec else folder
 
-        # While creating the resource bundle, optimistically look for a 
-        # resource bundle with the same contents, and reuse it if possible. 
-        # The resource bundle carries a property 'resource_bundle_checksum' 
-        # that indicates the checksum; the way in which the checksum is 
-        # computed is given below.   If the checksum matches  (and 
+        # While creating the resource bundle, optimistically look for a
+        # resource bundle with the same contents, and reuse it if possible.
+        # The resource bundle carries a property 'resource_bundle_checksum'
+        # that indicates the checksum; the way in which the checksum is
+        # computed is given below.   If the checksum matches  (and
         # ensure_upload is False), then we will use the existing file,
         # otherwise, we will compress and upload the tarball.
-        
-        
+
+
         # The input to the SHA1 contains entries of the form (whitespace
         # only included here for readability):
         #
@@ -242,25 +246,16 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False, for
         # specified below), followed by a numeric representation of the
         # mode, and the mtime in milliseconds since the epoch.
         #
-        # Note when looking at a link, unless --force-symlinks is specified, 
-        # and the link is to be dereferenced the mtime is taken to be the 
-        # maximum of the mtime of the link itself and of the link's target.  
-        # Additionally, the mode is the mode of the link's target.  
-        # If --force-symlinks is specified, or if the link is not to be 
-        # dereferenced, we'll only look at the link itself.
-        
-        # get the absolute resource directory, and go there
-        # this is the equivalent of "-C" in tar
-        abs_resources_dir = os.path.abspath(resources_dir)
-        old_curdir = os.getcwd()
-        os.chdir(abs_resources_dir)
-        
-        
+        # Note when looking at a link, if the link is to be dereferenced,
+        # the mtime and mode used are that of the target (using os.stat())
+        # If the link is to be kept as a link, the mtime and mode are those
+        # of the link itself (using os.lstat())
+
         with tempfile.NamedTemporaryFile(suffix=".tar") as tar_tmp_fh:
-        
+
             output_sha1 = hashlib.sha1()
-            tar_fh = tarfile.open(fileobj=tar_tmp_fh, mode='w')        
-        
+            tar_fh = tarfile.open(fileobj=tar_tmp_fh, mode='w')
+
             for dirname, subdirs, files in os.walk(resources_dir):
                 if not dirname.startswith(resources_dir):
                     raise AssertionError('Expected %r to start with root directory %r' % (dirname, resources_dir))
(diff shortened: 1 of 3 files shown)


src/python/dxpy/app_builder.py
```diff
@@ -212,58 +161,199 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False):
...
+                        # make sure that the path NEVER extends outside the resources directory!
+                        d,l = os.path.split(link_target)
+                        link_parts = []
+                        while l:
+                            link_parts.append(l)
+                            d,l = os.path.split(d)
+                        curr_path = dirname
+                        for p in reversed(link_parts):
+                            curr_path = os.path.abspath(os.path.join(curr_path, p))
+                            is_local = (is_local and curr_path.startswith(resources_dir))
```
psung: To make this check work, `resources_dir` should also be made absolute. Otherwise you can end up with the following inappropriate error:

```
psung@exabase ~/source/cat $ dx build .
Error: Cannot include symlinks to directories outside of the resource directory.  u'./resources/a' points to directory u'/home/psung/source/cat/resources/b'
psung@exabase ~/source/cat $ ls -l resources
total 4.0K
lrwxrwxrwx 1 psung psung    1 Apr 25 21:58 a -> b
drwxrwxr-x 2 psung psung 4.0K Apr 25 22:04 b
```

src/python/dxpy/app_builder.py
```diff
@@ -212,58 +161,199 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False):
...
+                                               
+                        # If this is a local link, add it to the list of files (case Ib)
+                        # else raise an error
+                        if force_symlinks or is_local:
+                            files.append(subdir_name)
+                        else:
+                            raise AppBuilderException("Cannot include symlinks to directories outside of the resource directory.  %r points to directory %r" % (dir_path, dir_realpath))
```
psung: Minor, I'd prefer that error messages and the like don't include the Python "u" specifier that it attaches to strings (see the error I pasted above for an example). I suggest something like wrapping the paths in pipes.quote (which has the additional benefit of eliding quotes when they're not needed).

src/python/dxpy/app_builder.py
```diff
@@ -212,58 +161,199 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False):
...
+                # the directory (or --force-symlinks is given), we should add it 
+                # as a file.
+                for subdir_name in subdirs:
+                    dir_path = os.path.join(dirname, subdir_name)
+
+                    # If we do have a symlink,                     
+                    if os.path.islink(dir_path):
+                        # Let's get the pointed-to path to ensure that it is
+                        # still in the directory
+                        link_target = os.readlink(dir_path)
```
psung: Can this part be factored out into a method so as to unify it with the similar-looking code below?

psung: e.g. `is_link_target_local`

src/python/dxpy/app_builder.py
```diff
@@ -212,58 +161,199 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False):
...
+                dir_stat = os.lstat(dirname)
+                if not relative_dirname.startswith('/'):
+                    relative_dirname = '/' + relative_dirname
+
+                fields = [relative_dirname, str(dir_stat.st_mode), str(int(dir_stat.st_mtime * 1000))]
+                output_sha1.update(b''.join(s.encode('utf-8') + b'\0' for s in fields))
+                
+                # add an entry in the tar file for the current directory, but
+                # do not recurse!
+                subprocess.check_call(['tar', '-C', resources_dir, '-rf', tar_fh.name, '--no-recursion', '.' + relative_dirname])
```
psung: I'm concerned about the overhead of invoking tar here as well as below (once per directory and once per file being packed up). In the following benchmark, I tested building an app that contains the source distribution of node.js. A 19x slowdown is observed (though this is probably a pathologically large example, at 16,000 files and directories):

```
# master
psung@exabase ~/source/cat $ time dx build
[lint warnings stripped...]
real    0m32.804s
user    0m13.067s
sys     0m1.172s

# localSymlinkOnly
psung@exabase ~/source/cat $ time dx build
[lint warnings stripped...]
WARNING:dxpy:PUT <function get_upload_url_and_headers at 0x7f59308dcaa0>: ProtocolError: ('Connection aborted.', error(104, 'Connection reset by peer')). Waiting 1 seconds before retry 1 of 6...
INFO:dxpy:PUT <function get_upload_url_and_headers at 0x7f59308dcaa0>: Recovered after 1 retries
{"id": "applet-BvPPfp80BGv7g5Fx7JVqBzvb"}

real    10m38.361s
user    4m0.288s
sys     6m1.237s
```

Maybe @asimenos can point us to a large app? (Ideally, many small files) I'd be curious to see whether for a more realistic app the total time is dominated by the time to upload anyway.

Additionally, would it be possible to batch these calls so as to make `tar` process many files in a single command-line invocation?

src/python/dxpy/app_builder.py
```diff
@@ -212,58 +161,199 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False):
...
+
+                    # If we do have a symlink,                     
+                    if os.path.islink(dir_path):
+                        # Let's get the pointed-to path to ensure that it is
+                        # still in the directory
+                        link_target = os.readlink(dir_path)
+                        is_local=(not os.path.isabs(link_target))
+                        
+                        dir_realpath=os.path.abspath(os.path.join(dirname, link_target))
+                        # make sure that the path NEVER extends outside the resources directory!
```
psung: Is the idea below just to prevent, say, my app resources directory `/home/psung/myapp/resources` from containing a link to

```
../../../../home/psung/myapp/resources/target
```

which resolves on my machine, but wouldn't when packed up, despite being a valid link whose target lives in `resources`?

By the way, does this have some security implication, or is the idea just to not transfer links that would be broken? (If just the latter, it seems like a lot of trouble to go to, but I don't mind keeping this here.)

src/python/dxpy/app_builder.py
```diff
@@ -212,58 +161,199 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False):
...
+                                deref_link = True
+                            except OSError:
+                                # uh-oh! looks like we have a broken link!
+                                # since this is guaranteed to cause problems (and
+                                # we know we're not forcing symlinks here), we
+                                # should throw an error
+                                raise AppBuilderException("Broken symlink: Link '%r' points to '%r', which does not exist" % (true_filename, linkpath_filename) )
+                    
+                    
+                    fields = [relative_filename, str(file_stat.st_mode), str(int(file_stat.st_mtime * 1000))]
```
psung: When a link is being dereferenced, should we perhaps take the maximum of the mtimes on the link and on its referent? Otherwise, it seems like you could modify the referent and the code here wouldn't realize anything had changed (and thereby allow use of a stale cached tarball).

src/python/dxpy/app_builder.py
```diff
@@ -212,58 +161,199 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False):
     if os.path.exists(resources_dir) and len(os.listdir(resources_dir)) > 0:
         target_folder = applet_spec['folder'] if 'folder' in applet_spec else folder
 
-        # Optimistically look for a resource bundle with the same
-        # contents, and reuse it if possible. The resource bundle
-        # carries a property 'resource_bundle_checksum' that indicates
-        # the checksum; the way in which the checksum is computed is
-        # given in the documentation of _directory_checksum.
+        # Copy the code previously in _directory_checksum here and we will create
```
psung: Since you removed _directory_checksum, you can remove its mention here. However, please preserve the comment that used to be in that method that specified the way in which the checksum is computed.

src/python/test/test_dxclient.py
```diff
@@ -6210,6 +6210,260 @@ def test_upload_resources(self):
...
         self.assertEqual(resources_file_describe['folder'], '/subfolder')
 
+    def _build_check_resources(self, app_dir, args=""):
+        # create applet and get the resource_file id
+        new_applet = json.loads(run("dx build -f --json " + args + " " + app_dir))
+        applet_describe = dxpy.api.applet_describe(new_applet["id"])
+        resources_file = applet_describe['runSpec']['bundledDepends'][0]['id']['$dnanexus_link']
+        id1 = dxpy.api.file_describe(resources_file)['id']
+
+        # download resources tar and extract it and check lib/libz.so is copied to resources/libz.so
```
psung: Is this comment current? I don't see any assertion to this effect, if that's what you meant to say.

src/python/dxpy/app_builder.py
johnwallace123: Even pathologically large, it should be avoided.  Will refactor using python's tarfile library and avoid any calls to the command line.

src/python/dxpy/app_builder.py
johnwallace123: Will do.  I'll make sure to add this to the test cases.

src/python/dxpy/app_builder.py
johnwallace123: Correct; the idea is to transfer links that would break; this would fall under case 8.  I agree that it's a pathological example to have a link like that, but nothing stops someone from doing something silly like that.

I agree that it's a lot of trouble to protect people from this pathological use case, but I don't think there's a huge computational burden.  It would be my preference to fail early, when possible, rather than failing when trying to reference the link.  Should we put some debugging statements for when links are dereferenced (and if so, what's the "DNAnexus way" of doing so)?

src/python/dxpy/app_builder.py
johnwallace123: I'm not sure there's anything wrong with using the max of the mtimes, but I think because the checksum also includes the filename, it should cover that case.  If you change the name of the link but not the target, the bundle would get refreshed, and if you changed the destination of the link, the mtime of the target would change, causing the bundle to be refreshed.

@manu148 Can we devise some test cases to add to the test_dxclient that would exercise this?  We should add them to the test_upload_resources_advanced.

johnwallace123: After trying to devise some test cases, I think the desired behavior is to use the mtime of the target (here, os.stat follows all of the links in the chain).  I could think of the following cases:

1. If you modify the referent file, the target's mtime will change, causing a refresh.
2. If you change the name of the link, you'll get a refresh because the filename is part of the hash.  Also, the mtime of the containing directory will change.
3. If you change the target of the link to a different file, the target's mtime will change.  Also, the mtime of the containing directory will change.
4. If there is a link to another link which changes in such a way as to point to the same file, we should NOT refresh, as the interceding links have no relevance on the content.

Additionally in case (4), the max of the mtime of the link and the target would not have changed either, making the two approaches equivalent.

src/python/dxpy/app_builder.py
johnwallace123: When writing the test case, it appears that the current master fails when building an app from a relative directory (not related to this branch).  I've added the test case, and when that issue is resolved can be added here.

johnwallace123:
@psung: I've updated the branch with a commit that incorporates the majority of your comments.  At your convenience, could you take a look and see if you have any further comments.  If you approve, I'm happy to squash into a single commit for merging.

johnwallace123:
@psung: I've updated this branch to solve another issue that we've had with permissions in the resource bundle causing issues when unpacked, as seen in ticket 1163.  This fix depends on the solution in fd1b39b, so I've put it in this PR to fix two issues that we have simultaneously.

src/python/dxpy/app_builder.py
psung: Sounds good to me. Thanks.

src/python/dxpy/app_builder.py
```diff
@@ -161,14 +195,44 @@ def upload_resources(src_dir, project=None, folder='/', ensure_upload=False, for
...
+        # /foo \0 MODE \0 MTIME \0
+        # ...
+        #
+        # where there is one entry for each directory or file (order is
+        # specified below), followed by a numeric representation of the
+        # mode, and the mtime in milliseconds since the epoch.
+        #
+        # Note when looking at a link, unless --force-symlinks is specified, 
+        # and the link is to be dereferenced the mtime is taken to be the 
+        # maximum of the mtime of the link itself and of the link's target.
```
psung: You didn't actually implement this, right?

johnwallace123:
Phil, dc66fa4 should incorporate all changes from https://github.com/johnwallace123/dx-toolkit/pull/1.  

If you'd like me to squash and rebase everything, let me know.  If you'd prefer to keep the history of the major commits, I hope everything looks good enough for a merge.

Thanks!

psung:
Thanks, John! It looks good to me and I've squash merged it as 6f0e72e.

PR #143, opened 2016-04-26
PTFM-13142 dx get enhancements

storozhilov-dnanexus:
Hello @psung,

Following points from PTFM-13142 have been implemented:

> Show useful information to stdout, such as that you are creating a new folder locally named after the applet, etc. instead of the currently confusing tar output.

I print following lines to stdout on applet/application fetching:

```
Creating "..." output directory
Downloading applet data
Unpacking resources
```

> Enforce a particular order within input spec keys, for example name, label, help, class, patterns, optional, default, choices, suggestions, group

I use alphabetical ordering

> Write patterns in the same line, i.e. ["\*.fa", "\*.fasta"]

This has been implemented using regexp.

A respective auto-testing has been implemented as well in scope of `TestDXGetExecutables.test_get_applet` test


src/python/dxpy/scripts/dx.py
```diff
@@ -1737,6 +1737,7 @@ def get_output_path(obj_name, obj_class, args):
         os.unlink(output_path)
     # Here, output_path either points to a directory or a nonexistent path
     if not os.path.exists(output_path):
+        print('Creating "{f}" output directory'.format(f=output_path))
```
psung: Please direct these to stderr so we can keep the output clean (I don't think the output of this command is meant to be parseable necessarily but it might become so in the future).

src/python/dxpy/utils/app_unbuilder.py
```diff
@@ -46,6 +47,21 @@ def _recursive_cleanup(foo):
             if val == "" or val == [] or val == {}:
                 del foo[key]
 
+def flatten_patterns(metadata):
```
psung: Can you please add a unit test of this method as well? It would be good to verify what it does on a more or less complete JSON string.

src/python/dxpy/utils/app_unbuilder.py
```diff
@@ -46,6 +47,21 @@ def _recursive_cleanup(foo):
...
+    """
+    Flattens string representations of 'inputSpec.patterns' and
+    'outputSpec.patterns' arrays using regular expressions.
+
+    :param metadata: Executable metadata
+    :type metadata: str
+    """
+
+    result = re.sub("\"patterns\": \\[\r?\n\\s*", "\"patterns\": [", metadata, flags=re.MULTILINE)
```
psung: You can use single quotes to quote the string here, so as to avoid having to escape all the interior quotes.

src/python/dxpy/utils/app_unbuilder.py
```diff
@@ -94,7 +110,8 @@ def dump_executable(executable, destination_directory, omit_resources=False, des
                     handler_id = handler.get_id()
                     fname = "resources/%s.tar.gz" % (handler_id)
                     download_dxfile(handler_id, fname)
-                    subprocess.check_call(["tar", "-C", "resources", "-zxvf", fname], shell=False)
+                    print("Unpacking resources")
+                    subprocess.check_call(["tar", "-C", "resources", "-zxf", fname], stdout=subprocess.PIPE, shell=False)
```
psung: Was the idea just to suppress the "tar" output here? According to the documentation, using PIPE in conjunction with check_call should be avoided because it can deadlock the process (since you have no mechanism to consume stdout while the program is running). I think you can omit this since "tar" shouldn't be printing anything in normal operation anyway.

src/python/dxpy/utils/app_unbuilder.py
```diff
@@ -144,7 +161,7 @@ def dump_executable(executable, destination_directory, omit_resources=False, des
 
         # Write dxapp.json, Readme.md, and Readme.developer.md
         with open("dxapp.json", "w") as f:
-            f.write(json.dumps(dxapp_json, sort_keys=False, indent=2, separators=(',', ': ')))
+            f.write(flatten_patterns(json.dumps(dxapp_json, sort_keys=True, indent=2, separators=(',', ': '))))
```
psung: The idea here is to print the items in the order that the user is likely to want to read them. dx-app-wizard implements something similar by using an OrderedDict and setting fields in the following order:

```
name
title
summary
dxapi
version
...
```

Have a look at that file: https://github.com/dnanexus/dx-toolkit/blob/master/src/python/scripts/dx-app-wizard to get the keys in an order you can emulate.

src/python/dxpy/scripts/dx.py
storozhilov-dnanexus: Resolved

src/python/dxpy/utils/app_unbuilder.py
storozhilov-dnanexus: Resolved

src/python/dxpy/utils/app_unbuilder.py
storozhilov-dnanexus: Resolved

src/python/dxpy/utils/app_unbuilder.py
storozhilov-dnanexus: Resolved

src/python/dxpy/utils/app_unbuilder.py
storozhilov-dnanexus: Resolved

src/python/dxpy/utils/app_unbuilder.py
```diff
@@ -122,6 +123,23 @@ def dump_executable(executable, destination_directory, omit_resources=False, des
         for dep in deps_to_remove:
             dxapp_json["runSpec"]["bundledDepends"].remove(dep)
 
+        # Ordering input/output spec keys
+        ordered_spec_keys = "name", "label", "help", "class", "type", "patterns", "optional", "default", "choices", "suggestions", "group"
```
psung: Please wrap this line to no more than 119 characters.

psung:
LGTM (looks good to me), nice work! Merged as 82fd76c

PR #144, opened 2016-04-28
PTFM-17818 Allow user to specify all --debug-on options at once in dx run

ivoloshanenko:
Allow user to specify all --debug-on options at once in dx run
https://dnanexus.jira.com/browse/PTFM-17818

### DIFF
--- a/src/python/test/test_dxclient.py
+++ b/src/python/test/test_dxclient.py
@@ -878,6 +878,33 @@ class TestDXClient(DXTestCase):
             dx.setwinsize(20, 90)
             dx.expect("dnanexus@", timeout=1200)
 
+    @unittest.skipUnless(testutil.TEST_RUN_JOBS, "Skipping test that would run jobs")
+    def test_dx_run_debug_on_all(self):
+        with self.configure_ssh() as wd:
+            crash_applet = dxpy.api.applet_new(dict(name="crash",
+                                                    runSpec={"code": "exit 5", "interpreter": "bash",
+                                                             "execDepends": [{"name": "dx-toolkit"}]},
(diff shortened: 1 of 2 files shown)


psung:
Thanks @ivoloshanenko!

@nicolasbockg, would you care to review this? Suggestions for the interface/usage (as well as its implementation) are welcome.

nicolasbockg:
LGTM, the implementation now provides a catch-all option, which is simple to use.

src/python/test/test_dxclient.py
```diff
@@ -878,6 +878,33 @@ def test_dx_run_debug_on(self):
             dx.setwinsize(20, 90)
             dx.expect("dnanexus@", timeout=1200)
 
+    @unittest.skipUnless(testutil.TEST_RUN_JOBS, "Skipping test that would run jobs")
+    def test_dx_run_debug_on_all(self):
+        with self.configure_ssh() as wd:
```
nicolasbockg: Given that there is a very similar test just above this one, that also connects to the worker using ssh, we don't really need to test for that again. 
What we really need to verify is that the 'debug-on' option for this job is set to '['AppError', 'AppInternalError', 'ExecutionError']', when the 'All' option is given. Please consider removing the ssh configuration part, and running the command 'dx run ...' just the same, and then making a call to the api to describe the job with the 'job_id' you obtain. The description will contain a field called 'debug' which has the specified debug options.  Thanks!

ivoloshanenko:
Test file updated

nicolasbockg:
Thanks for the changes to the test, it looks good.

PR #145, opened 2016-05-06
PTFM-18056 java incorrect paginated search behavior

storozhilov-dnanexus:
Hello @psung ,
Here is the test and a respective fix - would be nice if you can look on that.
Thank you very much!

### DIFF
--- a/src/java/src/main/java/com/dnanexus/DXSearch.java
+++ b/src/java/src/main/java/com/dnanexus/DXSearch.java
@@ -900,11 +900,11 @@ public final class DXSearch {
         /**
          * Iterator implementation for findDataObjects results.
          */
-        private class ResultIterator
+        public class ResultIterator
                 extends
                 PaginatingFindResultIterator<T, FindDataObjectsRequest, FindDataObjectsResultPage> {
 
-            public ResultIterator() {
+            private ResultIterator() {
                 super(baseQuery);
             }
 
@@ -1935,6 +1935,7 @@ public final class DXSearch {
         private Q query;
         private P currentPage;
         private int nextResultIndex = 0;
+        private int currentPageNo = 0;
 
         /**
          * Initializes the iterator with the specified initial query.
@@ -1945,6 +1946,7 @@ public final class DXSearch {
         private PaginatingFindResultIterator(Q initialQuery) {
             this.query = initialQuery;
             this.currentPage = issueQuery(initialQuery);
+            this.currentPageNo++;
         }
 
         /**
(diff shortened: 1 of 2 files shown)


src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -900,11 +909,11 @@ public int size() {
         /**
          * Iterator implementation for findDataObjects results.
          */
-        private class ResultIterator
+        public class ResultIterator
```
psung: Please make this default-visibility and label it "@VisibleForTesting". That indicates it's being made available for tests and doesn't make it part of our public API.

Same goes for pageNo later.

src/java/src/test/java/com/dnanexus/DXSearchTest.java
```diff
@@ -585,6 +586,17 @@ public void testFindDataObjectsWithPaging() {
                 ImmutableList.copyOf(DXSearch.findDataObjects().inProject(testProject)
                         .nameMatchesGlob("foo*").withClassRecord().execute(3));
         Assert.assertEquals(outputRecords, outputRecordStreamWithPaging);
+
+        // Checking an amount of fetched pages
+        DXSearch.FindDataObjectsResult<DXRecord> findResult =
+                DXSearch.findDataObjects().inProject(testProject).nameMatchesGlob("foo*").withClassRecord().execute(3);
+        Iterator<DXRecord> iter = findResult.iterator();
+        List<DXRecord> iterativelyFetchedRecords = Lists.newArrayList();
```
psung: You should be able to automatically populate the list from the iterable:

```
... = Lists.newArrayList(findResult.iterator());
```

storozhilov-dnanexus: Fixed, thank you - very good to know!

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: I tried to make **ResultIterator** private and labeled as *@VisibleForTesting*, but compiler claimed about that:

```
        @VisibleForTesting
        private class ResultIterator
...
[INFO] Compilation failure
/vagrant/src/java/src/test/java/com/dnanexus/DXSearchTest.java:[596,74] error: FindDataObjectsResult.ResultIterator has private access in FindDataObjectsResult
```

If to leave **ResultIterator** public and to make **int pageNo()** method private compiler could not find a respective symbol:

```
        @VisibleForTesting
        private int pageNo() {
            return this.currentPageNo;
        }
...
[INFO] Compilation failure
/vagrant/src/java/src/test/java/com/dnanexus/DXSearchTest.java:[596,95] error: cannot find symbol
```
I tried to label a **PaginatingFindResultIterator** as *@VisibleForTesting*, but it did not help.

Will look to that tomorrow.

psung: Not private, but default-visibility:

```
@VisibleForTesting
class ResultIterator {

...

@VisibleForTesting
int pageNo() {
```

(The VisibleForTesting doesn't actually mean anything to the compiler-- it's just a convention for humans to understand that the visibility of the method/class was intentionally widened to allow tests to peek at the internals.)

storozhilov-dnanexus: Fixed. Thank you!

PR #146, opened 2016-05-10
PTFM-17957 Allow users to "dx uninstall" apps by ID to which they no longer have access

ivoloshanenko:
https://dnanexus.jira.com/browse/PTFM-17957

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -2514,7 +2514,15 @@ def install(args):
 def uninstall(args):
     app_desc = get_app_from_path(args.app)
     if app_desc is None:
-        parser.exit(1, 'Could not find the app\n')
+        try:
+            dxpy.api.app_uninstall(args.app)
+            print('Uninstalled application with id: ' + args.app)
+        except:
+            try:
+                dxpy.api.app_uninstall('app-' + args.app)
+                print('Uninstalled application with name: ' + args.app)
+            except:
+                err_exit()
     else:
         try:
             dxpy.api.app_uninstall(app_desc['id'])
--- a/src/python/test/test_dxclient.py
+++ b/src/python/test/test_dxclient.py
@@ -7489,6 +7489,43 @@ class TestDXGetExecutables(DXTestCase):
                     break
             self.assertTrue(seenResources)
 
+    @unittest.skipUnless(testutil.TEST_ENV, 'skipping test that would clobber your local environment')
+    @unittest.skipUnless(testutil.TEST_ISOLATED_ENV, 'skipping test that would create apps')
+    @unittest.skipUnless(testutil.TEST_MULTIPLE_USERS, 'skipping test that would require another user')
+    def test_uninstall_app(self):
+        second = json.loads(os.environ['DXTEST_SECOND_USER'])
+        second_user_id = second['user']
+        authorized_users = [second_user_id]
+        name = 'uninstall_test'
+        app_spec = {
+            "name": name,
+            "title": name,
+            "dxapi": "1.0.0",
+            "runSpec": {"file": "code.py", "interpreter": "python2.7"},
+            "inputSpec": [{"name": "in1", "class": "file"}],
+            "outputSpec": [{"name": "out1", "class": "file"}],
+            "description": "Description\n",
+            "developerNotes": "Developer notes\n",
+            "authorizedUsers": authorized_users,
+            "openSource": True,
+            "version": "0.0.1"
+            }
+        app_dir = self.write_app_directory(name, json.dumps(app_spec), "code.py", code_content="import os\n")
+        os.mkdir(os.path.join(app_dir, "resources"))
+        with open(os.path.join(app_dir, "resources", "resources_file"), 'w') as f:
+            f.write('content\n')
+        build_cmd = "dx build --create-app --json --publish "
+        app_json = json.loads(run(build_cmd + app_dir))
+        self.assertEqual(app_json['name'], name)
+        run("dx install " + name, env=as_second_user())
+        output = json.loads(run("dx find apps --installed --json", env=as_second_user()))
+        self.assertIn(name, [x['describe']['name'] for x in output])
+        dxpy.api.app_remove_authorized_users(app_json['id'],
+                                                      input_params={'authorizedUsers': list(authorized_users)})
+        output = dxpy.api.app_describe(app_json['id'])
+        self.assertNotIn(second_user_id, output['authorizedUsers'])
+        output = json.loads(run("dx uninstall %s" % name, env=as_second_user()))
+        self.assertEqual(name, output['name'])
 
 class TestDXBuildReportHtml(unittest.TestCase):
     js = "console.log('javascript');"


ivoloshanenko:
When user trying uninstall application via 'dx uninstall', first of all we trying send Describe command for this app to api server. If access was restricted, then execution failed on this step and this is right behavior. User can uninstall app via 'dx api <> uninstall' or with this patch, providing appropriate Id to 'dx uninstall <>'

src/python/dxpy/scripts/dx.py
```diff
@@ -2514,7 +2514,13 @@ def install(args):
 def uninstall(args):
     app_desc = get_app_from_path(args.app)
     if app_desc is None:
-        parser.exit(1, 'Could not find the app\n')
+        #in case we lost permission for this app, try to remove it by Id.
```
nicolasbockg: Given that here we are trying to remove the app using its name, please consider rephrasing this comment as:
'# In case the user lost access to the app, try to remove it using the app's name'

src/python/dxpy/scripts/dx.py
```diff
@@ -2514,7 +2514,13 @@ def install(args):
...
+        try:
+            dxpy.api.app_uninstall(args.app)
+            print('Uninstalled application with id: ' + args.app)
```
nicolasbockg: I think this print statement can be the similar to the original:
`print('Uninstalled the ' + args.app + ' app')`

src/python/dxpy/scripts/dx.py
```diff
@@ -2514,7 +2514,13 @@ def install(args):
...
+        except Exception as error:
+            parser.exit(1, 'Could not find the app! If application still in list, '
+                        'please try uninstall it providing appropriate Id\n' + str(error))
```
nicolasbockg: Please consider rephrasing this as:
"Could not find the app.\n If the app still appears on the list, please try to uninstall it by providing the appropriate Id" + str(error)

However, if the user does not have access to the app, would he still be able to obtain the app id with `dx describe appname`? If not, then we might need a different message.

nicolasbockg:
I think we should also add a couple tests, possibly in this suite: class TestDXGetExecutables, where an app is created by a second user, with and without allowing the first user access, that way we can test the uninstall procedure in dx. You can look at the _test_get_app for ideas.

src/python/dxpy/scripts/dx.py
ivoloshanenko: In this case we trying to remove exactly by Id.

ivoloshanenko: In this case we trying to remove exactly by Id.

src/python/dxpy/scripts/dx.py
```diff
@@ -2513,8 +2513,17 @@ def install(args):
 
 def uninstall(args):
     app_desc = get_app_from_path(args.app)
+    print("-- uninstalling application --")
```
nicolasbockg: This print statement can be removed.

nicolasbockg:
The test looks good, thanks Igor! Just the minor change I suggested and then I can merge it into master.

ivoloshanenko:
fixed

nicolasbockg:
Merging into master, thanks Igor!

src/python/dxpy/scripts/dx.py
```diff
@@ -2514,7 +2514,15 @@ def install(args):
 def uninstall(args):
     app_desc = get_app_from_path(args.app)
     if app_desc is None:
-        parser.exit(1, 'Could not find the app\n')
+        try:
```
nicolasbockg: From @ptn24:
Please consider using the `try_call` function here.

src/python/dxpy/scripts/dx.py
```diff
@@ -2514,7 +2514,15 @@ def install(args):
...
+            dxpy.api.app_uninstall(args.app)
```
nicolasbockg: From @ptn24 : Rather than simply invoking /app-x/uninstall here, perhaps we can verify whether the specified app is present in the appsInstalled field of the self describe response, and then only uninstall if present. Otherwise, we can raise an error and exit, as before. Does this make sense at all?

src/python/dxpy/scripts/dx.py
```diff
@@ -2514,7 +2514,15 @@ def install(args):
...
+            print('Uninstalled application with id: ' + args.app)
```
nicolasbockg: From @ptn24:
Please keep the error messages consistent. For example, we refer to apps as "apps" to end users, not as "applications". Furthermore, to preserve the syntax of the existing error message, I think it will be better to use the error messages
`print("Uninstalled the {id} app".format(id=args.app))`
and
`print("Uninstalled the {name} app".format(name=args.app))`
for the two code execution flows.

src/python/test/test_dxclient.py
```diff
@@ -7489,6 +7489,43 @@ def test_get_app_omit_resources(self):
                     break
             self.assertTrue(seenResources)
 
+    @unittest.skipUnless(testutil.TEST_ENV, 'skipping test that would clobber your local environment')
```
nicolasbockg: from @ptn24: Why is this necessary?
from me: We can remove this line.

src/python/test/test_dxclient.py
```diff
@@ -7489,6 +7489,43 @@ def test_get_app_omit_resources(self):
...
+    @unittest.skipUnless(testutil.TEST_ISOLATED_ENV, 'skipping test that would create apps')
+    @unittest.skipUnless(testutil.TEST_MULTIPLE_USERS, 'skipping test that would require another user')
```
nicolasbockg: from @ptn24: Note that we can combine these into something like:
`@unittest.skipUnless(FOO and BAR, 'skipping test that would foo and bar')`

src/python/test/test_dxclient.py
```diff
@@ -7489,6 +7489,43 @@ def test_get_app_omit_resources(self):
...
+    def test_uninstall_app(self):
+        second = json.loads(os.environ['DXTEST_SECOND_USER'])
+        second_user_id = second['user']
```
nicolasbockg: From @ptn24: This can go in one line, no?

src/python/test/test_dxclient.py
```diff
@@ -7489,6 +7489,43 @@ def test_get_app_omit_resources(self):
...
+        authorized_users = [second_user_id]
+        name = 'uninstall_test'
+        app_spec = {
+            "name": name,
+            "title": name,
+            "dxapi": "1.0.0",
+            "runSpec": {"file": "code.py", "interpreter": "python2.7"},
+            "inputSpec": [{"name": "in1", "class": "file"}],
+            "outputSpec": [{"name": "out1", "class": "file"}],
```
nicolasbockg: From @ptn24:
Are inputSpec and outputSpec necessary? That is, we can specify empty lists for the input and output specifications, respectively. In general, I prefer to keep unnecessary information out of the tests to improve readability and maintainability.

src/python/test/test_dxclient.py
```diff
@@ -7489,6 +7489,43 @@ def test_get_app_omit_resources(self):
...
+            "description": "Description\n",
+            "developerNotes": "Developer notes\n",
+            "authorizedUsers": authorized_users,
+            "openSource": True,
+            "version": "0.0.1"
+            }
+        app_dir = self.write_app_directory(name, json.dumps(app_spec), "code.py", code_content="import os\n")
+        os.mkdir(os.path.join(app_dir, "resources"))
```
nicolasbockg: From @ptn24: Minor: Is the resources dir necessary here?

src/python/test/test_dxclient.py
```diff
@@ -7489,6 +7489,43 @@ def test_get_app_omit_resources(self):
...
+        with open(os.path.join(app_dir, "resources", "resources_file"), 'w') as f:
+            f.write('content\n')
+        build_cmd = "dx build --create-app --json --publish "
+        app_json = json.loads(run(build_cmd + app_dir))
+        self.assertEqual(app_json['name'], name)
+        run("dx install " + name, env=as_second_user())
+        output = json.loads(run("dx find apps --installed --json", env=as_second_user()))
+        self.assertIn(name, [x['describe']['name'] for x in output])
+        dxpy.api.app_remove_authorized_users(app_json['id'],
```
nicolasbockg: From @ptn24: Please wrap properly here.

src/python/test/test_dxclient.py
```diff
@@ -7489,6 +7489,43 @@ def test_get_app_omit_resources(self):
...
+                                                      input_params={'authorizedUsers': list(authorized_users)})
+        output = dxpy.api.app_describe(app_json['id'])
+        self.assertNotIn(second_user_id, output['authorizedUsers'])
+        output = json.loads(run("dx uninstall %s" % name, env=as_second_user()))
```
nicolasbockg: From @ptn24 : 
To effectively repro the bug described in PTFM-17957, the important things to assert here, after being removed from the list of authorized users, are:
1. The `appsInstalled` field in the response of /user-x/describe contains the app before uninstalling.
2. The `appsInstalled` field in the response of /user-x/describe does not contain the app after uninstalling.

PR #147, opened 2016-05-13
Escape file name with spaces for dx-unpack

adasgt:
@psung please review.

If 'bundledDepends' refers to a file name containing un-escaped spaces, dx-unpack fails as it reads each string in the file name as an argument. The error looks like as follows when a file name is "my asset test.tar.gz":

'
_Downloading bundled file my asset test.tar.gz
>>> Unpacking my to /
 *   dx_helpers.sh, line 20: my does not exist
  File "/usr/bin/dx-prepare-exec-env", line 45, in <module>
    installer.install()
  File "/usr/share/dnanexus/lib/python2.7/site-packages/dxpy/utils/exec_utils.py", line 395, in install
    self._install_dep_bundle(dep_group["deps"][0])
  File "/usr/share/dnanexus/lib/python2.7/site-packages/dxpy/utils/exec_utils.py", line 388, in _install_dep_bundle
    self.run("dx-unpack '{}'".format(bundle["name"]))
  File "/usr/share/dnanexus/lib/python2.7/site-packages/dxpy/utils/exec_utils.py", line 368, in run
    subprocess.check_call(cmd, shell=True, stdout=log_fh, stderr=log_fh)
  File "/usr/lib/python2.7/subprocess.py", line 511, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command 'dx-unpack 'my asset test.tar.gz'' returned non-zero exit status 1_
'

So escaped the file name in subprocess call.

### DIFF
--- a/bin/dx-unpack
+++ b/bin/dx-unpack
@@ -38,7 +38,7 @@ while [[ $# > 0 ]]; do
         output="$2"
         shift 2
     else
-        A="$A $1"
+        A="$A \"$1\""
         shift
     fi
 done
--- a/bin/dx-unpack
+++ b/bin/dx-unpack
@@ -38,7 +38,7 @@ while [[ $# > 0 ]]; do
         output="$2"
         shift 2
     else
-        A="$A \"$1\""
+        A="$A '${1//\'/\'\\\'\'}'"
         shift
     fi
 done
--- a/src/python/dxpy/utils/exec_utils.py
+++ b/src/python/dxpy/utils/exec_utils.py
@@ -23,6 +23,7 @@ from __future__ import print_function, unicode_literals, division, absolute_impo
 import os, sys, json, re, collections, logging, argparse, string, itertools, subprocess, tempfile
 from functools import wraps
 from collections import namedtuple
+import pipes
 
 import dxpy
 from ..compat import USING_PYTHON2, open
@@ -385,7 +386,7 @@ class DXExecDependencyInstaller(object):
         if bundle["id"].get("$dnanexus_link", "").startswith("file-"):
             self.log("Downloading bundled file {name}".format(**bundle))
             dxpy.download_dxfile(bundle["id"], bundle["name"])
-            self.run("dx-unpack '{}'".format(re.escape(bundle["name"])))
+            self.run("dx-unpack {}".format(pipes.quote(bundle["name"])))
         else:
             self.log('Skipping bundled dependency "{name}" because it does not refer to a file'.format(**bundle))
 
--- a/src/python/dxpy/utils/exec_utils.py
+++ b/src/python/dxpy/utils/exec_utils.py
@@ -385,7 +385,7 @@ class DXExecDependencyInstaller(object):
         if bundle["id"].get("$dnanexus_link", "").startswith("file-"):
             self.log("Downloading bundled file {name}".format(**bundle))
             dxpy.download_dxfile(bundle["id"], bundle["name"])
-            self.run("dx-unpack '{}'".format(bundle["name"]))
+            self.run("dx-unpack '{}'".format(re.escape(bundle["name"])))
         else:
             self.log('Skipping bundled dependency "{name}" because it does not refer to a file'.format(**bundle))
 
--- a/src/python/scripts/dx-fetch-bundled-depends
+++ b/src/python/scripts/dx-fetch-bundled-depends
@@ -20,7 +20,7 @@
 Downloads the contents of runSpec.bundledDepends of a job running in the execution environment.
 '''
 
-import os, sys, argparse, subprocess, re
+import os, sys, argparse, subprocess, pipes
 import dxpy
 
 parser = argparse.ArgumentParser(description=__doc__)
@@ -41,4 +41,4 @@ if 'DX_JOB_ID' in os.environ:
             if dep['id']['$dnanexus_link'].startswith('file-'):
                 print "*** Downloading bundled file", dep['name']
                 dxpy.download_dxfile(dep['id'], dep['name'])
-                subprocess.check_call(['dx-unpack', re.escape(dep['name'])])
+                subprocess.check_call(['dx-unpack', pipes.quote(dep['name'])])
--- a/src/python/scripts/dx-fetch-bundled-depends
+++ b/src/python/scripts/dx-fetch-bundled-depends
@@ -20,7 +20,7 @@
 Downloads the contents of runSpec.bundledDepends of a job running in the execution environment.
 '''
 
-import os, sys, argparse, subprocess
+import os, sys, argparse, subprocess, re
 import dxpy
 
 parser = argparse.ArgumentParser(description=__doc__)
@@ -41,4 +41,4 @@ if 'DX_JOB_ID' in os.environ:
             if dep['id']['$dnanexus_link'].startswith('file-'):
                 print "*** Downloading bundled file", dep['name']
                 dxpy.download_dxfile(dep['id'], dep['name'])
-                subprocess.check_call(['dx-unpack', dep['name']])
+                subprocess.check_call(['dx-unpack', re.escape(dep['name'])])
--- a/src/python/scripts/dx-fetch-bundled-depends
+++ b/src/python/scripts/dx-fetch-bundled-depends
@@ -20,7 +20,7 @@
 Downloads the contents of runSpec.bundledDepends of a job running in the execution environment.
 '''
 
-import os, sys, argparse, subprocess, pipes
+import os, sys, argparse, subprocess
 import dxpy
 
 parser = argparse.ArgumentParser(description=__doc__)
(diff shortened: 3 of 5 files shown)


src/python/dxpy/utils/exec_utils.py
```diff
@@ -385,7 +385,7 @@ def _install_dep_bundle(self, bundle):
         if bundle["id"].get("$dnanexus_link", "").startswith("file-"):
             self.log("Downloading bundled file {name}".format(**bundle))
             dxpy.download_dxfile(bundle["id"], bundle["name"])
-            self.run("dx-unpack '{}'".format(bundle["name"]))
+            self.run("dx-unpack '{}'".format(re.escape(bundle["name"])))
```
psung: re.escape is not the right thing to use for shell escaping-- see pipes.quote instead (you can then remove the quotes here).

src/python/scripts/dx-fetch-bundled-depends
```diff
@@ -41,4 +41,4 @@ if 'DX_JOB_ID' in os.environ:
             if dep['id']['$dnanexus_link'].startswith('file-'):
                 print "*** Downloading bundled file", dep['name']
                 dxpy.download_dxfile(dep['id'], dep['name'])
-                subprocess.check_call(['dx-unpack', dep['name']])
+                subprocess.check_call(['dx-unpack', re.escape(dep['name'])])
```
psung: Is the escaping even needed here? The command has already been tokenized.

psung:
Thanks! Can you please add a unit test?

src/python/scripts/dx-fetch-bundled-depends
adasgt: I tried testing it out and it throws exception if not escaped. Here is a snippet of the test:

>>> f_name="space in name.tar.gz"
>>> subprocess.check_call(['dx-unpack', f_name])
>>> Unpacking space to /home/test/tmp
 *   dx_helpers.sh, line 20: space does not exist
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['dx-unpack', 'space in name.tar.gz']' returned non-zero exit status 1

psung: I think that's because dx-unpack itself has a bug in its treatment of the filename. We should wait for that fix to land so as not to force callers of dx-unpack to escape the command line.

src/python/dxpy/utils/exec_utils.py
adasgt: Done, thanks!

adasgt:
Added a positive test: 
./src/python/test/test_dxclient.py TestDXClientRun.test_bundledDepends_name_with_whitespaces

adasgt:
@psung please review.

src/python/test/test_dxclient.py
```diff
@@ -2728,6 +2728,33 @@ def main():
         shell.close()
         self.assertEqual(3, shell.exitstatus)
 
+    @unittest.skipUnless(testutil.TEST_RUN_JOBS, 'skipping tests that would run jobs')
+    def test_bundledDepends_name_with_whitespaces(self):
+        # upload a tar.gz file with spaces in its name
+        bundle_name = "test bundle with spaces.tar.gz"
```
psung: Since the content of this file isn't valid tar/gzip data (I presume), wouldn't dx-unpack fail even if it were fixed to handle filenames correctly?

src/python/dxpy/utils/exec_utils.py
```diff
@@ -23,6 +23,7 @@
 import os, sys, json, re, collections, logging, argparse, string, itertools, subprocess, tempfile
 from functools import wraps
 from collections import namedtuple
+import pipes
```
psung: I think the change in this file is appropriate, but I don't know about the other two files. Do you want to wait until you have a fix for dx-unpack, or do you want to just land this part now?

src/python/scripts/dx-fetch-bundled-depends
```diff
@@ -41,4 +41,4 @@ if 'DX_JOB_ID' in os.environ:
             if dep['id']['$dnanexus_link'].startswith('file-'):
                 print "*** Downloading bundled file", dep['name']
                 dxpy.download_dxfile(dep['id'], dep['name'])
-                subprocess.check_call(['dx-unpack', dep['name']])
+                subprocess.check_call(['dx-unpack', pipes.quote(dep['name'])])
```
psung: As we discussed, I think this is something that should be fixed in dx-unpack itself.

src/python/dxpy/utils/exec_utils.py
adasgt: No. Added fix for the dx-unpack. Please check.

src/python/scripts/dx-fetch-bundled-depends
adasgt: Reverted. Fixed dx-unpack, please review.

src/python/test/test_dxclient.py
adasgt: Fixed. Created a valid tar.gz file.

adasgt:
@psung please review.

bin/dx-unpack
```diff
@@ -38,7 +38,7 @@ while [[ $# > 0 ]]; do
         output="$2"
         shift 2
     else
-        A="$A $1"
+        A="$A \"$1\""
```
psung: I am not sure this is the right fix. The addition of the quotes seems to introduce another problem when the filename itself contains quotes:

```
$ echo "a" > \"foo\ bar\"
$ dx-unpack \"foo\ bar\"
>>> Unpacking foo to /some/path
 *   dx_helpers.sh, line 20: foo does not exist
```

I think the "eval set" below is already supposed to fix up any problems with quoting as a result of this first phase. So I suspect any problems are downstream of that... perhaps in the call to unpack the arguments are being improperly tokenized (again)? I wonder if another judicious use of "eval set" could fix that...

src/python/test/test_dxclient.py
```diff
@@ -2728,6 +2728,40 @@ def main():
         shell.close()
         self.assertEqual(3, shell.exitstatus)
 
+    @unittest.skipUnless(testutil.TEST_RUN_JOBS, 'skipping tests that would run jobs')
```
psung: Can you add as well unit tests of dx-unpack that run locally? I imagine that will help you to track down a lot more of these quoting issues locally. In addition to taking possibly minutes (instead of milliseconds), this test has the disadvantage that you won't see it pass until the fixed dx-unpack has actually been redeployed to staging.

bin/dx-unpack
psung: You could also consider inlining the function "unpack" or converting this whole thing to a Python script (don't know if either of those would make things any simpler... but it's something to evaluate to avoid bash nonsense).

adasgt: As per my investigation, the problem is here with quoting of the arguments. 
Fixed it by quoting it twice. 
ex.
dx-unpack "test 'bundle' \"with\" \"$@#^&%()[]{}\" spaces.tar.gz"
>>> Unpacking test 'bundle' "with" "$@#^&%()[]{}" spaces.tar.gz to /home/abhiram/test-concepts/tmp

src/python/test/test_dxclient.py
adasgt: Added a test case. Thanks!

adasgt:
@psung please review.

src/python/test/test_dxclient.py
```diff
@@ -2762,6 +2762,17 @@ def test_bundledDepends_name_with_whitespaces(self):
         applet_job.wait_on_done()
         self.assertEqual(applet_job.describe()['state'], 'done')
 
+    def test_bundledDepends_name_with_specialChars_locally(self):
```
psung: special_chars

src/python/test/test_dxclient.py
```diff
@@ -2762,6 +2762,17 @@ def test_bundledDepends_name_with_whitespaces(self):
...
+        # upload a tar.gz file with spaces, quotes and escape chars in its name
+        bundle_name = "test 'bundle' \"with\" \"$@#^&%()[]{}\" spaces.tar.gz"
+        bundle_tmp_dir = tempfile.mkdtemp()
+        os.mkdir(os.path.join(bundle_tmp_dir, "a"))
+        with open(os.path.join(bundle_tmp_dir, 'a', 'foo.txt'), 'w') as file_in_bundle:
+            file_in_bundle.write('foo\n')
+        subprocess.check_call(['tar', '-czf', os.path.join(bundle_tmp_dir, bundle_name),
+                               '-C', os.path.join(bundle_tmp_dir, 'a'), '.'])
+        subprocess.check_call(["dx-unpack", os.path.join(bundle_tmp_dir, bundle_name)])
```
psung: Does extraction of the tarball leave around files that should be cleaned up afterwards?

psung:
LGTM, thanks!

src/python/test/test_dxclient.py
adasgt: Thanks!
Removed the extracted file.
os.remove(os.path.join(os.getcwd(), 'foo.txt'))

src/python/test/test_dxclient.py
adasgt: Done.

PR #149, opened 2016-05-23
Allow users to "dx uninstall" apps by ID without access

ivoloshanenko:
https://dnanexus.jira.com/browse/PTFM-17957

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -2513,22 +2513,18 @@ def install(args):
 
 def uninstall(args):
     app_desc = get_app_from_path(args.app)
-    if app_desc is None:
-        try:
-            dxpy.api.app_uninstall(args.app)
-            print('Uninstalled application with id: ' + args.app)
-        except:
-            try:
-                dxpy.api.app_uninstall('app-' + args.app)
-                print('Uninstalled application with name: ' + args.app)
-            except:
-                err_exit()
+    if app_desc:
+        try_call(dxpy.api.app_uninstall, app_desc['id'])
     else:
-        try:
-            dxpy.api.app_uninstall(app_desc['id'])
-            print('Uninstalled the ' + app_desc['name'] + ' app')
-        except:
-            err_exit()
+        user_data = dxpy.api.user_describe(dxpy.whoami(), {"fields": {"appsInstalled": True}})
+        if args.app in user_data['appsInstalled']:
+            args.app = 'app-' + args.app
+        if args.app.startswith('app-'):
+            try_call(dxpy.api.app_uninstall, args.app)
+            print('Uninstalled the {app} app'.format(app=args.app))
+        else:
+            parser.exit(1, 'Could not find the app\n')
+
 
 def run_one(args, executable, dest_proj, dest_path, preset_inputs=None, input_name_prefix=None,
             is_the_only_job=True):
--- a/src/python/test/test_dxclient.py
+++ b/src/python/test/test_dxclient.py
@@ -25,6 +25,7 @@ import pipes
 import stat
 import hashlib
 import collections
+import string
 from contextlib import contextmanager
 import pexpect
 import requests
@@ -7463,12 +7464,10 @@ class TestDXGetExecutables(DXTestCaseBuildApps):
                     break
             self.assertTrue(seenResources)
 
-    @unittest.skipUnless(testutil.TEST_ENV, 'skipping test that would clobber your local environment')
-    @unittest.skipUnless(testutil.TEST_ISOLATED_ENV, 'skipping test that would create apps')
-    @unittest.skipUnless(testutil.TEST_MULTIPLE_USERS, 'skipping test that would require another user')
+    @unittest.skipUnless(testutil.TEST_ISOLATED_ENV and testutil.TEST_MULTIPLE_USERS,
+                         'skipping test that would create apps and another user')
     def test_uninstall_app(self):
-        second = json.loads(os.environ['DXTEST_SECOND_USER'])
-        second_user_id = second['user']
+        second_user_id = json.loads(os.environ['DXTEST_SECOND_USER'])['user']
         authorized_users = [second_user_id]
         name = 'uninstall_test'
         app_spec = {
@@ -7476,8 +7475,8 @@ class TestDXGetExecutables(DXTestCaseBuildApps):
             "title": name,
             "dxapi": "1.0.0",
             "runSpec": {"file": "code.py", "interpreter": "python2.7"},
-            "inputSpec": [{"name": "in1", "class": "file"}],
-            "outputSpec": [{"name": "out1", "class": "file"}],
+            "inputSpec": [],
+            "outputSpec": [],
             "description": "Description\n",
             "developerNotes": "Developer notes\n",
             "authorizedUsers": authorized_users,
@@ -7485,22 +7484,41 @@ class TestDXGetExecutables(DXTestCaseBuildApps):
             "version": "0.0.1"
             }
         app_dir = self.write_app_directory(name, json.dumps(app_spec), "code.py", code_content="import os\n")
-        os.mkdir(os.path.join(app_dir, "resources"))
-        with open(os.path.join(app_dir, "resources", "resources_file"), 'w') as f:
-            f.write('content\n')
         build_cmd = "dx build --create-app --json --publish "
         app_json = json.loads(run(build_cmd + app_dir))
         self.assertEqual(app_json['name'], name)
+        # Install and check uninstall by name
         run("dx install " + name, env=as_second_user())
         output = json.loads(run("dx find apps --installed --json", env=as_second_user()))
         self.assertIn(name, [x['describe']['name'] for x in output])
         dxpy.api.app_remove_authorized_users(app_json['id'],
-                                                      input_params={'authorizedUsers': list(authorized_users)})
+                                             input_params={'authorizedUsers': list(authorized_users)})
         output = dxpy.api.app_describe(app_json['id'])
         self.assertNotIn(second_user_id, output['authorizedUsers'])
+        dxpy.api.project_invite(self.project, input_params={'invitee': second_user_id, 'level': 'VIEW'})
+        user_data = json.loads(run('dx describe --json ' + second_user_id, env=as_second_user()))
+        self.assertIn(name, user_data['appsInstalled'])
         run("dx uninstall %s" % name, env=as_second_user())
         output = json.loads(run("dx find apps --installed --json", env=as_second_user()))
         self.assertNotIn(name, [x['describe']['name'] for x in output])
+        user_data = json.loads(run('dx describe --json ' + second_user_id, env=as_second_user()))
+        self.assertNotIn(name, user_data['appsInstalled'])
+        # Install and check uninstall by ID
+        dxpy.api.app_add_authorized_users(app_json['id'],
+                                          input_params={'authorizedUsers': list(authorized_users)})
+        run("dx install " + name, env=as_second_user())
+        dxpy.api.app_remove_authorized_users(app_json['id'],
+                                             input_params={'authorizedUsers': list(authorized_users)})
+        user_data = json.loads(run('dx describe --json ' + second_user_id, env=as_second_user()))
+        self.assertIn(name, user_data['appsInstalled'])
+        run("dx uninstall %s" % app_json['id'], env=as_second_user())
+        user_data = json.loads(run('dx describe --json ' + second_user_id, env=as_second_user()))
+        self.assertNotIn(name, user_data['appsInstalled'])
+        # Check for App not found
+        app_unknown_name = ''.join(random.choice(string.ascii_lowercase) for _ in range(12))
+        with self.assertSubprocessFailure(stderr_regexp='Could not find the app', exit_code=1):
+            run("dx uninstall %s" % app_unknown_name, env=as_second_user())
+        pass
 
 class TestDXBuildReportHtml(unittest.TestCase):
     js = "console.log('javascript');"


src/python/dxpy/scripts/dx.py
```diff
@@ -2516,19 +2516,17 @@ def uninstall(args):
...
             dxpy.api.app_uninstall(args.app)
-            print('Uninstalled application with id: ' + args.app)
+            print('Uninstalled the {id} app'.format(id=args.app))
         except:
-            try:
-                dxpy.api.app_uninstall('app-' + args.app)
-                print('Uninstalled application with name: ' + args.app)
-            except:
-                err_exit()
+            user_data = dxpy.api.user_describe(dxpy.whoami(), {'appsInstalled': True})
```
ptn24: Please front-load the /user-x/describe invocation before attempting any uninstall. Once we verify that either `args.app` or `"app-" + args.app` is in the `appsInstalled` field of the describe response, then we can proceed with the uninstall via /app-x/uninstall (note that this is using app-x, whether that is from `args.app` or from `"app-" + args.app`).

src/python/test/test_dxclient.py
```diff
@@ -7463,44 +7463,44 @@ def test_get_app_omit_resources(self):
...
         self.assertEqual(app_json['name'], name)
         run("dx install " + name, env=as_second_user())
         output = json.loads(run("dx find apps --installed --json", env=as_second_user()))
         self.assertIn(name, [x['describe']['name'] for x in output])
         dxpy.api.app_remove_authorized_users(app_json['id'],
-                                                      input_params={'authorizedUsers': list(authorized_users)})
+                                             input_params={'authorizedUsers': list(authorized_users)})
         output = dxpy.api.app_describe(app_json['id'])
         self.assertNotIn(second_user_id, output['authorizedUsers'])
+        dxpy.api.project_invite(self.project, input_params={'invitee': second_user_id, 'level': 'VIEW'})
```
ptn24: Hm, why is this necessary?

ivoloshanenko: Without this invite, test failed and we getting error:
DXCalledProcessError: Command 'dx describe --json user-000000000000000000000001' returned non-zero exit status 3, stderr:
DXCLIError: At least VIEW permission is required to search within a project, code 401

ptn24: Strange. No project permissions are necessary to describe a user. Can you see what is going on in the describe invocation by prefixing it with "_DX_DEBUG=2 dx describe --json ..."?

psung: dx probably guesses that the thing being described could be the name of a data object and tries to look for it. This is probably something where we have room for improvement, especially when the thing being described is obviously a user, app, etc.

However, the command below is run from the perspective of the second user, and I believe that for that user the project context points to a project for which they don't have VIEW access (since it was created solely for user1). That's a situation that is very likely to cause weird things to happen (especially in "describe"), so some readable project should probably be set explicitly.

ivoloshanenko: with debug:
`
$ _DX_DEBUG=3 dx describe --json user-000000000000000000000001
POST http://localhost:8124/system/findDataObjects => {
  "scope": {
    "project": "project-Bx2pxv80qJ287gqPFQQK3V5z", 
    "folder": "/", 
    "recurse": false
  }, 
  "describe": {
    "properties": true, 
    "details": true
  }, 
  "limit": 100, 
  "name": {
    "glob": "user-000000000000000000000001"
  }, 
  "visibility": "either"
}

  File "/home/ivoloshanenko/dx-toolkit/share/dnanexus/lib/python2.7/site-packages/dxpy/scripts/dx.py", line 1110, in describe
    raise DXCLIError(str(details))
DXCLIError: At least VIEW permission is required to search within a project, code 401
`

src/python/dxpy/scripts/dx.py
```diff
@@ -2512,23 +2512,12 @@ def install(args):
...
-        try:
-            dxpy.api.app_uninstall(args.app)
-            print('Uninstalled application with id: ' + args.app)
-        except:
-            try:
-                dxpy.api.app_uninstall('app-' + args.app)
-                print('Uninstalled application with name: ' + args.app)
-            except:
-                err_exit()
+    user_data = dxpy.api.user_describe(dxpy.whoami(), {'appsInstalled': True})
```
ptn24: Please invoke this method with request data format `{"fields": {"appsInstalled": True}}`; the previous data format has been deprecated.

src/python/dxpy/scripts/dx.py
```diff
@@ -2512,23 +2512,12 @@ def install(args):
         err_exit()
 
 def uninstall(args):
-    app_desc = get_app_from_path(args.app)
```
ptn24: I think this original flow is still necessary because we want to support users specifying the app by path. Please bring this back in. I apologize if my previous comment misled you.

ivoloshanenko:
Commit updated

ivoloshanenko:
@ptn24 @psung PR updated with latest comments from your both.

Can you please review it and merge if LGTM ?

src/python/dxpy/scripts/dx.py
```diff
@@ -2513,22 +2513,16 @@ def install(args):
...
     else:
-        try:
-            dxpy.api.app_uninstall(app_desc['id'])
-            print('Uninstalled the ' + app_desc['name'] + ' app')
-        except:
-            err_exit()
+        user_data = dxpy.api.user_describe(dxpy.whoami(), {"fields": {"appsInstalled": True}})
+        if args.app in user_data['appsInstalled']:
+            try_call(dxpy.api.app_uninstall, 'app-' + args.app)
+        else:
```
ptn24: Please only execute the next line if `args.app`, possibly without the leading "app-", is in the describe response.

src/python/dxpy/scripts/dx.py
```diff
@@ -2513,22 +2513,16 @@ def install(args):
...
+            try_call(dxpy.api.app_uninstall, args.app)
+        print('Uninstalled the {app} app'.format(app=args.app))
```
ptn24: Else, we can print the message that was originally from [here](https://github.com/dnanexus/dx-toolkit/commit/41788a0422c18f0ee8f043e315cc5c2c0ad7ece4).

src/python/dxpy/scripts/dx.py
ivoloshanenko: @ptn24 but we do exactly this thing...
We received response from describe to user_data vat and then compare with given args...
Or you mean do app_describe? If yes, than i don;t understand you previous comment... As we changed code to reflect appinstalled for user instead of simple app describe

src/python/dxpy/scripts/dx.py
ivoloshanenko: @ptn24  you mean replace print for else block with
 print('Uninstalled application with name: ' + args.app) ?

src/python/dxpy/scripts/dx.py
nicolasbockg: I think we only want to call app_uninstall in the case where args.app is in user_data['appsInstalled]. And without prepending the 'app-' (the api actually does that internally).
If the app is not in that list, then we want to show the message we had before: `parser.exit(1, 'Could not find the app\n')`
Basically something like this:
'user_data = dxpy.api.user_describe(dxpy.whoami(), {"fields": {"appsInstalled": True}})
        if args.app in user_data['appsInstalled']:
            try_call(dxpy.api.app_uninstall, args.app)
            print('Uninstalled the {app} app'.format(app=args.app))
        else:
            parser.exit(1, 'Could not find the app\n')
'

src/python/dxpy/scripts/dx.py
nicolasbockg: See my comment above for clarification.

ivoloshanenko: Tnx @nicolasbockg . Absolutely clear now.
Sorry for stupid questions

nicolasbockg: No Problem Igor, please don't hesitate to ask any questions.

src/python/dxpy/scripts/dx.py
ptn24: Let me show an example. Let the user be user-pnguyen.

```
pnguyen@pnguyen:~/env$ dx api user-`dx whoami` describe '{"fields": {"appsInstalled": true}}'
{
    "appsInstalled": {
        "url_fetcher": true
    }, 
    "id": "user-pnguyen"
}
```
For sake of example, suppose that user-pnguyen no longer has any authorizations to `url_fetcher`. I think it will make sense for the following to occur:

- `dx uninstall url_fetcher`: Uninstalls `url_fetcher` from the profile of user-pnguyen.
- `dx uninstall app-url_fetcher`: Uninstalls `url_fetcher` from the profile of user-pnguyen.
- `dx uninstall app-url_fletcher`: (**Note the typo**.) Throws an error. This makes it more intuitive to user-pnguyen that he was not able to uninstall the app that he may have wanted to uninstall. Does this make sense?

Additionally, we want to preserve the ability to specify the app by its path (hence the need to maintain `app_desc = get_app_from_path(args.app)`).

ivoloshanenko: Seems api doesn't add 'app-' prefix automatically. In case executing:
```
if args.app in user_data['appsInstalled']:
            try_call(dxpy.api.app_uninstall, args.app)
```
we will get
```
> dx describe user-000000000000000000000001 | grep Apps
Apps installed      identityFn, uninstall_test
> dx uninstall uninstall_test
ResourceNotFound: The specified URL could not be found, code 404
```

psung: That's correct... at the API level, you must supply the "app-" prefix as it's considered part of the ID of the app.

ivoloshanenko:
Pull request updated. Now we call app_uninstall in the case where args.app is in user_data['appsInstalled].

src/python/dxpy/scripts/dx.py
```diff
@@ -2513,22 +2513,18 @@ def install(args):
...
-        try:
-            dxpy.api.app_uninstall(app_desc['id'])
-            print('Uninstalled the ' + app_desc['name'] + ' app')
-        except:
-            err_exit()
+        user_data = dxpy.api.user_describe(dxpy.whoami(), {"fields": {"appsInstalled": True}})
+        if args.app in user_data['appsInstalled']:
+            try_call(dxpy.api.app_uninstall, 'app-' + args.app)
+        elif args.app.startwith('app-') and args.app[len('app-'):] in user_data['appsInstalled']:
+            parser.exit(1, 'Could not find the app\n')
```
ptn24: I think this should be flipped with L2526, right?

ptn24:
Please see the inline comment, and please consider adding regression tests to cover the specification that we are supporting.

src/python/dxpy/scripts/dx.py
ivoloshanenko: Not exactly, line 2526 we using when user trying to uninstall with app ID. If we continue support uninstalling by ID, we should accept any values, not only from 'user_describe' list. Additionally we can check 'app-' prefix or parameter length to find out if this is ID, or accept any and parse 404 response for custom 'app not found' message, but if this make sense?

ptn24: I do not understand. In L2523, we check whether the specified app is by id (e.g., app-foo) and, if so, whether the suffix (e.g., foo) is in the user describe response. I am confused why something that passes the checkpoint should result in the client throwing an exception. I think we should attempt to uninstall the app here, no?

Also, please note that L2526 is executed for a variety of reasons:
  - The specified app is by id and the id is not in the user describe response.
  - The specified app is by name and the name is not in the user describe response.
For both of the above, I think it will be better to throw an exception.

ivoloshanenko:
I'm confused a little, so I rewrote method,  now it calling uninstall only in one place. Let's check it line by line:
```
2514: def uninstall(args):
2515:     app_desc = get_app_from_path(args.app)
2516:     if app_desc:
2517:         args.app = app_desc['name']
2518:     user_data = dxpy.api.user_describe(dxpy.whoami(), {"fields": {"appsInstalled": True}})
2519:     if args.app in user_data['appsInstalled']:
2520:         args.app = 'app-' + args.app
2521:     if args.app.startswith('app-'):
2522:         try_call(dxpy.api.app_uninstall, args.app)
2523:         print('Uninstalled the {app} app'.format(app=args.app))
2524:     else:
2525:         parser.exit(1, 'Could not find the app\n')
```
line 2515-2517: if user provide full path, then we extracting app name.
line 2518: receiving list of installed applications (names)
line 2519-2520: if user provide correct application name, then we add 'app-' prefix
line 2521-2523: if user provide ID of application or correct name (both of them now have app- prefix) then call uninstall
line 2524-2525: else print not found message

Whether this version is acceptable?

nicolasbockg:
I think you should still have two cases, one for when the call to 
`app_desc = get_app_from_path(args.app)` succeedes, and one for when it fails. When this call succeeds, you just call uninstall with the `id`: `dxpy.api.app_uninstall(app_desc['id'])`. This was the original code so far. 
The new code is for the second case, when app_desc is None. In this case you need to determine if the app is installed and then call uninstall by name. 
I would do it, similar to what you had in a previous revision:
```
app_desc = get_app_from_path(args.app)
if app_desc:  # the description succeded, so we uninstall by id:
    try:
            dxpy.api.app_uninstall(app_desc['id'])
            print('Uninstalled the ' + app_desc['name'] + ' app')
        except:
            err_exit()
else:
    # the description failed, so we try to determine if the app installed at all
    user_data = dxpy.api.user_describe(dxpy.whoami(), {"fields": {"appsInstalled": True}})
    if args.app in user_data['appsInstalled']:
        try_call(dxpy.api.app_uninstall, 'app-' + args.app)
        print('Uninstalled the {app} app'.format(app=args.app))
    elif args.app.startwith('app-') and args.app[len('app-'):] in user_data['appsInstalled']:
        try_call(dxpy.api.app_uninstall, args.app)
        print('Uninstalled the {app} app'.format(app=args.app))
    else:
          parser.exit(1, 'Could not find the app\n')
```

ivoloshanenko:
Pull request updated. Line 2522 here we also indirectly check if removing by ID (all ID starting with app-). We need support this option, because:
```
> dx uninstall --help
...
positional arguments:
  app         ID or name of app to uninstall
```

ptn24:
This mostly LGTM. Please add regression tests to capture all, or most, of the cases we discussed. Thank you!

ivoloshanenko:
Test updated

ptn24:
LGTM. Thank you for adding the tests!

PR #152, opened 2016-05-31
PTFM-15653 Paginated search Java API update

storozhilov-dnanexus:
Hello @psung ,

I have implemented your proposal. I added just one new `public SearchPage<T> getPage(int pageSize)` method to `DXSearch.FindDataObjectsRequestBuilder<T>` class and made iterations to be done using 'public SearchPage<T> next()' method in `DXSearch.SearchPage<T>` class.

That is because this kind of API looks similar to `Iterator<E>`, but I'm not sure about that - it's just a proposal. If you are not comfortable with that - inform me, please, and I would update this PR. It looks like a 5-10 min change...

P.S. I tried to make `DXSearch.SearchPage<T>` to implement `Iterator<DXSearch.SearchPage<T>>` interface, but in that case iterator and actual value are represented by the same object and I do not know is it good or bad.

```
// Iterating current page's items
for (DXRecord r : page) {
    ...
}
// Switching to the next page
if (page.hasNext()) {
    page = page.next(); // <- I suspect this is not compatible with 'Iterator<E>' Java concept
}
```

All the best,
Ilya

### DIFF
--- a/src/java/src/main/java/com/dnanexus/DXSearch.java
+++ b/src/java/src/main/java/com/dnanexus/DXSearch.java
@@ -157,7 +157,7 @@ public final class DXSearch {
             this.created = previousQuery.created;
             this.describe = previousQuery.describe;
 
-            this.starting = starting.isNull() ? null : starting;
+            this.starting = (starting == null || starting.isNull()) ? null : starting;
             this.limit = limit;
         }
 
@@ -329,37 +329,6 @@ public final class DXSearch {
         public FindDataObjectsResult<T> execute(int pageSize) {
             return new FindDataObjectsResult<T>(this.buildRequestHash(null, pageSize), this.classConstraint, this.env,
                     pageSize);
-        }
-
-        /**
-         * Executes the query and returns the first page of results.
-         *
-         * @param pageSize number of elements to retrieve
-         *
-         * @return result set
-         */
-        public SearchPage<T> getFirstPage(int pageSize) {
-            if (pageSize <= 0) {
-                throw new IllegalArgumentException("Page size must be a positive integer");
-            }
-            return new SearchPage<T>(this.buildRequestHash(null, pageSize), this.classConstraint, this.env);
-        }
-
-        /**
-         * Executes the query, returning a subsequent page of results starting from the specified
-         * location.
-         *
-         * @param starting result of {@link SearchPage#getNext()} call on previous page
-         * @param pageSize number of elements to retrieve
-         *
-         * @return result set
-         */
-        public SearchPage<T> getSubsequentPage(JsonNode starting, int pageSize) {
-            Preconditions.checkNotNull(starting);
-            if (pageSize <= 0) {
-                throw new IllegalArgumentException("Page size must be a positive integer");
-            }
-            return new SearchPage<T>(this.buildRequestHash(starting, pageSize), this.classConstraint, this.env);
         }
 
         /**
@@ -926,6 +895,112 @@ public final class DXSearch {
      */
     @JsonIgnoreProperties(ignoreUnknown = true)
     public static class FindDataObjectsResult<T extends DXDataObject> extends ObjectProducerImpl<T> {
+
+        /**
+         * The page subset of data objects that matched a {@code findDataObjects} query.
+         */
+        public class Page implements Iterable<T> {
+            /**
+             * Iterator implementation for {@code findDataObjects} results page items.
+             */
+            private class PageIterator implements Iterator<T> {
+
+                int nextElementIndex;
+
+                private PageIterator() {
+                    this.nextElementIndex = 0;
+                }
+
+                @Override
+                public boolean hasNext() {
+                    return nextElementIndex < response.results.size();
+                }
+
+                @Override
+                public T next() {
+                    return getDataObjectInstanceFromResult(response.results.get(nextElementIndex++));
+                }
+
+                @Override
+                public void remove() {
+                    throw new UnsupportedOperationException();
+                }
+            }
+
+            private final FindDataObjectsResponse response;
+
+            /**
+             * Initializes result set page object with the specified page size.
+             */
+            private Page(int pageSize) {
+                this.response = DXAPI.systemFindDataObjects(new FindDataObjectsRequest(baseQuery, null, pageSize),
+                        FindDataObjectsResponse.class, env);
+            }
+
+            /**
+             * Initializes result set page object with the specified page size.
+             */
+            private Page(JsonNode starting, int pageSize) {
+                this.response = DXAPI.systemFindDataObjects(new FindDataObjectsRequest(baseQuery, starting, pageSize),
+                        FindDataObjectsResponse.class, env);
+            }
+
+            /**
+             * Returns an object that can be used in a later call to
+             * {@link DXSearch.FindDataObjectsResult#getSubsequentPage(JsonNode, int)}
+             * to continue retrieving results where this page leaves off.
+             */
+            public JsonNode getNext() {
+                return response.next.deepCopy();
+            }
+
+            public boolean hasNext() {
+                return response.next != null && !response.next.isNull();
+            }
+
+            @Override
+            public Iterator<T> iterator()
+            {
+                return new PageIterator();
+            }
+
+            /**
+             * Returns an amount of items on findDataObjects results page
+             */
+            public int size() {
+                return response.results.size();
+            }
+        }
+
+        /**
+         * Returns a first page of the {@code findDataObjects} results.
+         *
+         * @param pageSize number of elements to retrieve
+         *
+         * @return result set
+         */
+        public Page getFirstPage(int pageSize) {
+            if (pageSize <= 0) {
+                throw new IllegalArgumentException("Page size must be a positive integer");
+            }
+            return new Page(pageSize);
+        }
+
+        /**
+         * Returns a subsequent page of the {@code findDataObjects} results starting from the specified item.
+         *
+         * @param starting result of {@link DXSearch.FindDataObjectsResult.Page#getNext()} call on previous page
+         * @param pageSize number of elements to retrieve
+         *
+         * @return result set
+         */
+        public Page getSubsequentPage(JsonNode starting, int pageSize) {
+            Preconditions.checkNotNull(starting);
+            if (pageSize <= 0) {
+                throw new IllegalArgumentException("Page size must be a positive integer");
+            }
+            return new Page(starting, pageSize);
+        }
 
         /**
          * Wrapper from the findDataObjects result page class to the high-level interface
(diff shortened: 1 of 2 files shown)


src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -989,14 +1001,116 @@ private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
...
+        public Iterator<T> iterator()
+        {
+            return new ResultIterator();
+        }
+
+        public boolean hasNext() {
+            return response.next != null && !response.next.isNull();
+        }
+
+        public SearchPage<T> next() {
```
psung: I'd prefer if instead of having to serialize the whole SearchPage, the client could just get the same JSON thing that the API gives them, and then later resume the query by supplying the same (or equivalent) request and this deserialized value. e.g.

```
class SearchPage {
    public JsonNode getNext() { // Not sure what is a good name for this,
        // so as to not confuse it with the iterator "next"
}
```

And then above, in FindDataObjectsRequestBuilder, an additional method to resume the queries:

```
class FindDataObjectsRequestBuilder {
    public FindDataObjectsResult<T> execute() {...}

    public SearchPage<T> getFirstPage() {...}

    public SearchPage<T> getSubsequentPage(JsonNode starting) {...}
}
```

storozhilov-dnanexus: Hello @psung ,

thank you very much for the review - I've fixed that, see my last update, please.

All the best,
Ilya

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -308,11 +309,34 @@ FindDataObjectsRequest buildRequestHash(int limit) {
...
+        /**
+         * Executes the query and returns subsequent result set page.
+         *
+         * @param pageSize Page size
+         * @param starting Result of SearchPage<T>.getNext() call on previous page
+         *
+         * @return Iterable result set page
+         */
+        public SearchPage<T> getSubsequentPage(int pageSize, JsonNode starting) {
+            return new SearchPage<T>(this.buildRequestHash(pageSize, starting), this.classConstraint, this.env);
```
psung: In this code path, assert that starting is not null.

e.g. Preconditions.checkNotNull

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -989,14 +1013,111 @@ private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
...
             return new ResultIterator();
         }
     }
 
     /**
+     * The page subset of data objects that matched a {@code findDataObjects} query.
+     *
+     * @param <T> data object class to be returned
+     */
+    @JsonIgnoreProperties(ignoreUnknown = true)
```
psung: This no longer needs to be Jackson-serializable, I think, so this annotation could be removed.

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -989,14 +1013,111 @@ private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
...
+
+            int lastElementIndex;
+
+            private ResultIterator() {
+                this.lastElementIndex = -1;
+            }
+
+            @Override
+            public boolean hasNext() {
+                return lastElementIndex < (response.results.size() - 1);
```
psung: Minor: seems a bit more natural to name the counter nextElementIndex, start it from 0, and remove the "-1" from the condition here.

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -989,14 +1013,111 @@ private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
...
+                                      DXEnvironment env) {
+            this.classConstraint = classConstraint;
+            this.env = env;
+            this.response = DXAPI.systemFindDataObjects(request, FindDataObjectsResponse.class, env);
+        }
+
+        /**
+         * Returns an amount of items on findDataObjects results page
+         */
+        int size() {
```
psung: Mark this VisibleForTesting (unless you intended it to be part of the public API).

psung:
It mostly looks good to me, pending the above comments. Nice work!

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: Done

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: Done

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: Done

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: Actually I wanted it to be a part of public API, so I made it **public**. And unit-test checks that as well...

storozhilov-dnanexus:
Hi @psung ,
thank you for your review! I hope I've resolved all minor issues above - awaiting for feedback from your side.
Have a nice day,
Ilya

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -308,11 +309,43 @@ FindDataObjectsRequest buildRequestHash(int limit) {
...
+         * @param pageSize Page size
+         * @param starting Result of SearchPage<T>.getNext() call on previous page
+         *
+         * @return Iterable result set page
+         */
+        public SearchPage<T> getSubsequentPage(int pageSize, JsonNode starting) {
+            if (pageSize <= 0) {
+                throw new IllegalArgumentException("Invalid page size - should be positive");
+            }
+            if (starting == null) {
```
psung: Typically for errors of this type we would throw NullPointerException. In storozhilov-dnanexus#2 I changed this to use Preconditions.checkNotNull which, in addition to being a bit more compact, also changes the exception type that is thrown.

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -989,14 +1022,110 @@ private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
...
+                return getDataObjectInstanceFromResult(response.results.get(nextElementIndex++));
+            }
+
+            @Override
+            public void remove() {
+                throw new UnsupportedOperationException();
+            }
+
+            @SuppressWarnings("unchecked")
+            private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
```
psung: Can this be factored out so as to not duplicate the same method in FindDataObjectsResult.ResultIterator?

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -989,14 +1022,110 @@ private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
...
             return new ResultIterator();
         }
     }
 
     /**
+     * The page subset of data objects that matched a {@code findDataObjects} query.
+     *
+     * @param <T> data object class to be returned
+     */
+    public static class SearchPage<T extends DXDataObject> implements Iterable<T> {
```
psung: Can the SearchPage template parameter be made more generic so that the same class can be used for e.g. FindExecutions in the future? (Possibly, you would have to make this an abstract class and then write a concrete class that does the DXDataObject-specific transformations below.)

If the above isn't feasible for whatever reason, this should be renamed to FindDataObjectsSearchPage to avoid colliding with the similar classes we might introduce for FindExecutionsSearchPage, etc. in the future.

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -989,14 +1022,110 @@ private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
...
+        private final String classConstraint;
+        private final DXEnvironment env;
+        private final FindDataObjectsResponse response;
+
+        /**
+         * Iterator implementation for findDataObjects results page items.
+         */
+        @VisibleForTesting
```
psung: VisibleForTesting should only be applied to default-visibility methods.

src/java/src/main/java/com/dnanexus/DXSearch.java
psung: Once you've done so, consider whether Guava's Iterators.transform would be useful here (it seems like it might obviate the need for this ResultIterator class you've defined).

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -989,14 +1022,110 @@ private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
...
+        {
+            return new ResultIterator();
+        }
+
+        public boolean hasNext() {
+            return response.next != null && !response.next.isNull();
+        }
+
+        public JsonNode getNext() {
+            return response.next;
```
psung: Since we're returning an object to the caller here we should make a defensive copy. Maybe JsonNode.deepCopy, or I don't know if there's a way to present an immutable view of a JsonNode?

src/java/src/main/java/com/dnanexus/DXSearch.java
```diff
@@ -989,14 +1022,110 @@ private T getDataObjectInstanceFromResult(FindDataObjectsResponse.Entry e) {
...
+        public Iterator<T> iterator()
+        {
+            return new ResultIterator();
+        }
+
+        public boolean hasNext() {
+            return response.next != null && !response.next.isNull();
+        }
+
+        public JsonNode getNext() {
```
psung: Please add a docstring, e.g.

> Returns an object that can be used in a later call to getSubsequentPage to continue retrieving results where this page leaves off.

src/java/src/test/java/com/dnanexus/DXSearchTest.java
```diff
@@ -597,6 +597,117 @@ public void testFindDataObjectsWithPaging() {
...
+                            .setName("foo" + Integer.toString(i)).build();
+            records.add(record);
+            recordIds.add(record.getId());
+        }
+        List<DXRecord> outputRecords =
+                DXSearch.findDataObjects().inProject(testProject).nameMatchesGlob("foo*")
+                        .withClassRecord().execute().asList();
+        Assert.assertEquals(8, outputRecords.size());
+
+        List<DXRecord> outputRecordsWithPaging =
```
psung: Seems like outputRecordIds below isn't used, and in any case this entire block seems to duplicate code in another test. Can this block be removed entirely?

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: Got it. Thank you!

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: > Can this be factored out so as to not duplicate the same method in FindDataObjectsResult.ResultIterator?

Since **getDataObjectInstanceFromResult()** is a member of **FindDataObjectsResult<T>** class I decided to move **SearchPage** class there. And I've renamed it to **FindDataObjectsResult<T>::Page** because the name of outer class (**FindDataObjectsResult**) contains all necessary information about what this **Page** is about. Keep me informed, please, if you are not comfortable with such naming.

> Once you've done so, consider whether Guava's Iterators.transform would be useful here (it seems like it might obviate the need for this ResultIterator class you've defined).

I looked to Guava library and found it very interesting and helpful - thank you very much! So I've removed a **ResultIterator** class from the code and implemented that using Guava's **AbstractIterator**.

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: Done

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: Done. Good to know - would keep it in mind in future.

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: Done

src/java/src/test/java/com/dnanexus/DXSearchTest.java
storozhilov-dnanexus: Done

src/java/src/main/java/com/dnanexus/DXSearch.java
storozhilov-dnanexus: After last modifications the implementation of search page class became to [be quite trivial](https://github.com/storozhilov-dnanexus/dx-toolkit/blob/PTFM-15653-resolution/src/java/src/main/java/com/dnanexus/DXSearch.java#L903-L955). That is why I didn't introduce a base class for that so far, but decided to ask you first. Maybe it makes sense to do that when to implement the same pagination for executions search if we decide, that it's definitely needed?

Since I've moved **SearchPage** into **FindDataObjectsResult** it looks like naming is not an issue.

Keep me informed, please, if you think we are still having issues here.

psung: Thanks, this looks fine to me. The base class only seems useful if there is a way for clients to consume search results generically regardless of their type. Not sure if this will be useful, but we can always introduce it later by inserting it as a superclass.

psung:
LGTM, thanks! Merging shortly.

psung:
Merged as b68e1c4

PR #154, opened 2016-06-06
PTFM-17276 Improve dx build help message when app name is taken

ivoloshanenko:
https://dnanexus.jira.com/browse/PTFM-17276

### DIFF
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -472,6 +472,8 @@ def _verify_app_writable(app_name):
     except dxpy.exceptions.DXAPIError as e:
         if e.name == 'ResourceNotFound':
             app_name_already_exists = False
+        elif type(e) == dxpy.exceptions.PermissionDenied:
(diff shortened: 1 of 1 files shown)


src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -472,6 +472,8 @@ def _verify_app_writable(app_name):
     except dxpy.exceptions.DXAPIError as e:
         if e.name == 'ResourceNotFound':
             app_name_already_exists = False
+        elif type(e) == dxpy.exceptions.PermissionDenied:
```
psung: This should be in a similar format to the above e.g. `e.name == "PermissionDenied"`

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -472,6 +472,8 @@ def _verify_app_writable(app_name):
...
+            raise dxpy.app_builder.AppBuilderException('App with given name already exist under another project! Please choose another unique name!')
```
psung: This message is a bit misleading because apps don't exist within projects. Something like the following:

```
An app with the given name already exists and you are not a developer of that app
```

ivoloshanenko:
Fixed.

PR #155, opened 2016-06-08
PTFM-17466 Download folder python function

storozhilov-dnanexus:
Hi @psung ,

this is an implementation for PTFM-17466 - can you review that if have some time?

Thank you and have a nice day,
Ilya

P.S. FYI: **pypy3** build configuration fails even for master branch on TravisCI.


src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -466,3 +466,50 @@ def upload_string(to_upload, media_type=None, keep_open=False, wait_on_close=Fal
         handler.close(block=wait_on_close, **remaining_kwargs)
 
     return handler
+
+def download_folder(project, destdir, folder="/", chunksize=dxfile.DEFAULT_BUFFER_SIZE):
```
psung: Please allow the caller to supply additional kwargs, and pass them through to download_dxfile (download_dxfile itself does this). These can be used to customize the low-level request parameters.

psung: Can this method be unified with existing code in dxpy for downloading folders? For example, "dx download -r" must do something similar, so possibly it could be expressed in terms of this method...

src/python/test/test_dxpy.py
```diff
@@ -661,6 +661,68 @@ def test_part_splitting(self):
...
+
+        # Creating remote folders
+        dxproject.new_folder("/a/b/c/d", parents=True)
+        dxproject.new_folder("/a/e/f/g", parents=True)
+        dxproject.new_folder("/h/i/j/k", parents=True)
+
+        # Filling remote folders with objects
+        path = []
+        i = 1
+        for f in ["/", "a", "b", "c", "d"]:
```
psung: When you want to iterate over a collection and also count the number of items you've seen in parallel, you can use `enumerate`:

```
for i, f in enumerate(["/", "a", "b", "c", "d"], start=1):
```

src/python/test/test_dxpy.py
```diff
@@ -661,6 +661,68 @@ def test_part_splitting(self):
...
+            path.append(f)
+            folder = os.path.join(*path)
+            with dxpy.new_dxfile(name="file_{}.txt".format(i), folder=folder) as dxfile:
```
psung: Also see `upload_string` which does this in one shot.

src/python/test/test_dxpy.py
```diff
@@ -661,6 +661,68 @@ def test_part_splitting(self):
...
+        i = 2
+        for f in [a_dest_dir, "b", "c", "d"]:
+            path.append(f)
+            filename = os.path.join(os.path.join(*path), "file_{}.txt".format(i))
+            self.assertTrue(os.path.isfile(filename))
+            self.assertEquals("{}-th\n file\n content\n".format(i), open(filename, "r").read())
+            i += 1
+
+        # Checking download to existing file
+        self.assertRaises(DXFileError, dxpy.download_folder, self.proj_id, self.temp_file_path)
```
psung: The context-manager form of this is a bit easier to read:

```
with self.assertRaises(DXFileError):
    dxpy.download_folder(self.proj_id, self.temp_file_path)
```

src/python/test/test_dxpy.py
storozhilov-dnanexus: Fixed - thank you! Very good to know!

src/python/test/test_dxpy.py
storozhilov-dnanexus: Done

src/python/test/test_dxpy.py
storozhilov-dnanexus: Done

src/python/dxpy/bindings/dxfile_functions.py
storozhilov-dnanexus: > Please allow the caller to supply additional kwargs, and pass them through to download_dxfile (download_dxfile itself does this). These can be used to customize the low-level request parameters.

Done

> Can this method be unified with existing code in dxpy for downloading folders? For example, "dx download -r" must do something similar, so possibly it could be expressed in terms of this method...

I tried to implement that but not successfully. I added an **overwrite** parameter to **download_folder()** method to provide the same behavior as **dx download -r** command. But after that I found, that **dx download -r** used a [directory tree cache](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/cli/download.py#L66) to optimize it's work if to download several remote folders. And it seems to be not a good idea to introduce a similar user-provided directory tree cache to **download_folder()** public API function. What do you think?

Thank you!

psung: Thanks, that's a good point. Is it easy to move the common stuff into an internal method (that supports the cache) and have download_folder be a thin wrapper that disables or prevents access to the use of the caching features? That would allow us to share most of the code but defer this question about whether users of this API need caching.

If it's too complicated, don't bother. Thanks!

storozhilov-dnanexus: Hello @psung ,

thank you for the idea! I hope I managed to port a **dx download -r** command to use **dxpy.download_folder()** API call. To do that I introduced a **dxpy.list_subfolders()** function, which provides a directory listings cache support and which is called either by **dxpy.download_folder()** function or by **dx** CLI command (see [here](https://github.com/storozhilov-dnanexus/dx-toolkit/blob/PTFM-17466-python-download_folder-function/src/python/dxpy/cli/download.py#L134)).

But I have a little concern about the respective [TravisCI build](https://travis-ci.org/storozhilov-dnanexus/dx-toolkit/builds/137848025), which failed for each configuration with following error:

```
...
creating build/bdist.linux-x86_64/wheel/dxpy-c0b0e43.dist-info/WHEEL
unset PYTHONPATH; source "/home/travis/build/storozhilov-dnanexus/dx-toolkit/src/../build/py_env/bin/activate"; pip install --ignore-installed --prefix="/home/travis/build/storozhilov-dnanexus/dx-toolkit/src/.." python/dist/*.whl
dxpy-c0b0e43-py2.py3-none-any.whl is not a valid wheel filename.
make[1]: *** [/home/travis/build/storozhilov-dnanexus/dx-toolkit/src/../bin/dx] Error 1
make[1]: Leaving directory `/home/travis/build/storozhilov-dnanexus/dx-toolkit/src'
```

That looks strange, because Travis build runs fine on my workstation, so I can not reproduce that locally to be able to investigate. Maybe you can help somehow?

Thank you and have a nice day,
Ilya

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -466,3 +466,101 @@ def upload_string(to_upload, media_type=None, keep_open=False, wait_on_close=Fal
...
+        if not os.path.isdir(d):
+            if os.path.exists(d):
+                raise DXFileError("Destination location '{}' already exists and is not a directory".format(destdir))
+            logger.debug("Creating destination directory: '%s'", d)
+            os.makedirs(d)
+
+    def compose_local_dir(remote_subfolder):
+        return os.path.normpath(os.path.join(destdir, remote_subfolder[1:] if folder == "/" else remote_subfolder[len(folder) + 1:]))
+
+    folder = os.path.normpath(folder).strip()
```
psung: I'm not sure if it is permissible to use "os.path" here, since the paths will be handled using the system conventions. That is, this code may be running on Windows but `folder` must be interpreted like a unixy path regardless. Would it be sufficient to strip a trailing "/" here, or do you need more than that?

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -466,3 +466,101 @@ def upload_string(to_upload, media_type=None, keep_open=False, wait_on_close=Fal
...
+    if folder == "":
```
psung: Is there some nontrivial value of `folder` that gets turned into "" here? If so, it might be nice to print the original value of `folder` in the error message below.

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -466,3 +466,101 @@ def upload_string(to_upload, media_type=None, keep_open=False, wait_on_close=Fal
...
+    :param destdir: Local destination location
+    :type destdir: string
+    :param folder: Path to the remote folder to download
+    :type folder: string
+    :param overwrite: Overwrite existing files
+    :type overwrite: boolean
+    :param usecache: Use folders listing cache
+    :type usecache: boolean
+
+    Downloads a remote *folder* of the *project* and saves it to the *destdir* local location.
```
psung: Can you clarify the semantics here? I assume this means that all the contents of `folder` should be added to the directory `destdir`, such that a directory is NOT created with the same basename as that of `folder`.

If so, something like this should express that:

> Downloads the contents of the remote *folder* of the *project* into the local directory specified by *destdir*.

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -466,3 +466,101 @@ def upload_string(to_upload, media_type=None, keep_open=False, wait_on_close=Fal
...
+        project_folders = dxpy.get_handler(project).describe(input_params={'folders': True})['folders']
+    # TODO: support shell-style path globbing (i.e. /a*/c matches /ab/c but not /a/b/c)
+    # return pathmatch.filter(project_folders, os.path.join(path, '*'))
+    if recurse:
+        return (f for f in project_folders if f.startswith(path))
+    else:
+        return (f for f in project_folders if f.startswith(path) and '/' not in f[len(path)+1:])
+
+def download_folder(project, destdir, folder="/", overwrite=False, chunksize=dxfile.DEFAULT_BUFFER_SIZE,
+        usecache=True, **kwargs):
```
psung: I'm not sure anyone besides the existing methods will want to use `usecache`, so let's disable it by default...

psung: Actually, given that you're newly introducing this method, please just omit `usecache` entirely as an argument. You can still use it internally when invoking list_subfolders.

psung: One potential problem with this scheme for caching is that if you have a long-running process, data in the cache could grow arbitrarily stale. Previously the cache was at least scoped to a single call to "download", but this is no longer the case. Perhaps the caching logic could be removed entirely, unless there is some obvious place where there are going to be many repeated calls to the same subfolder in the course of performing some common operation.

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -466,3 +466,101 @@ def upload_string(to_upload, media_type=None, keep_open=False, wait_on_close=Fal
...
+    if folder == "":
+        raise DXFileError("Invalid remote folder name: '{}'".format(folder))
+    destdir = os.path.normpath(destdir).strip()
+    if destdir == "":
+        raise DXFileError("Invalid destination directory name: '{}'".format(destdir))
+
+    remote_folders = list_subfolders(project, folder, usecache=usecache, recurse=True)
+
+    # Creating target directory tree
+    remote_subfolders = [f for f in remote_folders if f.startswith(folder)]
```
psung: Why doesn't this end up being the same as `remote_folders` itself?

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -466,3 +466,101 @@ def upload_string(to_upload, media_type=None, keep_open=False, wait_on_close=Fal
...
+    if len(remote_subfolders) <= 0:
+        raise DXFileError("Remote folder '{}' not found".format(folder))
+    remote_subfolders.sort()
+    for remote_subfolder in remote_subfolders:
+        ensure_local_dir(compose_local_dir(remote_subfolder))
+
+    # Downloading files
+    for remote_file in dxpy.search.find_data_objects(classname='file', state='closed', project=project, folder=folder,
+            recurse=True, describe=True):
+        local_filename = os.path.join(compose_local_dir(remote_file['describe']['folder']), remote_file['describe']['name'])
```
psung: Again, I wonder if something special needs to be done to transform the unity path `folder` into a (possibly-Windows) local filename... if you aren't able to test this in action, please leave a comment saying that the behavior of this code on Windows needs to be looked at some more.

src/python/dxpy/bindings/dxfile_functions.py
storozhilov-dnanexus: Great catch - fixed!

src/python/dxpy/bindings/dxfile_functions.py
storozhilov-dnanexus: Fixed

src/python/dxpy/bindings/dxfile_functions.py
storozhilov-dnanexus: Fixed

src/python/dxpy/bindings/dxfile_functions.py
storozhilov-dnanexus: I've removed remote folders caching

src/python/dxpy/bindings/dxfile_functions.py
storozhilov-dnanexus: Fixed

src/python/dxpy/bindings/dxfile_functions.py
storozhilov-dnanexus: Will complete this soon

storozhilov-dnanexus: It's fixed. I've tested my work on windows-10-x86_64 either from it's original "Command Prompt" program or from mingw/msys bash.

storozhilov-dnanexus:
Hello @psung ,

hopefully I've fixed all your comments - take a look on that, please, if have some time.

Thank you and have a nice day,
Ilya

psung:
Thank you, and sorry for the delay in getting this merged. Merged as fd6f8ba1

PR #157, opened 2016-06-24
Removed gtable option from dx find data

thanhhahaha:
@psung, can you review this small change? A (hopefully) non-breaking change to the `dx find data` helpstring


psung:
Thanks!

PR #158, opened 2016-06-24
PTFM-7513, dx-build-app[let] should check validity of IDs that appear…

ivoloshanenko:
Pull request to solve PTFM-7513 

: dx-build-app-should-check-IDs-in-spec

… in the app spec

### DIFF
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -216,8 +216,7 @@ def _check_suggestions(src_dir, publish=False):
     Examines the specified dxapp.json file and warns about any
     violations of suggestions guidelines.
     """
-    dxapp_json_filename = os.path.join(src_dir, "dxapp.json")
-    app_spec = json.load(open(dxapp_json_filename))
+    app_spec = _parse_app_spec(src_dir)
     for input_field in app_spec.get('inputSpec', []):
         for suggestion in input_field.get('suggestions', []):
             if 'project' in suggestion:
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -240,9 +240,9 @@ def _check_suggestions(src_dir, publish=False):
             if '$dnanexus_link' in suggestion:
                 if suggestion['$dnanexus_link'].startswith(('file-', 'record-', 'gtable-')):
                     try:
-                        object = dxpy.describe(suggestion['$dnanexus_link'])
-                        if 'project' in object:
-                            if not object['project']:
+                        dnanexus_link = dxpy.describe(suggestion['$dnanexus_link'])
+                        if 'project' in dnanexus_link:
+                            if not dnanexus_link['project']:
                                 raise dxpy.app_builder.AppBuilderException(
                                     'Suggested object {name} does not belongs to any project'.format(
                                         name=suggestion['value']['$dnanexus_link']))
@@ -263,9 +263,9 @@ def _check_suggestions(src_dir, publish=False):
                     elif isinstance(suggestion['value']['$dnanexus_link'], basestring):
                         if suggestion['value']['$dnanexus_link'].startswith(('file-', 'record-', 'gtable-')):
                             try:
-                                object = dxpy.describe(suggestion['value']['$dnanexus_link'])
-                                if 'project' in object:
-                                    if not object['project']:
+                                dnanexus_link = dxpy.describe(suggestion['value']['$dnanexus_link'])
+                                if 'project' in dnanexus_link:
+                                    if not dnanexus_link['project']:
                                         raise dxpy.app_builder.AppBuilderException(
                                             'Suggested object {name} does not belongs to any project'.format(
                                                 name=suggestion['value']['$dnanexus_link']))
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -211,13 +211,12 @@ def parse_destination(dest_str):
     return try_call(resolve_path, dest_str)
 
 
-def _check_suggestions(src_dir, publish=False):
+def _check_suggestions(app_json, publish=False):
     """
     Examines the specified dxapp.json file and warns about any
     violations of suggestions guidelines.
     """
-    app_spec = _parse_app_spec(src_dir)
-    for input_field in app_spec.get('inputSpec', []):
+    for input_field in app_json.get('inputSpec', []):
         for suggestion in input_field.get('suggestions', []):
             if 'project' in suggestion:
                 try:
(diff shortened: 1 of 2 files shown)


src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -211,6 +211,59 @@ def parse_destination(dest_str):
     return try_call(resolve_path, dest_str)
 
 
+def _check_suggestions(src_dir, publish=False):
```
nicolasbockg: It seems like `publish` is not used in the code. Please remove if not needed.

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -211,6 +211,59 @@ def parse_destination(dest_str):
...
+                        try:
+                            project = dxpy.api.project_describe(suggestion['project'])
+                            if project['restricted']:
+                                logger.warn('Warning, project {name} is restricted!'.format(name=project['name']))
+                        except dxpy.exceptions.DXAPIError as e:
+                            if e.code == 404:
+                                raise dxpy.app_builder.AppBuilderException(
+                                    'Suggested project {name} does not exist, or not accessible by user'.format(
+                                        name=suggestion['project']))
+                        if 'path' in suggestion:
```
nicolasbockg: Is it possible that suggestions includes a `path`, but not a `project`? If that is the case, we would need to add verification for the path.

nicolasbockg:
Hi Igor, Thanks for this request, it would be nice to add some test cases to verify that the AppBuilderExceptions are raised correctly.

src/python/dxpy/scripts/dx_build_app.py
ivoloshanenko: Yes, you are right. Fixed in second commit.

src/python/dxpy/scripts/dx_build_app.py
ivoloshanenko: Not possible, according to WIKI documentation: "path string (optional; can only be present if project is also present)" https://wiki.dnanexus.com/API-Specification-v1.0.0/IO-and-Run-Specifications

ivoloshanenko:
Test added

nicolasbockg:
Thanks for adding the test, it looks good.
I looked through other test cases and I could not find a test that has suggestions, so I think it would be good to add a positive test to make sure we can build an app with a valid suggestion.

ivoloshanenko:
Test updated, build with a valid suggestion added

nicolasbockg:
Thanks Igor, it LGTM, thanks!

psung:
Hi @nicolasbockg and @ivoloshanenko, if you're both happy with this PR, then @nicolasbockg, please merge the PR (Github should create a squash merge).

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -211,6 +211,69 @@ def parse_destination(dest_str):
...
 
+def _check_suggestions(src_dir, publish=False):
+    """
+    Examines the specified dxapp.json file and warns about any
+    violations of suggestions guidelines.
+    """
+    dxapp_json_filename = os.path.join(src_dir, "dxapp.json")
+    app_spec = json.load(open(dxapp_json_filename))
+    if 'inputSpec' in app_spec:
+        for input_field in app_spec['inputSpec']:
```
psung: Here and elsewhere, you can reduce the nesting level a bit by using "get" with an appropriate default value, e.g.:

```
for input_field in app_spec.get('inputSpec', []):
    for suggestion in input_field.get('suggestions', []):
        if 'project' in suggestion:
```

nicolasbockg:
Hi @ivoloshanenko, when you get a chance to implement Phil's suggestion I can do the merge into master. 
Thanks for this work.

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -211,6 +211,67 @@ def parse_destination(dest_str):
...
+                    try:
+                        check_folder_exists(suggestion['project'], suggestion['path'], '')
+                    except ResolutionError as e:
+                        raise dxpy.app_builder.AppBuilderException(
+                            'Folder {path} could not be found in project {project}'.format(
+                                path=suggestion['path'], project=suggestion['project']))
+            if '$dnanexus_link' in suggestion:
+                if suggestion['$dnanexus_link'].startswith(('file-', 'record-', 'gtable-')):
+                    try:
+                        object = dxpy.describe(suggestion['$dnanexus_link'])
```
nicolasbockg: One last thing, please consider using a different name for the results from `describe` given that `object` is a python built-in function, and reassigning  it might have unwanted  side effects.
Also in line 266. 
Thanks

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -838,6 +899,8 @@ def _build_app(args, extra_args):
     TODO: remote app builds still return None, but we should fix this.
 
     """
+    _check_suggestions(args.src_dir, args.publish)
```
nicolasbockg: Hi Igor, I ran all the tests in TestDXBuildApp, but these two tests failed:
test_build_applet_with_no_dxapp_json and test_build_applet_with_malformed_dxapp_json
I think the solution is to move the call to _check_suggestions to Line 945 after calling _parse_app_spec, and also to line 756 also after calling _parse_app_spec. Let me know if you think that works. Thanks

ivoloshanenko:
Fixed this issue. Now config file loading via _parse_app_spec() method and it checked if file exist and correct.

nicolasbockg:
The tests are still failing for me. I think the issue is that _check_suggestions needs to be inside the try-except blocks. My recommendation is to do the following:

1. Change the signature of _check_suggestions to:
```
def _check_suggestions(app_json, publish=False):
```
2. Call `_check_suggestions` in Line 755 (in function build_and_upload_locally), after parsing the app_json:
```
def build_and_upload_locally(src_dir, ... )
    dxpy.app_builder.build(src_dir, parallel_build=do_parallel_build)
    app_json = _parse_app_spec(src_dir)
     _check_suggestions(app_json, publish=publish)
```
3. Call `_check_suggestions` in Line 994 (in function _build_app):
```
try:
        app_json = _parse_app_spec(args.src_dir)
       _check_suggestions(app_json, publish=args.publish)
        _verify_app_source_dir(args.src_dir, args.mode)
        if args.mode == "app" and not args.dry_run:
            _verify_app_writable(app_json['name'])
```

In this way we take advantage that the app_json has already been parsed, and we do the checking inside the try-except blocks. 
Let me know if you have other questions or if I can clarify anything. 
Thanks

ivoloshanenko:
Yes, this seems good decision. Code updated according to recommendation.

ivoloshanenko:
Test file updated. Moved the successful test in test_build_app_suggestions to its own test, it needs to be tested in isolated env.

nicolasbockg:
Thanks Igor, LGTM!

PR #160, opened 2016-06-27
PTFM-16098 dx download floors percentage

storozhilov-dnanexus:
Hello @psung ,

This a resolution for PTFM-16098 - review, please if have some time. Manually tested only.

All the best,
Ilya


psung:
Thanks!

PR #163, opened 2016-07-05
Handle ClosedPoolError during DXHTTPRequest

mckinsel:
DXHTTPRequest uses a global urllib3.PoolManager to actually make requests. The
PoolManager manages urllib3.ConnectionPools, reusing them when appropriate. If
a request raises an exception, the PoolManager is "cleared", which closes
every ConnectionPool it manages. In rare cases, a ConnectionPool can be
assigned to a request but then closed before the request is made. This causes
a urllib3.exceptions.ClosedPoolError to be raised, which presently is not
handled and causes the DXHTTPRequest to fail.

The ideal solution would fix the concurrency issue, but it appears to be
within urllib3 itself. So, this change just catches ClosedPoolError and
retries.

### DIFF
--- a/src/python/dxpy/__init__.py
+++ b/src/python/dxpy/__init__.py
@@ -489,8 +489,14 @@ def DXHTTPRequest(resource, data, method='POST', headers=None, auth=True,
             body = _maybe_trucate_request(_url, try_index, data)
 
             # throws BadStatusLine if the server returns nothing
-            response = _get_pool_manager(**pool_args).request(_method, _url, headers=_headers, body=body,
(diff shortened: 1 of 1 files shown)


psung:
Thanks. In this case, do we still need the mutex? Could that be removed, or does it protect against other problems?

mckinsel:
I suppose it should keep multiple threads from trying to create and assign PoolManagers to _pool_manager at the same time, so I'd be inclined to keep it.

psung:
Thanks! Merged.

PR #170, opened 2016-08-10
Fixing label for file size

godotgildor:
We were erroneously reporting KB, MB, GB, and TB when we really meant
KiB, MiB, GiB, and TiB.

### DIFF
--- a/src/python/dxpy/bindings/dxfile.py
+++ b/src/python/dxpy/bindings/dxfile.py
@@ -70,13 +70,13 @@ def _readable_part_size(num_bytes):
     if B < KB:
         return '{0} {1}'.format(B, 'bytes' if B != 1 else 'byte')
     elif KB <= B < MB:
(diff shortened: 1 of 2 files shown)


psung:
Thanks! Seems like tests in test_dxpy.py:TestDXFileFunctions.test_readable_part_size would need to be updated as well.

godotgildor:
Sorry I missed that test.  I have updated the test with commit 578ddfb69fb38e354fed690053c76b61b23c6318

kedarmhaswade:
Interesting. Are KiB et. al.  more _readable_ than KB et. al. respectively? Or should we change the multipliers to be powers of 10 (rather than powers of 2) and revert to KB, MB etc.?

psung:
DNAnexus storage billing is defined in terms of binary units, and the web frontend shows sizes in binary units, so I think we should be consistent with both to minimize confusion.

PR #177, opened 2016-09-08
Attempt to pre-load download url cache

godotgildor:
When downloading a file 8 chunks are downloaded in parallel.  Because all of these download threads are launched simultaneously, when they each go to get a download url, they have a cache-miss, and so 8 proxyDownload requests are made where only 1 should be necessary.  This is my attempt to fix this by requesting a download url prior to the actual downloading of the individual chunks.  This first request is just to pre-load the cache and no data is actually downloaded.  Then the threadpool is used to actually download chunks.


psung:
After some discussion, we decided that the fix should go elsewhere.

PR #178, opened 2016-09-13
Export upload file buffer size to cli

orodeh:
Allow performing a dx upload, while setting the buffer chunk size. This should significantly reduce the overhead on the backend database when uploading large files. This is a situation we run into while executing parallel GLnexus runs.

### DIFF
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -350,7 +350,7 @@ def upload_local_file(filename=None, file=None, media_type=None, keep_open=False
 
         # Use 'a' mode because we will be responsible for closing the file
         # ourselves later (if requested).
-        handler = new_dxfile(mode='a', media_type=media_type, write_buffer_size=dxfile.DEFAULT_BUFFER_SIZE,
+        handler = new_dxfile(mode='a', media_type=media_type,
                              expected_file_size=file_size, file_is_mmapd=file_is_mmapd, **creation_kwargs)
 
     # For subsequent API calls, don't supply the dataobject metadata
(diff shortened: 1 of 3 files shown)


mlin:
@JoeSachmo @psung I'd like to merge this low-risk diff in a couple of days if no objections in the meantime. It just exposes an existing knob as a command-line argument, keeping the current default.

src/python/dxpy/scripts/dx.py
```diff
@@ -3741,6 +3744,7 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
 parser_upload.add_argument('--wait', help='Wait until the file has finished closing', action='store_true')
 parser_upload.add_argument('--no-progress', help='Do not show a progress bar', dest='show_progress',
                            action='store_false', default=sys.stderr.isatty())
+parser_upload.add_argument('--buffer-size', help='Set the write buffer size', dest='write_buffer_size')
```
psung: Mention that this should be specified in units of bytes? It would be nice at some point in the future to allow specifying these with a suffix, like ``--buffer-size=16M`.

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -288,7 +288,7 @@ def verify_part(_part_id, got_bytes, hasher):
 
         return True
 
-def upload_local_file(filename=None, file=None, media_type=None, keep_open=False,
+def upload_local_file(filename=None, file=None, media_type=None, keep_open=False, write_buffer_size=None,
```
psung: New keyword args should be specified after any existing keyword args for compatibility with existing callers.

PR #179, opened 2016-09-25
Adding bulk data object describe to dxpy.describe

godotgildor:
A previous commit added the system_describe_data_objects method to the api module, allowing somewhat easier access to the bulk describeDataObjects API method.  This commit extends this so that users can make use of dxpy.describe() on a list of dxlinks or object-id's to get a list of descriptions back.  This parallels nicely with the existing dxpy.describe function.

Unit test is included.

### DIFF
--- a/src/python/dxpy/bindings/dxdataobject_functions.py
+++ b/src/python/dxpy/bindings/dxdataobject_functions.py
@@ -146,7 +146,8 @@ def get_handler(id_or_link, project=None):
 
 def describe(id_or_link, **kwargs):
     '''
-    :param id_or_link: String containing an object ID or dict containing a DXLink
+    :param id_or_link: String containing an object ID or dict containing a DXLink,
+                       or a list of object ID's or dict's containing a DXLink.
 
     Given an object ID, calls :meth:`~dxpy.bindings.DXDataObject.describe` on the object.
 
@@ -154,8 +155,23 @@ def describe(id_or_link, **kwargs):
 
         describe("file-1234")
     '''
-    handler = get_handler(id_or_link)
-    return handler.describe(**kwargs)
+    # If this is a list, extract the ids.
+    if isinstance(id_or_link, list):
+        links = []
+        for link in id_or_link:
+            # If this entry is a dxlink, then get the id.
+            if is_dxlink(link):
+                # Guaranteed by is_dxlink that one of the following will work
+                if isinstance(link['$dnanexus_link'], basestring):
+                    link = link['$dnanexus_link']
+                else:
+                    link = link['$dnanexus_link']['id']
+            links.append(link)
+        retval = dxpy.api.system_describe_data_objects({'objects': links})
+        return [rv['describe'] for rv in retval['results']]
+    else:
+        handler = get_handler(id_or_link)
+        return handler.describe(**kwargs)
 
 def get_details(id_or_link, **kwargs):
     '''
--- a/src/python/dxpy/bindings/dxdataobject_functions.py
+++ b/src/python/dxpy/bindings/dxdataobject_functions.py
@@ -156,7 +156,10 @@ def describe(id_or_link, **kwargs):
         describe("file-1234")
     '''
     # If this is a list, extract the ids.
-    if isinstance(id_or_link, list):
+    if isinstance(id_or_link, string) or is_dxlink(id_or_link):
+        handler = get_handler(id_or_link)
+        return handler.describe(**kwargs)
+    else:
         links = []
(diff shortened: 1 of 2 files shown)


src/python/dxpy/bindings/dxdataobject_functions.py
```diff
@@ -146,16 +146,32 @@ def get_handler(id_or_link, project=None):
...
     Given an object ID, calls :meth:`~dxpy.bindings.DXDataObject.describe` on the object.
 
     Example::
 
         describe("file-1234")
     '''
-    handler = get_handler(id_or_link)
-    return handler.describe(**kwargs)
+    # If this is a list, extract the ids.
+    if isinstance(id_or_link, list):
```
psung: It would be nice to treat tuples, sets, or other sequence types in the same way you treat lists. What do you think about processing them in the opposite order...

```
if isinstance(id_or_link, basestring) or isinstance(id_or_link, dict):
    handler = get_handler(id_or_link)
    return handler.describe(**kwargs)
else:
    # id_or_link is now presumed to be a sequence of str/dict values...
    links = []
    # etc.
```

src/python/dxpy/bindings/dxdataobject_functions.py
```diff
@@ -146,16 +146,32 @@ def get_handler(id_or_link, project=None):
...
+        for link in id_or_link:
+            # If this entry is a dxlink, then get the id.
+            if is_dxlink(link):
+                # Guaranteed by is_dxlink that one of the following will work
+                if isinstance(link['$dnanexus_link'], basestring):
+                    link = link['$dnanexus_link']
+                else:
+                    link = link['$dnanexus_link']['id']
+            links.append(link)
+        retval = dxpy.api.system_describe_data_objects({'objects': links})
```
psung: Call this something other than `retval` since it's not the value that you're eventually going to return from this method.

src/python/test/test_dxpy.py
```diff
@@ -2592,6 +2592,42 @@ def test_get_handler(self):
...
+        ids = [o.get_id() for o in objects]
+        desc = dxpy.describe(ids)
+
+        self.assertEqual(len(ids), len(desc))
+        for i in xrange(len(desc)):
+            self.assertEqual(desc[i]["project"], self.proj_id)
+            self.assertEqual(desc[i]["id"], ids[i])
+            self.assertEqual(desc[i]["class"], types[i])
+            self.assertEqual(desc[i]["types"], []) 
+            self.assertTrue("created" in desc[i])
```
psung: Here and below, use the assertions self.assertIn and self.assertNotIn, which will give you nicer error messages when the condition is violated.

src/python/dxpy/bindings/dxdataobject_functions.py
godotgildor: We can certainly accept tuples and treat them the same as a list.  Since sets are unordered though, returning a list of the descriptions as we do now may not make sense.

We could have another path that accepts sets and then returns a dict with keys = set elements and vals set to the description.

What do you think?

src/python/test/test_dxpy.py
godotgildor: Do you mean change
self.assertEqual(desc[i]["id"], ids[i])
to 
self.assertIn(desc[i]["id"], ids)
?

That would no longer test that the order is proper though, right?  We want to make sure that the order of the descriptions we get back matches the order of the id's we send in, right?

src/python/dxpy/bindings/dxdataobject_functions.py
psung: Yeah, maybe sets isn't the best example of a useful sequence type, but it seems reasonable to be somewhat agnostic of the exact thing the user passes in, and return a list that has the same iteration order as whatever collection the user supplied. If the user specified a data structure with an unspecified iteration order, that's their problem.

The dict idea would be a more general solution. If you think that would be useful and you feel like implementing it, go for it. I think what you have here is good though.

godotgildor:
Phil, 
  I just pushed updates that I believe address your suggestions.  I changed the order of checking the datatype passed into describe.  Now if the object is either
a) a basestring
b) a dx_link
it will follow the previous describe route.  Otherwise it assumes the object is an iterable (list, tuple, set, etc.) and will create a list of those iterable values to pass to the bulk describe API method.

I have also changed the variable name for the descriptions returned from the describe helper.

Finally, I updated the test to use the AssertIn method rather than AssertTrue on a statement testing whether an object is in the container.

PR #182, opened 2016-10-13
AZURE-224 More detailed logging

storozhilov-dnanexus:
Hi @psung ,

this is what I changed in DX to print out a URL instead of "[function_name]" in debug mode - maybe it makes sense to merge that.

All the best,
Ilya


psung:
Merged as 5dc121a

PR #183, opened 2016-10-26
AZURE-314 Upload empty files support

storozhilov-dnanexus:
Hi @psung ,

this is a solution for AZURE-314 & auto-test - look if have some time, please.

Thank you and all the best,
Ilya

### DIFF
--- a/src/python/dxpy/bindings/dxfile.py
+++ b/src/python/dxpy/bindings/dxfile.py
@@ -586,6 +586,11 @@ class DXFile(DXDataObject):
         defaults to 1. This probably only makes sense if this is the
         only part to be uploaded.
         """
+    
+        # AZURE-314: Do not upload a first (single) empty part if region settings do not allow that
+        if index == 1 and len(data) == 0 and not self._empty_last_part_allowed:
+            return
+
         req_input = {}
         if index is not None:
             req_input["index"] = int(index)
--- a/src/python/dxpy/bindings/dxfile.py
+++ b/src/python/dxpy/bindings/dxfile.py
@@ -538,14 +538,13 @@ class DXFile(DXDataObject):
         self.flush(**kwargs)
 
         # Also populates emptyLastPartAllowed
-        #
-        # TODO: only upload a zero-length part condition on
-        # emptyLastPartAllowed
         self._ensure_write_bufsize(**kwargs)
 
-        if self._num_uploaded_parts == 0:
-            # We haven't uploaded any parts in this session. In case no parts have been uploaded at all, try to upload
-            # an empty part (files with 0 parts cannot be closed).
+        if self._num_uploaded_parts == 0 and self._empty_last_part_allowed:
+            # We haven't uploaded any parts in this session.
+            # In case no parts have been uploaded at all and region
+            # settings allow last empty part upload, try to upload
+            # an empty part (otherwise files with 0 parts cannot be closed).
             try:
                 self.upload_part('', 1, **kwargs)
             except dxpy.exceptions.InvalidState:
@@ -586,11 +585,6 @@ class DXFile(DXDataObject):
         defaults to 1. This probably only makes sense if this is the
         only part to be uploaded.
         """
-    
-        # AZURE-314: Do not upload a first (single) empty part if region settings do not allow that
-        if index == 1 and len(data) == 0 and not self._empty_last_part_allowed:
-            return
-
         req_input = {}
         if index is not None:
             req_input["index"] = int(index)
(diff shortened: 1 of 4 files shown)


src/python/dxpy/bindings/dxfile.py
```diff
@@ -586,6 +586,11 @@ def upload_part(self, data, index=None, display_progress=False, report_progress_
         defaults to 1. This probably only makes sense if this is the
         only part to be uploaded.
         """
+    
+        # AZURE-314: Do not upload a first (single) empty part if region settings do not allow that
+        if index == 1 and len(data) == 0 and not self._empty_last_part_allowed:
```
psung: Would it suffice to add this logic in self.close? It it makes sense, I think it would be a better place for this since (1) there is already a check there to detect the "special case" in S3 of an empty file being uploaded, which seems to be analogous to the case you are trying to handle here and (2) it preserves the property that every invocation of this method uploads exactly one part.

src/python/test/dxpy_testutil.py
```diff
@@ -159,9 +159,16 @@ def temporary_project(name='dx client tests temporary project', cleanup=True, re
...
     :type select: bool
+    :param region:
+        Region name to create a project in. If None the project is created
+        in the default region.
+    :type region: str
 
     """
-    temp_project = dxpy.DXProject(dxpy.api.project_new({'name': name}, **kwargs)['id'])
+    input_params = {'name': name}
+    if region != None:
```
psung: Please use "is not" to compare to None, since None is a singleton.

src/python/test/test_dxclient.py
```diff
@@ -1419,6 +1419,15 @@ def test_dx_upload_mult_paths_with_dest(self):
                 listing = run("dx ls /destdir/a").split("\n")
                 self.assertIn(os.path.basename(fd2.name), listing)
 
+    def test_dx_upload_empty_file(self):
+        testdir = tempfile.mkdtemp()
+        os.mkdir(os.path.join(testdir, 'a'))
+        with testutil.TemporaryFile(dir=os.path.join(testdir, "a")) as fd:
+            fd.close()
+            with temporary_project("test_dx_upload_empty_file default", select=True) as upload_empty_file_project_default:
+                run("dx upload {}".format(fd.name))
```
psung: I'd prefer to test this by invoking one of the dxpy methods directly (e.g. dxpy.upload_local_file) since the code you added is in the DXFile bindings and not in "dx". It will also eliminate the overhead and annoyances associated with making a subprocess call (get better stack traces, etc.).

storozhilov-dnanexus: Done.

I created a new test in `src/python/test/test_dxpy.py`, but didn't remove this one (to be sure it works fine on all layers). Inform me, please, if testing that from the CLI level is not necessary - I would remove this test.

src/python/dxpy/bindings/dxfile.py
storozhilov-dnanexus: Done.

src/python/test/dxpy_testutil.py
storozhilov-dnanexus: Done.

src/python/test/test_dxpy.py
```diff
@@ -458,6 +458,17 @@ def test_upload_string_dxfile(self):
 
         self.assertTrue(filecmp.cmp(self.foo_file.name, self.new_file.name))
 
+    def test_upload_empty_dxfile(self):
+        # Checking default backend
+        dxpy.new_dxfile().close()
+        dxpy.upload_local_file(filename=self.new_file.name, project=self.proj_id, wait_on_close=True)
+        # Checking azure backend
+        with testutil.temporary_project("TestDXFile.test_upload_empty_dxfile azure", select=True, region="azure:westus") as ap:
+            with dxpy.DXFile(project=ap.get_id()) as f:
```
psung: Is the idea to have this be analogous to what's done in the default region above? If so, please create the file in the same way in both regions, just to make it clear that there's no difference. The shortest idiomatic ways are the following:

```
# (1) Use new_dxfile
with new_dxfile(project=...) as my_file:
    pass # Write an empty file
# (2) Use upload_string
my_file_2 = upload_string('', project=...)
```

Also, please verify (in at least one of the cases above) that the file size is 0 after the file is closed.

src/python/test/test_dxclient.py
psung: Thanks, it's good to keep the tests at both levels!

src/python/test/test_dxpy.py
storozhilov-dnanexus: Done.

PR #185, opened 2016-10-28
Allow disabling "watch" on remote builds

godotgildor:
Currently both the asset builder and our remote app/applet builder code paths automatically run "dx watch" on the jobs that are building either the asset or app/applet.   Some customers are sitting behind firewalls that prohibit WebSockets traffic and are therefore unable to run "dx watch".  This causes the asset building and remote app/applet building paths to fail in their environment.

This simple pull request adds a "--skip-watch" flag to these two tools so that one can optionally skip watching the jobs that are building the assets or app/applets.  The code will still wait for the job to complete, but it will remain quiet while the jobs run.

### DIFF
--- a/src/python/dxpy/asset_builder.py
+++ b/src/python/dxpy/asset_builder.py
@@ -222,7 +222,7 @@ def build_asset(args):
 
         if not args.json:
             print("\nStarted job '" + str(job_id) + "' to build the asset bundle.\n", file=sys.stderr)
-            if not args.skip_watch:
+            if args.watch:
                 try:
                     subprocess.check_call(["dx", "watch", job_id])
                 except subprocess.CalledProcessError as e:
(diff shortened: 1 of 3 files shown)


src/python/dxpy/asset_builder.py
```diff
@@ -222,18 +222,19 @@ def build_asset(args):
...
-                if e.returncode == 3:
-                    # Some kind of failure to build the asset. The reason
-                    # for the failure is probably self-evident from the
-                    # job log (and if it's not, the CalledProcessError
-                    # is not informative anyway), so just propagate the
-                    # return code without additional remarks.
-                    sys.exit(3)
-                else:
-                    raise e
+            if not args.skip_watch:
```
psung: To avoid this kind of double negative you can define the flag as:

```
add_argument('--no-watch', ..., action="store_false", dest="watch")
```

and then do:

```
if args.watch:
    # dx watch
```

Also, in most cases "negative" flags should be named "--no-FOO".

godotgildor:
Thanks @psung.  I just submitted a commit that changes the name to --no-watch and removes double negatives.  Let me know if there's anything else that's needed.

PR #187, opened 2016-11-10
STACK-197: Adding logging of dx-docker subcommands

geetduggal:
@nicolasbockg Could you take a look at this when you have a chance?  Thanks in advance!

### DIFF
--- a/src/python/scripts/dx-docker
+++ b/src/python/scripts/dx-docker
@@ -115,6 +115,13 @@ def get_aci(image, imgname, prefix="", quiet=False):
     os.rename(glob.glob(tmpdir+"/*.aci")[0], img_dirname+"/"+urllib.quote_plus(image)+".aci")
     shutil.rmtree(tmpdir)
 
+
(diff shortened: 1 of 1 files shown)


src/python/scripts/dx-docker
```diff
@@ -115,6 +115,13 @@ def get_aci(image, imgname, prefix="", quiet=False):
     os.rename(glob.glob(tmpdir+"/*.aci")[0], img_dirname+"/"+urllib.quote_plus(image)+".aci")
     shutil.rmtree(tmpdir)
 
+
+def register_docker_subcommand(cmd):
+    orig_user_agent = dxpy.USER_AGENT
+    dxpy.USER_AGENT += " dx-docker {}".format(cmd)
+    dxpy.whoami()
```
nicolasbockg: As we discussed maybe use system/greet instead.

geetduggal: Thanks for the suggestion!  Done.

PR #188, opened 2016-11-14
Adding stages to workflows

orodeh:
Allowing stages to be added to workflows from the java API.


orodeh:
@psung, please treat this an a Non Pull Request (NPR). See if you like the overall direction, and write feedback. I will send a Pull Request as a second step.

psung:
This seems to underscore how annoying it is to extract the stage outputs (in particular that you need to know the stage IDs to deserialize the output). It does seem possible to use the `stages` field of the analysis describe, and then you can sort of address them by numeric index in order to retrieve the outputs: `stages[0].execution.output.OUTPUTFIELD`.

That's not a problem with this code, though. I think it looks like a nice approach, so I'd be happy to review it for real whenever you like.

orodeh:
Great! Please treat this as a full fledged pull request. It is ready for review.

I will bear in mind the idea about stage output retrieval. I will probably need it, or something like it, when implementing getStage.

src/java/src/main/java/com/dnanexus/DXWorkflow.java
```diff
@@ -239,4 +242,42 @@ public Describe getCachedDescribe() {
...
+        public int editVersion;
+
+        @JsonProperty
+        public String stage;
+    }
+
+    public DXStage addStage(DXApplet applet,
+                            String name,
+                            Object stageInputs,
+                            int editVersion) {
```
psung: The workflow should probably have a method to retrieve its own editVersion number, since if you implement any other methods on workflows, their callers will also need to supply editVersion. In that case DXStage doesn't need its own editVersion field.

Do you think this class should manage the editVersions itself (i.e. remember the last editVersion of each operation, and supply it on the next operation)? Or should that responsibility be passed on to the user (as you've done so far)?

src/java/src/test/java/com/dnanexus/DXWorkflowTest.java
```diff
@@ -17,12 +17,14 @@
 package com.dnanexus;
 
 import java.io.IOException;
+import java.util.*;
```
psung: Please expand this out to all the classes that are imported.

src/java/src/test/java/com/dnanexus/DXWorkflowTest.java
```diff
@@ -179,36 +152,74 @@ public void testRunWorkflow() {
...
+        } catch (Exception e) {
+            System.err.println("Caught exception" + e.getStackTrace());
+        }
+    }
 
-        WorkflowAddStageInput addStage1Input = new WorkflowAddStageInput();
-        addStage1Input.editVersion = addStage0Output.editVersion;
-        addStage1Input.executable = applet.getId();
+    @Test
+    public void testRunWorkflowWithDependencies() {
```
psung: This test should only run conditional on RUN_JOBS.

src/java/src/test/java/com/dnanexus/DXWorkflowTest.java
```diff
@@ -179,36 +152,74 @@ public void testRunWorkflow() {
...
-        analysis.terminate();
+        analysis.waitUntilDone();
+
+        // The results are supposed to be something like this:
+        //{
+        //   "stage-F0b4zz807vqPqYzGJbxQ712k.sum" : 3,
+        //   "stage-F0b4zzQ07vqFgvfJ5xfjPg16.sum" : 7
+        //}
+        ObjectNode jnode = analysis.getOutput(ObjectNode.class);
+        prettyPrintJsonNode(jnode);
```
psung: Please add an assertion that this output is what you want it to be.

orodeh:
@psung, I think I addressed the comments. One thing I was not able to do, is make use of the "self-serialization" of DXDataObjects. I made the makeDXLink public instead. This, at least, avoids code duplication.

orodeh:
@psung, are there more comments? 

If this is good, then I can merge it, and work on the next improvement.

src/java/src/main/java/com/dnanexus/DXDataObject.java
```diff
@@ -1062,7 +1062,7 @@ public Describe getCachedDescribe() {
      * @return a DNAnexus link
      */
     @JsonValue
-    private JsonNode getDXLink() {
+    public JsonNode getDXLink() {
```
psung: Let's call this `getLinkAsJson` since we might want to have a `getDXLink` in the future that returns some native class (implemented by us) representing a link.

Take the part about the JSON serializer and remove it from the public docstring for this method. You could move it in some form to the main docstring for this class...

> Instances of this class are JSON-serialized as a DNAnexus link, which makes it suitable to provide an objects with references to DXDataObjects as the input when running a DXApplet, etc.

src/java/src/main/java/com/dnanexus/DXUtil.java
```diff
@@ -20,6 +20,7 @@
 import java.io.IOException;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
```
psung: Revert the changes to this class?

src/java/src/main/java/com/dnanexus/DXWorkflow.java
```diff
@@ -239,4 +243,101 @@ public Describe getCachedDescribe() {
         return ExecutableRunner.getWorkflowRunnerWithEnvironment(this.getId(), this.env);
     }
 
+    /**
+     * A workflow stage
+     *
+     */
+    public static class DXStage {
```
psung: I prefer `DXWorkflow.Stage` to `DXWorkflow.DXStage`... other nested classes also aren't named with a "DX" prefix.

src/java/src/main/java/com/dnanexus/DXWorkflow.java
```diff
@@ -239,4 +243,101 @@ public Describe getCachedDescribe() {
...
+
+        /**
+         * Create a link to an output field. This is used in workflows, to
+         * link results between stages.
+         *
+         * @param fieldName  name of an output field
+         *
+         * @return JSON representation of a link. Can be used as an input to a workflow stage.
+         */
+        public ObjectNode getOutputReference(String  fieldName) {
```
psung: Maybe make this declare return type `JsonNode` to be more similar to `getDXLink`...

Also, `fieldName` => `outputName`, and similarly `inputName` below

src/java/src/main/java/com/dnanexus/DXWorkflow.java
```diff
@@ -239,4 +243,101 @@ public Describe getCachedDescribe() {
...
+        DXStage(String ID) {
+            this.ID = ID;
+        }
+
+        public String getId() {
+            return ID;
+        }
+
+        /**
+         * Create a link to an output field. This is used in workflows, to
```
psung: Move the second sentence to a different paragraph (broken off by `<p>...</p>`)

src/java/src/main/java/com/dnanexus/DXWorkflow.java
```diff
@@ -239,4 +243,101 @@ public Describe getCachedDescribe() {
...
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    private static class WorkflowAddStageOutput {
+        @JsonProperty
+        public int editVersion;
+
+        @JsonProperty
+        public String stage;
+    }
+
+    public Modification<DXStage> addStage(DXApplet applet,
```
psung: Add a docstring here.

src/java/src/main/java/com/dnanexus/DXWorkflow.java
```diff
@@ -239,4 +243,101 @@ public Describe getCachedDescribe() {
...
+                .put("inputField", fieldName).build();
+            return DXJSON.getObjectBuilder().put("$dnanexus_link", dxlink).build();
+        }
+    }
+
+    /** Some workflow update operations return an edit version. This class is a way
+     *  to report the new edit version, together with the operation result.
+     */
+    public static class Modification<T> {
+        public int editVersion;
```
psung: Please provide getters for these fields `getEditVersion`, `getValue` and make them `final`.

src/java/src/test/java/com/dnanexus/DXWorkflowTest.java
```diff
@@ -179,36 +158,104 @@ public void testRunWorkflow() {
...
+        //}
+        ObjectNode jnode = analysis.getOutput(ObjectNode.class);
+        java.util.Iterator<java.util.Map.Entry<String, JsonNode>> iter = jnode.fields();
+        while (iter.hasNext()){
+            java.util.Map.Entry<String, JsonNode> pair = iter.next();
+            String key = pair.getKey();
+            int val = pair.getValue().asInt();
+
+            if (key.contains(stage1.getId())) {
+                Assert.assertTrue(val == 3);
```
psung: Use `assertEquals` so you get a more useful message if the assertion fails.

orodeh:
I believe the issues have been fixed. Please take a look. 

Thanks, 
      Ohad.

src/java/src/main/java/com/dnanexus/DXDataObject.java
```diff
@@ -42,7 +42,10 @@
 import com.google.common.collect.Maps;
 
 /**
- * Base class for all data object classes in the DNAnexus Platform.
+ * Base class for all data object classes in the DNAnexus
+ * Platform. Instances of this class are JSON-serialized as a DNAnexus
```
psung: (Here and elsewhere...) Per the [Javadoc conventions](http://www.oracle.com/technetwork/articles/java/index-137868.html), the first sentence of the docstring should stand alone (since the class summary or method list will just show this one sentence) and subsequent commentary should be set off with `<p>...</p>`.

PR #189, opened 2016-11-24
AZURE-82 SSH to worker support

storozhilov-dnanexus:
Hi @psung ,
this is a proposed solution for AZURE-82 on **dx-toolkit** side. 

I merged testing AWS and Azure in one test - does it makes sense (the code is the same, only project id and instance type differs)? It looks all tests should check all supported providers, so some kind of additional work is needed to test everything...

I couldn't test it using both cloud providers, because **staging** does not have my Nucleus modifications installed, but **branchtest** does not work with AWS.

Thank you and all the best,
Ilya


psung:
Thanks! I factored out the region-specific code into a test method that is then called from both tests (one for each cloud), so that we can run one or both tests in each test environment, depending on what regions are enabled.

PR #190, opened 2016-12-14
PTFM-17112 - Allow remote builds on 14.04 workers

geetduggal:
@psung please take a look at your convenience, and thanks in advance!  This is the first draft at allowing for remote builds on 14.04 workers and I also created a companion pull request for the 'dx_app_builder' repository.

The entire test suite was run and I added a simple test for building remotely on 14.04 workers in addition.

### DIFF
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -576,6 +576,10 @@ def _build_app_remote(mode, src_dir, publish=False, destination_override=None,
     else:
         builder_app = 'app-tarball_applet_builder'
 
+    app_spec = _parse_app_spec(src_dir)
+    if 'distribution' in app_spec['runSpec'] and 'release' in app_spec['runSpec'] and app_spec['runSpec']['release'] == '14.04':
+        builder_app += "_trusty"
+
     temp_dir = tempfile.mkdtemp()
 
     build_options = {'dx_toolkit_autodep': dx_toolkit_autodep}
@@ -590,7 +594,6 @@ def _build_app_remote(mode, src_dir, publish=False, destination_override=None,
         # before uploading, we need to determine the correct version to
         # use here and pass it in to the interior run of dx-build-app.
         if do_version_autonumbering:
-            app_spec = _parse_app_spec(src_dir)
             original_version = app_spec['version']
             app_describe = None
             try:
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -577,7 +577,7 @@ def _build_app_remote(mode, src_dir, publish=False, destination_override=None,
         builder_app = 'app-tarball_applet_builder'
 
     app_spec = _parse_app_spec(src_dir)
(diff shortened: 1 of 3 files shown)


src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -576,6 +576,10 @@ def _build_app_remote(mode, src_dir, publish=False, destination_override=None,
     else:
         builder_app = 'app-tarball_applet_builder'
 
+    app_spec = _parse_app_spec(src_dir)
+    if 'distribution' in app_spec['runSpec'] and 'release' in app_spec['runSpec'] and app_spec['runSpec']['release'] == '14.04':
```
psung: Seems like `distribution` is presumed to be "Ubuntu", so it should be legal to just write `{"release": "14.04", ...}` (without `distribution`) in your `runSpec` and have it be treated the same way.

src/python/test/test_dxclient.py
```diff
@@ -5831,6 +5831,19 @@ def test_remote_build_app(self):
         app_dir = self.write_app_directory("minimal_remote_build_åpp", json.dumps(app_spec), "code.py")
         run("dx build --remote --app " + app_dir)
 
+    def test_remote_build_app_trusty(self):
+        app_spec = {
+            "name": "minimal_remote_build_app_trusty",
+            "dxapi": "1.0.0",
+            # Use a package specific to trusty but not in precise as part of the execdepends to ensure it is installed properly
+            "runSpec": {"file": "code.py", "interpreter": "python2.7", "distribution": "Ubuntu", "release": "14.04", "execDepends": [{"name": "postgresql-9.3"}], "systemRequirements": {"*": {"instanceType": "mem1_ssd1_x4"}}},
```
psung: Test needs to use "buildDepends" since that's what the builder app is going to look at. (I think it ignores "execDepends", so I suspect this test would pass as written even if the build happened on 12.04.)

src/python/test/test_dxclient.py
```diff
@@ -5831,6 +5831,19 @@ def test_remote_build_app(self):
         app_dir = self.write_app_directory("minimal_remote_build_åpp", json.dumps(app_spec), "code.py")
         run("dx build --remote --app " + app_dir)
 
+    def test_remote_build_app_trusty(self):
```
psung: Can you add a test of building an applet, too?

src/python/dxpy/scripts/dx_build_app.py
geetduggal: Thanks, Done.

src/python/test/test_dxclient.py
geetduggal: Yes, thanks for pointing out this oversight.  Done!

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -576,6 +576,10 @@ def _build_app_remote(mode, src_dir, publish=False, destination_override=None,
     else:
         builder_app = 'app-tarball_applet_builder'
 
+    app_spec = _parse_app_spec(src_dir)
+    if 'release' in app_spec['runSpec'] and app_spec['runSpec']['release'] == '14.04':
```
psung: Could also use

```
if app_spec['runSpec'].get('release') == '14.04':
```

PR #192, opened 2016-12-22
dx-docker: Handle workdir properly when overriding image metadata

geetduggal:
@psung Please take a look at your convenience. This is a quick fix to `dx-docker` so that we handle the working directory properly when overriding image metadata. 

I ran into this when testing out the utility in a way I haven't before.  I also added a test to the test suite in a companion pull request.  Thanks in advance!

### DIFF
--- a/src/python/scripts/dx-docker
+++ b/src/python/scripts/dx-docker
@@ -210,9 +210,11 @@ def run(args):
 
     container_cmd = entrypoint + cmd
 
-    workdir = args.workdir
+    workdir = "/"
     if 'workingDirectory' in imagemeta['app']:
        workdir = imagemeta['app']['workingDirectory']
+    if args.workdir:
+       workdir = args.workdir
 
     env = []
     if 'environment' in imagemeta['app']:


psung:
LGTM

src/python/scripts/dx-docker
```diff
@@ -210,9 +210,11 @@ def run(args):
 
     container_cmd = entrypoint + cmd
 
-    workdir = args.workdir
+    workdir = "/"
     if 'workingDirectory' in imagemeta['app']:
        workdir = imagemeta['app']['workingDirectory']
+    if args.workdir:
```
mlin: Hi @geetduggal there's a new problem which is that `args.workdir` has a default setting of `/` so this condition is always true. Probably there should be no default value of `args.workdir`.

This broke viral-ngs for @dpark @tomkinsc I'm afraid...eg

```
$ dx-docker run -v /home/dnanexus:/user-data broadinstitute/viral-ngs:1.13.3 metagenomics.py
Image Repo      broadinstitute/viral-ngs
Image Tag       1.13.3
Image Size      3.6G
Image ID        sha256:1466e5a40aa0bd6b68b2285954bcd33002d19bb9e1ba09c3807a93fa3337bcd6
Last Updated    2016-12-13T16:21:10.093282599Z (33d 8h ago)
Registry        registry-1.docker.io
ENTRYPOINT      ["./env_wrapper.sh"]


/bin/sh: 1: ./env_wrapper.sh: not found
```

Can this be hotfixed?

Side comment, perhaps WORKDIR is worth showing in the log message too.

tomkinsc: For this container, the working directory (or at least the one containing `env_wrapper.sh`) is the home folder of our non-root container user, `~viral-ngs-user`.

geetduggal: Thanks for catching this! Looks like your comments are spot on. I am currently testing out a fix and adding a few useful test cases. Also outputting metadata for the working directory for future reference :).

PR #196, opened 2017-01-23
Add ZSH compatibility for environment shell script

TMiguelT:
zsh is generally compatible with bash, but I had to add zsh specific code for:

- Finding the current filename
- Adding autocomplete
- Enabling nullglob

And I changed the `==` operator to the identical `=` operator which also works with zsh


jtratner:
this would be super helpful! are there blockers to merging this?

kpjensen:
@jtratner this is now included in the latest stable release v0.249.0

PR #198, opened 2017-02-01
AZURE-349 new ABS backend

storozhilov-dnanexus:
Hi @psung ,

this is my `dx-toolkit` updates for AZURE-349 - review at your convenience, please.

All the best,
Ilya


psung:
Closing this but I will open a discussion regarding the test you added (which I think should be folded into a set of mini-benchmarks).

PR #200, opened 2017-02-02
DEVEX-88: Fix for dx download fail with both overwrite and recursive …

commandlinegirl:
…options set

### DIFF
--- a/src/python/test/test_dxclient.py
+++ b/src/python/test/test_dxclient.py
@@ -1192,6 +1192,21 @@ class TestDXRename(DXTestCase):
 
 
 class TestDXClientUploadDownload(DXTestCase):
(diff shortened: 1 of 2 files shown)


src/python/test/test_dxclient.py
```diff
@@ -1192,6 +1192,21 @@ def test_rename(self):
 
 
 class TestDXClientUploadDownload(DXTestCase):
+    def test_dx_download_recursive_overwrite(self):
+        if os.path.exists("foodir/"):
+            run("rm -rf foodir")
+        wd = "foodir"
```
geetduggal: Please move `wd = "foodir"` to the top and use it in the subsequent strings.  Thanks!

PR #204, opened 2017-02-16
Mark routes as retryable when a nonce is supplied

psung:
Routes that supply a nonce should be retryable (this being the whole
point of adding the nonce). We don't do this by marking such routes as
retryable in the wrapper_table.json, since not all the language bindings
know how to generate the nonces, and this could result in bindings that
didn't supply the nonce but went ahead and considered the route to be
retryable anyway. Therefore, the *bindings generators* for each language
are responsible for marking each route as retryable if and only if they
inserted nonce generation code.

In particular, note that in the Java bindings, each API route has
multiple overloads and we added nonce support in only some of them (not
bothering to add nonce support in the older and deprecated overloads).
In this case, *only* the overloads that specifically support nonces also
mark the request as retryable.

### DIFF
--- a/src/api_wrappers/generatePythonAPIWrappers.py
+++ b/src/api_wrappers/generatePythonAPIWrappers.py
@@ -109,8 +109,8 @@ print(preamble)
 for method in json.loads(sys.stdin.read()):
     route, signature, opts = method
     wrapper_method_name = camel_case_to_underscore(signature.split("(")[0])
-    retry = "True" if (opts['retryable']) else "False"
-    accept_nonce = True if 'acceptNonce' in opts else False
+    accept_nonce = opts.get('acceptNonce', False)
+    retry = opts['retryable'] or accept_nonce
     if (opts['objectMethod']):
         root, oid_route, api_method_name = route.split("/")
         if oid_route == 'app-xxxx':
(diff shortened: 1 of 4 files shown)


src/api_wrappers/generatePythonAPIWrappers.py
```diff
@@ -109,8 +109,8 @@ def camel_case_to_underscore(name):
 for method in json.loads(sys.stdin.read()):
     route, signature, opts = method
     wrapper_method_name = camel_case_to_underscore(signature.split("(")[0])
-    retry = "True" if (opts['retryable']) else "False"
-    accept_nonce = True if 'acceptNonce' in opts else False
+    accept_nonce = opts.get('acceptNonce', False)
+    retry = opts['retryable'] or accept_nonce
```
geetduggal: The logic makes sense.  Is there a case where a call is not retryable but does accept a Nonce, or is this more of a historical artifact? https://github.com/dnanexus/dx-toolkit/blob/master/src/api_wrappers/wrapper_table.json#L533

psung: The way it's done for now is, all nonce-taking routes are marked as **not** retryable in wrapper_table.json, and then this line makes it so that they are actually eligible for retry, in these bindings (so, exactly the situation you are asking about, I think). That means there are sort of three classes of routes: not retryable (`retryable=false`), retryable with a nonce (`retryable=false` in the wrapper table but it's marked as retryable in the bindings), and retryable (`retryable=true`).

If at some point we decide that all language bindings are required to implement the nonce support correctly, then we could remove this special logic here and just mark all nonce-taking routes as actually retryable.

psung:
Thanks! If you wouldn't mind, please merge this at your convenience.

PR #205, opened 2017-02-22
Devex 3 upload agent resume message

kpjensen:
Simple check if a file is 100% uploaded


kpjensen:
I added a small test in cells to check for the proper output message.

PR #209, opened 2017-03-17
Adding option for supressing running check when exiting ssh

godotgildor:
We occasionally set up a tunnel over ssh to a worker.  When setting up the tunnel,
we want to exit the ssh command with the tunnel still open.  Today, when dx sees the
ssh exit, it prompts users to ask if they'd like to terminate the job.  We'd like to
be able to optionally supress this inquiry.

This commit adds a hidden command line to ssh to allow users to optionally supress this
prompt.


psung:
Seems fine to me, can you add a comment (not a helpstring, just a comment of the "# ..." variety) right next to the flag definition with your motivating use case?

godotgildor:
O.K.  I just added a comment.  Let me know if it needs any further clarification.

PR #210, opened 2017-03-20
dx-docker makedirs

kpjensen:
Other OSError exceptions could be raised, but we only want to pass if the directory exists. E.g. permission could be denied, and currently would be passed.


kpjensen:
@geetduggal Can you glance at this?

PR #211, opened 2017-04-04
Fix inconsistent version numbers in docs

geekdave:
Update docs to use consistent version number for `gcc` dependency install


geekdave:
Updated macports section as well.

kpjensen:
We should probably update this Readme in more detail since we build with clang internally now.

PR #216, opened 2017-05-23
dxcpp supports upload v2

psung:
Not working in its current state, but maybe you can figure out where I went wrong.


orodeh:
@kpjensen, is this still relevant?

PR #227, opened 2017-07-27
Notebook playground

godotgildor:
I have created a feature that allows users to spin up a jupyter or rstudio notebook in our platform and then access it via an ssh tunnel between their local machine and the worker running the server.  

Currently we would like to make it available to a few customers to try out, but are not quite ready to advertise it globally yet.  For this pull request, it will add the ability to run `dx notebook`, but it won't advertise the feature through `dx help all`.


geetduggal:
This is awesome.  Thanks Brett!

PR #228, opened 2017-07-28
PTFM-20076: Add dx-toolkit support for downloadRestricted flag

robosoto:
Add --download-restricted flag to dx-toolkit.

Test plan: run_integration_tests.sh --client_tests_only --tests test_dxpy.TestDXProject

### DIFF
--- a/src/python/dxpy/bindings/dxproject.py
+++ b/src/python/dxpy/bindings/dxproject.py
@@ -283,7 +283,7 @@ class DXProject(DXContainer):
     _class = "project"
 
     def new(self, name, summary=None, description=None, protected=None,
-            restricted=None, contains_phi=None, tags=None,
+            restricted=None, download_restricted=None, contains_phi=None, tags=None,
             properties=None, bill_to=None, **kwargs):
         """
         :param name: The name of the project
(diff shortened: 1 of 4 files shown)


src/python/dxpy/bindings/dxproject.py
```diff
@@ -296,6 +296,8 @@ def new(self, name, summary=None, description=None, protected=None,
         :type protected: boolean
         :param restricted: If provided, whether the project should be restricted
         :type restricted: boolean
+        :param download_restricted: If provided, whether the project should be download restricted
```
nicolasbockg: Consider rewording this as: "If provided, whether external downloads should be restricted"

src/python/dxpy/bindings/dxproject.py
```diff
@@ -348,6 +352,8 @@ def update(self, name=None, summary=None, description=None, protected=None,
...
```
nicolasbockg: Same here.

src/python/dxpy/bindings/dxproject.py
robosoto: Good call

nicolasbockg:
LGTM. 
@geetduggal Just FYI, we are adding a `downloadRestricted` Flag to projects.

PR #230, opened 2017-08-10
dx-app-wizard python test template failure fix. test added. (#229)

Damien-Black:
**Motivation**

The `dx-app-wizard` script generates a `test/test.py` functional test template for apps. This is very useful; however, after [this commit](https://github.com/dnanexus/dx-toolkit/commit/545f4ff3f77cd4c8d25a878c2e7fd84b721cc75b) the script it generates doesn't run due to:

* *dxpy.executable_builder* import error, logic for `inline_documentation_files` was moved here. Since we call *dxpy.app_builder* from *dx_build_app.py* we never saw the missing *dxpy.executable_builder* import.
* Minor typo (wrong var used) in [this commit](https://github.com/dnanexus/dx-toolkit/commit/75ccb04c9e4f1e809dc3ff473bec95c3c2e5f07d#diff-d38613cee6417a94e5af7a9ccf728beeR532).

**Consideration**

Work-around does exist, import *dxpy.executable_builder* and specify `project` param in `dxpy.app_builder.upload_applet` in the actual `test.py` script.

Thought about using and importing `dxpy.scripts import build_and_upload_locally` directly. But it seems `dxpy.app_builder.upload_applet` is meant to be exposed. 

**Test**

Test ran [LINK](https://jenkins2.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging/78/). The one failing integration test fails on [Master too](https://jenkins2.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging/81/console).

### DIFF
--- a/src/python/test/test_dx_app_wizard_and_run_app_locally.py
+++ b/src/python/test/test_dx_app_wizard_and_run_app_locally.py
@@ -205,6 +205,19 @@ class TestDXAppWizardAndRunAppLocally(DXTestCase):
         result = remote_job.describe()
         self.assertEqual(result["output"]["out1"], 140)
 
+    def test_dx_verify_build_app(self):
+        appdir = create_app_dir()
+        print("Setting current project to", self.project)
+        dxpy.WORKSPACE_ID = self.project
+        dxpy.PROJECT_CONTEXT_ID = self.project
+        bundled_resources = dxpy.app_builder.upload_resources(appdir)
+        applet_id, _ignored_applet_spec = dxpy.app_builder.upload_applet(appdir, bundled_resources, overwrite=True, dx_toolkit_autodep=False)
+        app_obj = dxpy.DXApplet(applet_id)
+        try:
+            app_obj.describe()
+        except dxpy.exceptions.ResourceNotFound as dxrne:
+            self.fail(dxrne.error_message())
+
     def test_file_download(self):
         '''
         This test assumes a well-formed input spec and tests that the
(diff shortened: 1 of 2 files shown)


src/python/test/test_dx_app_wizard_and_run_app_locally.py
```diff
@@ -205,6 +205,19 @@ def test_dx_run_app_locally_and_compare_results(self):
...
+        dxpy.WORKSPACE_ID = self.project
+        dxpy.PROJECT_CONTEXT_ID = self.project
+        bundled_resources = dxpy.app_builder.upload_resources(appdir)
+        applet_id, _ignored_applet_spec = dxpy.app_builder.upload_applet(appdir, bundled_resources, overwrite=True, dx_toolkit_autodep=False)
+        app_obj = dxpy.DXApplet(applet_id)
+        try:
+            app_obj.describe()
+        except dxpy.exceptions.ResourceNotFound as dxrne:
+            self.fail(dxrne.error_message())
+
```
commandlinegirl: It would be useful to have a test name that describes better what exactly is being tested test, for example test_upload_applet_with_current_context_project (or sth similar).

Since we expect the build to succeed, we can add an assert, for example:
`self.assertEquals(app_obj.describe()['id'], app_obj.get_id())`
or
`self.assertEquals(app_obj.describe()['project'], dxpy.WORKSPACE_ID)`

Damien-Black:
Good point on the test name. Updated the test per your specs.

PR #237, opened 2017-09-19
PTFM-20347 TLS version check

nicolasbockg:
Added the ability in the ua to display system messages as part of the
ua --test flag. The system/greet route will return messages regarding
the TLS version used to establish the connection.

### DIFF
--- a/src/ua/ua_test.cpp
+++ b/src/ua/ua_test.cpp
@@ -161,7 +161,7 @@ void testWhoAmI(){
   } catch (DXConnectionError &e) {
     cout << "Error contacting the api: " << e.what() << endl;
   } catch (JSONException &e) {
-    cout << "Error contacting the api " << e.what() << endl;
+    cout << "Error contacting the api: " << e.what() << endl;
   } catch (...) {
     cout << "Error contacting the api." << endl;
   }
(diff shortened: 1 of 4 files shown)


nicolasbockg:
Api server changes in https://github.com/dnanexus/nucleus/pull/676

src/ua/ua_test.cpp
```diff
@@ -149,14 +152,54 @@ void certificateFile() {
...
     JSON res = systemWhoami(string("{}"), false);
     cout  << res["id"].get<string>() << endl;
   } catch(DXAPIError &e) {
     cout << "Error contacting the api: " << e.what() << endl;
   } catch (DXConnectionError &e) {
     cout << "Error contacting the api: " << e.what() << endl;
+  } catch (JSONException &e) {
+    cout << "Error contacting the api " << e.what() << endl;
+  } catch (...) {
+    cout << "Error contacting the api." << endl;
```
geetduggal: Pedantically minor: there is a period at the end of this default message but are there for the other exceptions?

nicolasbockg: No, the e.what() does not have a period at the end. Should I add one to all?

PR #241, opened 2017-09-27
DEVEX-436: Change workflowInputSpec to inputs and workflowOutputSpec to outputs

commandlinegirl:
This changes naming so that it is synced up with the apiserver.


commandlinegirl:
Let me please merge this change before the release in order to keep the API change and toolkit in line. This ticket is about changing naming. I tested it manually and in a private branch on jenkins. If there's anything found in the review we can revert it before it goes to prod.

PR #242, opened 2017-09-27
DEVEX-391: Include info about Azure instance types in app wizard

commandlinegirl:


commandlinegirl:
Thank you for the review. I will add that check. 

We also added mem3_ssd2 instance types (AWS) (they are available from the UI).

PR #246, opened 2017-10-03
Updating dx-app-wizard

godotgildor:
Moving the instance type from being in runSpec to being in regionalOptions
in the template dxapp.json produced from dx-app-wizard.  This will better
setup users to make their apps multi-region.


geetduggal:
Thanks for this There are a few changes we need to make to this that @commandlinegirl noticed.  She will take over this branch and add to it so we can get this change in.

commandlinegirl:
Thanks for this PR! The change was included in https://github.com/dnanexus/dx-toolkit/commit/093fc83a9805ee5a67b5fcf4203295c3a37e170f, so I'll close it. Thanks.

PR #250, opened 2017-10-11
DEVEX-476  env for 4 http proxy variables

kpjensen:


src/python/dxpy/__init__.py
```diff
@@ -244,6 +244,13 @@ def _get_proxy_info(url):
 
     return proxy_info
 
+def _get_env_var_proxy():
+  proxy_tuple = ('http_proxy', 'HTTP_PROXY', 'https_proxy', 'HTTPS_PROXY')
+  proxy = None
+  for env_proxy in proxy_tuple:
+    if env_proxy in os.environ:
+      proxy = os.environ[env_proxy]
+  return proxy
```
geetduggal: May want to return the first match here and print a message about what env var is being used in case it matters.  I don't know if this happens often but you can imagine a case where the env variables above are set to different URLs.

PR #253, opened 2017-10-12
 DEVEX-330: Update app building tests to include Ubuntu release

commandlinegirl:


commandlinegirl:
Since a lot of tests are affected, testing in a branch right now. Also tested locally with api server temporarily modified to require release spec. Thanks for taking a look!

PR #256, opened 2017-10-16
DEVEX-362: Throw an error when old pricing policy scheme is used

commandlinegirl:


ptn24:
LGTM. I prefer a warning message, but this is fine too. Thank you!

commandlinegirl:
I agree, a warning is better here and more in line with the existing dxapp.json validation. Updated. Thanks!

PR #273, opened 2017-11-07
dx-docker: Cleanup of quote code. Potential regression fix

geetduggal:
Hi @commandlinegirl this is a PR from a branch that I had lying around to clean up some code I wrote a while back.  Interestingly, I think it might also resolve a regression that was introduced by that commit.  I added the complex quote example to the tests as well.

### DIFF
--- a/src/python/scripts/dx-docker
+++ b/src/python/scripts/dx-docker
@@ -252,11 +252,11 @@ def run(args):
 
     if args.rm and not args.quiet:
         sys.stderr.write(dxpy.utils.printing.YELLOW("--rm: Note that by default the container is not preserved after execution\n"))
-        
+
     if args.user and not args.quiet:
         sys.stderr.write(dxpy.utils.printing.YELLOW("--user: Note that the current implementation ignores the user command\n"))
 
-        
+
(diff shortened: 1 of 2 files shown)


src/python/scripts/dx-docker
```diff
@@ -271,35 +271,15 @@ def run(args):
...
-    # Note that we assume /dx-docker-runme.sh doesn't already exist in the container
-    bounded_dirs.extend(['-b', tempsh_file_path + ':/dx-docker-runme.sh'])
-
     proot_cmd =  ["proot", '-r', '{tmpdir}/rootfs'.format(tmpdir=tmpdir)] + \
                  bounded_dirs + \
                  volume + \
                  ["-w", workdir] + \
-                 ['/dx-docker-runme.sh']
+                 env + ["/bin/sh", "-c"] + \
+                 [subprocess.list2cmdline(container_cmd)]
```
commandlinegirl: Correct me if I'm wrong: if container_cmd is an empty list (entrypoint & cmd were not set), then this will not error out?

geetduggal:
Thanks! I updated the script to throw an exception in that case.   Tests pass

PR #281, opened 2017-11-28
Remove error that blocks org from being added as developers

msimbirsky:
Orgs as developers is supported now at API level so this operation should not be
blocked


commandlinegirl:
Thanks for this update! Please, remove this from tests, thanks:
https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/test_dxclient.py#L7153

msimbirsky:
Added the change requested by Ola

msimbirsky:
Thanks for the catch!

PR #287, opened 2018-01-18
Tip 51 database apis

durae:
New database api routes available from dx-toolkit through dx api ..., or through commands specified for data objects. Bindings or other supported languages were generated by make api_wrappers


ksoni:
Looks good, merging now.

PR #291, opened 2018-01-29
Batch CLI: wrapper to generate inputs API call

geetduggal:
This PR is a simple wrapper around the upcoming `generateBatchInputs` API call along with some basic tests and corner cases.   It has been tested in local environments with the API call available and works as expected.


src/python/dxpy/utils/resolver.py
```diff
@@ -1240,3 +1240,16 @@ def resolve_to_objects_or_project(path, all_matching_results=False):
                                   if you were attempting to refer to a project by name,
                                   please append a colon ":" to indicate that it is a project.''')
     return project, folderpath, entity_results
+
+# Generic function to parse an input key-value pair of the form '-ikey=val'
+# e.g. returns ("key", "val") in the example above
+def parse_input_keyval(keyeqval):
```
orodeh: Nice refactoring.

src/python/dxpy/scripts/dx.py
```diff
@@ -3444,6 +3445,69 @@ def upgrade(args):
...
+    input_names = input_dict.keys()
+
+    # Output TSV Batch
+    for i,batch in enumerate(batches):
+        def flatten_batch(b):
+            return [b['batchPattern']] + [ival['name'] for iname, ival in sorted(b['inputs'].items())] + [ival['ids'][0] for iname, ival in sorted(b['inputs'].items())]  
+
+        batch_fname = "{}.{:04d}.tsv".format(args.output_prefix, i)
+        with open(batch_fname, 'wb') as csvfile:
+            batchwriter = csv.writer(csvfile, delimiter='\t'.encode('ascii'))
```
orodeh: What kind of encoding are we supposed to use here? I used utf-8. I am now not sure which is correct.

src/python/dxpy/scripts/dx.py
```diff
@@ -3444,6 +3445,69 @@ def upgrade(args):
...
+
+    batches = list(chunks(successful, MAX_BATCH_SIZE))
+
+    eprint("Found {num_success} valid batch IDs matching desired pattern.".format(num_success=len(successful)))
+
+    input_names = input_dict.keys()
+
+    # Output TSV Batch
+    for i,batch in enumerate(batches):
+        def flatten_batch(b):
```
orodeh: Please add a comment, perhaps: "this method generates a line in a TSV file".

src/python/dxpy/scripts/dx.py
geetduggal: So if I recall, If I didn't encode this delimiter as ascii, the `csv.writer` errors out.  I'm assuming that since we import `unicode_literals` from `__future__` the '\t' string is UTF-8 which causes the writer to error out and that it requires an ASCII character delimiter.

geetduggal:
Thanks @orodeh! I addressed your comments.  I'll merge this when our API changes are available on staging.

Issue #286, opened 2018-01-10
dnanexus java api: infinite socket timeouts


vpiyanov:
DXSearch call can hang in the ``SocketInputStream.socketRead0()`` if network connection is unstable.
The root cause is that DXHTTPRequest creates Apache HttpClient with default parameters. As result the infinite socket timeout is used.

The issue can be reproduce with a synthetic test case:
```
    @Test(expected = DXHTTPException.class)
    public void test() throws Exception {
        new Thread(() -> {
            try {
                System.err.println("start server thread");
                new ServerSocket(17291).accept();
                System.err.println("accepted");
            } catch (Exception e) {
                throw new RuntimeException(e);
            }
        }).start();

        DXEnvironment environment = DXEnvironment.Builder.fromDefaults()
                                                         .setApiserverHost("localhost")
                                                         .setApiserverPort(17291)
                                                         .build();

        DXSearch.findDataObjectsWithEnvironment(environment).withClassRecord().execute().asList();
    }
```
Expected result: The ``DXHTTPException`` exception should be thrown.
Actual result: Test hangs with stacktrace:
```
	at java.net.SocketInputStream.socketRead0(Native Method)
	at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
	at java.net.SocketInputStream.read(SocketInputStream.java:170)
	at java.net.SocketInputStream.read(SocketInputStream.java:141)
	at sun.security.ssl.InputRecord.readFully(InputRecord.java:465)
	at sun.security.ssl.InputRecord.read(InputRecord.java:503)
	at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:973)
	- locked <0x00000006cd309320> (a java.lang.Object)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1375)
	- locked <0x00000006cd309330> (a java.lang.Object)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1403)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1387)
	at org.apache.http.conn.ssl.SSLConnectionSocketFactory.createLayeredSocket(SSLConnectionSocketFactory.java:394)
	at org.apache.http.conn.ssl.SSLConnectionSocketFactory.connectSocket(SSLConnectionSocketFactory.java:353)
[...]

PR #297, opened 2018-02-08
Add socketTimeout and connectionTimeout parameters

pkokoshnikov:
These parameters are required to prevent stuck of threads on socket read operation

Issue: https://github.com/dnanexus/dx-toolkit/issues/286


commandlinegirl:
Thanks for this update! Let me run it through our test suite before we merge.

PR #301, opened 2018-02-16
SCI-744-clone-asset-ssl-error

Damien-Black:
**Motivation**

Multiprocessing with dxpy request encounters `SSLErrors`.

Using subprocess is a known workaround.

### DIFF
--- a/src/python/scripts/dx-clone-asset
+++ b/src/python/scripts/dx-clone-asset
@@ -4,6 +4,7 @@ from __future__ import print_function
 import argparse
 import subprocess
 import multiprocessing
+import traceback
 import time
 import functools
 import sys
@@ -36,7 +37,8 @@ def _parse_args():
                     help='Record-id of asset to clone.',
                     required=True)
     ap.add_argument('--regions',
-                    help='Regions to clone asset into.  Permitted regions are:\n[{0}]'.format(', '.join(SUPPORTED_REGIONS)),
+                    help='Regions to clone asset into.  Permitted regions are:\n[{supported_regions}]'.format(
+                        supported_regions=', '.join(SUPPORTED_REGIONS)),
                     choices=SUPPORTED_REGIONS,
                     default=SUPPORTED_REGIONS,
                     nargs='+',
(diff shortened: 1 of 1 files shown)


src/python/scripts/dx-clone-asset
```diff
@@ -89,18 +98,22 @@ def _clone_asset_into_region(region, record_name, asset_properties, asset_file_n
...
         try:
-            job.wait_on_done()
-            q.put(region)
-            return {region: job.describe()['output']['asset_bundle']}
-        except DXJobFailureError:
+            cmd = 'dx wait {0} '.format(job)
+            subprocess.check_output(cmd, shell=True)
+        except subprocess.CalledProcessError as cpe:
+            print(repr(cpe))
             pass
```
geetduggal: I believe if you're printing the `pass` statement is no longer necessary

src/python/scripts/dx-clone-asset
```diff
@@ -89,18 +98,22 @@ def _clone_asset_into_region(region, record_name, asset_properties, asset_file_n
...
+        cmd = 'dx describe {0} --json '.format(job)
+        job_desc = json.loads(subprocess.check_output(cmd, shell=True).strip())
```
geetduggal: Minor: it may be more convenient at times to use dxpy directly so there's less dependency on strings from stdout, e.g. https://github.com/dnanexus/dx-cwl/blob/master/dx-cwl#L556 (though this is code that eventually gets compiled to scatter app code so could be harder to read).  Maybe a better example is here: https://github.com/dnanexus/dx_app_builder/blob/orodeh-app-scatter/app_scatter/code.py

src/python/scripts/dx-clone-asset
```diff
@@ -56,19 +56,28 @@ def _parse_args():
...
+    # for the given region, or if no project is found and one
+    # can't be created, return None indicating there was a problem.
     try:
-        projects = dxpy.find_projects(name=project_name, region=region)
-        project = next(
-            projects, dxpy.api.project_new({'name': project_name, 'region': region}))
-        return project['id']
-    except Exception as e:
-        print(repr(e))
+        cmd = 'dx find projects --level CONTRIBUTE --name "{0}" --region "{1}" --brief '
```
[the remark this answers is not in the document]
Damien-Black: Totally agree, making this changes

src/python/scripts/dx-clone-asset
Damien-Black: Unfortunately when using the `multiprocessing` map_async functionality with the `dxpy` package directly to make request (no subprocess), the `requests` package throws `requests.packages.urllib3.exceptions.SSLError` error.

Damien-Black:
Made the changes waiting for @godotgildor review. Also noticed that finished items are put into the Queue manager only on job success. Made a minor change to catch job failures too. 

FYI Queue manager can stall the script if a job-id is removed (project being deleted) before the job reaches a terminal state. Only case I can think of this happening is if someone deletes the project the job is in, so not covering that case.

src/python/scripts/dx-clone-asset
```diff
@@ -88,23 +101,29 @@ def _clone_asset_into_region(region, record_name, asset_properties, asset_file_n
...
+            subprocess.check_output(cmd, shell=True)
+        except subprocess.CalledProcessError:
+            traceback.print_exc()
+
+        cmd = 'dx describe {job_id} --json '.format(job_id=job)
+        job_desc = json.loads(subprocess.check_output(cmd, shell=True).strip())
+
+        if job_desc['state'] == 'done':
+            record_id = job_desc['output']['asset_bundle']
+            return {region: job_desc['output']['asset_bundle']}
```
godotgildor: It doesn't look like you are placing the region into our queue if we take this codepath.  Perhaps replace the return statement with a simple break?  That way you have set the record_id, it will exit the while loop, the region will be placed in the queue on line 125 and line 126 will return the record_id set here.

godotgildor:
Thanks Damien!  Looks good.

PR #302, opened 2018-02-16
Fetch c-ares 1.12.0 from c-ares.haxx.se

billagee:
c-ares 1.13.0 has an issue preventing it from building on Windows (https://github.com/c-ares/c-ares/issues/119). Until the fix makes its way into a tagged release, we can keep using 1.12.0.


kpjensen:
Thanks Bill!

PR #310, opened 2018-03-09
dx download for symbolic links

orodeh:
Uplifting `dx download` to support symbolic links.  Uses `wget` for downloads, and `md5sum` for checksum validation.


geetduggal:
Closing as this is outmoded.

PR #311, opened 2018-03-13
Symlink download in python

orodeh:


geetduggal:
Thanks!  This LGTM.   I added some minor comments inline in the code.  The main blocker for this landing is a simple test in https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/test_dxclient.py.

PR #313, opened 2018-03-23
DEVEX-665 Add basic Python bindings for global workflow

commandlinegirl:


src/python/dxpy/bindings/dxglobalworkflow.py
```diff
@@ -0,0 +1,145 @@
...
+from ..compat import basestring
+
+####################
+# DXGlobalWorkflow #
+####################
+
+class DXGlobalWorkflow(DXObject, DXExecutable):
+    '''
+    Remote global workflow object handler.
+
```
geetduggal: Extra line

src/python/dxpy/bindings/dxglobalworkflow.py
```diff
@@ -0,0 +1,145 @@
...
+        elif name is not None:
+            self._name = name
+            if not isinstance(name, basestring):
+                raise DXError("Global workflow name needs to be a string: %r" % (name,))
+            if alias is not None:
+                if not isinstance(alias, basestring):
+                    raise DXError("Global workflow alias needs to be a string: %r" % (alias,))
+                self._alias = alias
+            else:
+                self._alias = 'default'
```
geetduggal: May want to add an `else` clause here to handle the no-input case if it's not handled well otherwise.

commandlinegirl: Thanks for the note. In this case we don't need an else since we won't be doing anything i.e. setting id, name, or alias if neither id or name values were provided in the arguments.

PR #321, opened 2018-04-03
DEVEX-676 Add initial support for dx build global workflow

commandlinegirl:


src/python/dxpy/workflow_builder.py
```diff
@@ -108,91 +110,275 @@ def _get_unsupported_keys(keys, supported_keys):
...
+    Precondition: json_spec must be validated
     """
+    workflow_id = dxpy.api.workflow_new(json_spec)["id"]
+    dxpy.api.workflow_close(workflow_id)
+    return workflow_id
+
+
+def _build_underlying_workflows(json_spec, args):
+    """
+    Creates a workflow in a temporary project for each enabled region.
```
geetduggal: Minor: I would put the TODO below as a part of this documentation.  Specifically, the documentation would say that it will build the workflow in the region of the current project but will be expanded in the future to build in all regions.

src/python/test/dx_batch.0000.tsv
```diff
@@ -0,0 +1,2 @@
+batch ID	pair1	pair2	pair1 ID	pair2 ID
+23	SRR123_1.gz	SRR223_2.gz	file-FF1kYF00bzpzP5BXK5b8QB0Z	file-FF1kYF00bzpzP5BXK5b8QB0f
```
geetduggal: Does this file need to be a part of this request or was it by accident?

src/python/test/dxpy_testutil.py
```diff
@@ -466,8 +468,8 @@ def write_workflow_directory(self, workflow_name, dxworkflow_str,
             readme_file.write(readme_content)
         return os.path.join(self.temp_file_path, workflow_name)
 
-    def create_workflow_spec(self, project_id):
-        applet_id = dxpy.api.applet_new({"name": "my_first_applet",
+    def create_applet(self, project_id):
+        return dxpy.api.applet_new({"name": "my_first_applet",
                                          "project": project_id,
```
geetduggal: Minor: indentation mismatch

PR #324, opened 2018-04-05
PTFM-21231: inline cluster bootstrapScript

robosoto:
Users can provide a filename under `clusterSpec.bootstrapScript`, where the listed file contains code to run during each cluster worker's setup. The contents of this file need to be read and inlined in the applet doc at build time (just like we do when a user provides their applet's source code filename under `runSpec.file`)

**Question:** Do I need to make changes anywhere else besides `dxpy`?

Test plan: 
`nucleus$ export DXTEST_ISOLATED_ENV=1`
`nucleus$ run_integration_tests.sh --client_tests_only --tests test_dxclient.TestDXBuildApp.test_build_cluster_app_bootstrap_script_inlined`

### DIFF
--- a/src/python/dxpy/app_builder.py
+++ b/src/python/dxpy/app_builder.py
@@ -541,11 +541,12 @@ def upload_applet(src_dir, uploaded_resources, check_name_collisions=True, overw
 
     # If this is a cluster applet, inline all provided bootstrapScript code
     # NOTE: assumes bootstrapScript is always provided as a filename
-    sys_reqs = applet_spec["runSpec"]["systemRequirements"]
-    for entry_point in sys_reqs:
-        if "clusterSpec" in sys_reqs[entry_point]:
-            with open(os.path.join(src_dir, sys_reqs[entry_point]["clusterSpec"]["bootstrapScript"])) as code_fh:
-                sys_reqs[entry_point]["clusterSpec"]["bootstrapScript"] = code_fh.read()
+    if "systemRequirements" in applet_spec["runSpec"]:
+        sys_reqs = applet_spec["runSpec"]["systemRequirements"]
+        for entry_point in sys_reqs:
+            if "clusterSpec" in sys_reqs[entry_point]:
+                with open(os.path.join(src_dir, sys_reqs[entry_point]["clusterSpec"]["bootstrapScript"])) as code_fh:
+                    sys_reqs[entry_point]["clusterSpec"]["bootstrapScript"] = code_fh.read()
 
     # Attach bundled resources to the app
     if uploaded_resources is not None:
--- a/src/python/dxpy/app_builder.py
+++ b/src/python/dxpy/app_builder.py
@@ -539,6 +539,14 @@ def upload_applet(src_dir, uploaded_resources, check_name_collisions=True, overw
             applet_spec["runSpec"]["code"] = code_fh.read()
             del applet_spec["runSpec"]["file"]
 
+    # If this is a cluster applet, inline all provided bootstrapScript code
+    # NOTE: assumes bootstrapScript is always provided as a filename
+    sys_reqs = applet_spec["runSpec"]["systemRequirements"]
+    for entry_point in sys_reqs:
+        if "clusterSpec" in sys_reqs[entry_point]:
+            with open(os.path.join(src_dir, sys_reqs[entry_point]["clusterSpec"]["bootstrapScript"])) as code_fh:
+                sys_reqs[entry_point]["clusterSpec"]["bootstrapScript"] = code_fh.read()
+
     # Attach bundled resources to the app
     if uploaded_resources is not None:
         applet_spec["runSpec"]["bundledDepends"].extend(uploaded_resources)
--- a/src/python/test/test_dxclient.py
+++ b/src/python/test/test_dxclient.py
@@ -6424,6 +6424,36 @@ class TestDXBuildApp(DXTestCaseBuildApps):
             container_content = dxpy.api.container_list_folder(app_container, {"folder": "/"})
             self.assertIn(file_id, [item["id"] for item in container_content["objects"]])
 
+    @unittest.skipUnless(testutil.TEST_ISOLATED_ENV,
+                         'skipping test that would create apps')
+    def test_build_cluster_app_bootstrap_script_inlined(self):
+        app_name = "cluster_app"
+        cluster_spec = {"type": "spark",
+                        "version": "2.2.0",
+                        "initialInstanceCount": 5,
+                        "bootstrapScript": "clusterBootstrap.py"}
+        app_spec = dict(self.base_app_spec, name=app_name,
+                        regionalOptions = {
+                            "aws:us-east-1": {
+                                "systemRequirements": {
+                                    "cluster_1": {
+                                        "instanceType": "mem2_hdd2_x1",
+                                        "clusterSpec": cluster_spec
+                                    },
+                                    "cluster_2": {
+                                        "instanceType": "mem2_hdd2_x4",
+                                        "clusterSpec": cluster_spec
+                                    }
+                                }
+                            }})
+        app_dir = self.write_app_directory(app_name, json.dumps(app_spec), "code.py")
+        self.write_app_directory(app_name, json.dumps(app_spec), "clusterBootstrap.py", code_content="testing")
+        app_id = json.loads(run("dx build --create-app --json " + app_dir))["id"]
+        app_desc_res = json.loads(run("dx describe --json " + app_id))
+        sys_reqs_res = app_desc_res["runSpec"]["systemRequirements"]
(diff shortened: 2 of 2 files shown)


robosoto:
@commandlinegirl just looking to get some eyes from your team on this as well. If you can take a look, thanks!

src/python/test/test_dxclient.py
```diff
@@ -6424,6 +6424,36 @@ def test_build_app_with_resources(self):
...
+                                    "cluster_1": {
+                                        "instanceType": "mem2_hdd2_x1",
+                                        "clusterSpec": cluster_spec
+                                    },
+                                    "cluster_2": {
+                                        "instanceType": "mem2_hdd2_x4",
+                                        "clusterSpec": cluster_spec
+                                    }
+                                }
+                            }})
```
commandlinegirl: I would rename this test to `test_build_and_get_cluster_app_bootstrap_script_inlined` and additionally check if subsequent getting of the app (with `dx get app-xxxx`) unpacks the app configuration into `dxapp.json` so that we can later build the app again with `dx build`, e.g.:

```
+                            "aws:us-east-1": {
+                                "systemRequirements": {
+                                    "cluster_1": {
+                                        "instanceType": "mem2_hdd2_x1",
+                                        "clusterSpec": {"type": "spark",
+                                                                    "version": "2.2.0",
+                                                                    "initialInstanceCount": 5,
+                                                                    "bootstrapScript": "clusterBootstrap.py"}
+                                                                    }
+                                    "cluster_2": {
+                                        "instanceType": "mem2_hdd2_x4",
+                                        "clusterSpec": <as above>
+                                    }
+                                }
+                            }}
```

The code for unpaking the app with `dx get` is here: https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/utils/executable_unbuilder.py#L238

Getting needs to result in dxapp.json whose specification is here: https://wiki.dnanexus.com/dxapp.json

Also is the file `clusterBootstrap.py` region-specific or the same for all regions?

src/python/dxpy/app_builder.py
```diff
@@ -539,6 +539,15 @@ def upload_applet(src_dir, uploaded_resources, check_name_collisions=True, overw
...
             del applet_spec["runSpec"]["file"]
 
+    # If this is a cluster applet, inline all provided bootstrapScript code
+    # NOTE: assumes bootstrapScript is always provided as a filename
+    if "systemRequirements" in applet_spec["runSpec"]:
+        sys_reqs = applet_spec["runSpec"]["systemRequirements"]
+        for entry_point in sys_reqs:
+            if "clusterSpec" in sys_reqs[entry_point]:
+                with open(os.path.join(src_dir, sys_reqs[entry_point]["clusterSpec"]["bootstrapScript"])) as code_fh:
+                    sys_reqs[entry_point]["clusterSpec"]["bootstrapScript"] = code_fh.read()
```
commandlinegirl: We should try/catch here if the file doesn't exist or cannot be accessed.

commandlinegirl: Also, we perform some syntax validation on the source code of the app so maybe we want to do it for this file as well?

robosoto:
Added syntax validation of `bootstrapScript` and handling of bootstrap scripts by `dx get`.

@commandlinegirl To answer your question: the `bootstrapScript` is not region-specific. 

However, any `clusterSpec` (which optionally includes the bootstrapScript filename) is associated with a specific entry point. Any subset of an app(let)'s entry points may request its own cluster and the different entry points may use the same or different `bootstrapScript`s

Note that this code assumes `clusterSpec.bootstrapScript` is provided as a filename (e.g. "src/myClusterBootstrap.sh").

src/python/test/test_dxclient.py
```diff
@@ -6424,6 +6424,104 @@ def test_build_app_with_resources(self):
...
+            run("dx build " + app_dir)
+        # get rid of syntax error
+        bootstrap_code_aws = "import sys\n"
+        self.write_app_directory(app_name, json.dumps(app_spec), "clusterBootstrapAws.py", code_content=bootstrap_code_aws)
+
+        def build_and_verify_bootstrap_script_inlined(app_dir):
+            # build cluster app with multiple bootstrap scripts and regions
+            # expect bootstrap scripts to be inlined in the app doc
+            app_id = json.loads(run("dx build --create-app --json " + app_dir))["id"]
+            app_desc_res = json.loads(run("dx describe --json " + app_id))
```
commandlinegirl: `dx build` with the --json flag returns a describe output of the built app so we can spare this additional `describe` API call :-)

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -101,6 +101,13 @@ def _dump_app_or_applet(executable, omit_resources=False, describe_output=[]):
     with open(script, "w") as f:
         f.write(info["runSpec"]["code"])
 
+    # Used to unpack cluster bootstrapScript(s) if they exist for this app(let)
+    def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
```
commandlinegirl: Our convention is to use triple double quotes for methods comments/descriptions. Thanks!

src/python/dxpy/app_builder.py
```diff
@@ -539,6 +539,22 @@ def upload_applet(src_dir, uploaded_resources, check_name_collisions=True, overw
...
             del applet_spec["runSpec"]["file"]
 
+    # If this is applet requires a cluster, inline any bootstrapScript code that may be provided.
+    # bootstrapScript is an *optional* clusterSpec parameter.
+    # NOTE: assumes bootstrapScript is always provided as a filename
+    if "systemRequirements" in applet_spec["runSpec"]:
+        sys_reqs = applet_spec["runSpec"]["systemRequirements"]
+        for entry_point in sys_reqs:
+            try:
+                bootstrapScript = os.path.join(src_dir, sys_reqs[entry_point]["clusterSpec"]["bootstrapScript"])
```
commandlinegirl: We can require the script names to follow the same format we use for `dx get`, with region prefix, etc. This way, it will be guaranteed the files generated by dx get have always the same names as the original files used for `build`.

Also, if the bootstrap script contents is not region-specific, why do we need a different script for each region? Wouldn't one file be enough? Maybe I'm missing something.

src/python/test/test_dxclient.py
```diff
@@ -6424,6 +6424,104 @@ def test_build_app_with_resources(self):
             container_content = dxpy.api.container_list_folder(app_container, {"folder": "/"})
             self.assertIn(file_id, [item["id"] for item in container_content["objects"]])
 
+    @unittest.skipUnless(testutil.TEST_ISOLATED_ENV,
```
commandlinegirl: That's a great, comprehensive test, thanks!

robosoto:
@commandlinegirl Thanks for your in-person question about the bootstrapScript and regions. I take back what I said earlier. While `bootstrapScript` could be identical across regions,  it could certainly be used to perform region-specific setup. 

One simple example could be retrieving objects from different S3 buckets / Azure blobs based on the region.

src/python/dxpy/app_builder.py
commandlinegirl: Hmm, a thought after our discussion: requiring a specific name format during `dx build` doesn't make sense though since users will probably want to use one script stored locally for all or most regions.

robosoto: Yes, I agree, users should be able to write one script and reference it from multiple regions and/or entry points in the spec. `dx get` will write the script back into a file(s) that follows a standard naming scheme, but imposing a naming scheme when the user is writing the app spec isn't necessary.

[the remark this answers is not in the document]
robosoto: Too much JS 🐫

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -402,6 +416,33 @@ def _verify_app_source_dir_impl(src_dir, temp_dir, mode, enforce=True):
         logging.warn('%s contained syntax errors, see above for details' % (files_str,))
 
 
+def _get_all_cluster_bootstrap_script_names(manifest):
+    script_names = []
+
+    if 'systemRequirements' in manifest['runSpec']:
```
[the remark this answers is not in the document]
robosoto: If we throw that error here with no other changes, it will appear for _any_ attempt to build an app that specifies `runSpec.systemRequirements`. Do we want that?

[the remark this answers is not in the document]
commandlinegirl: I meant we could only throw an error when a bootstrapping config is defined under `runSpec.systemRequirements` and only allow it in `regionalOptions`. When we `dx get` an app/applet we also don't inline this into dxapp.json any more - the `systemRequirements` is only under `regionalOptions`.

PR #348, opened 2018-05-03
Better checking for inputs and/or inputSpec

orodeh:
It is possible for inputs to be None, which passes the if/else, causing the __get_types_for_inputs_ method to fail.


commandlinegirl:
One note: what happens when both inputSpec and inputs are empty lists?

Also, can add to the commit header the ticker ID: DEVEX-727? Thanks!

PR #352, opened 2018-05-31
Wait file list

godotgildor:
Adding ability to provide a file which contains a list of jobs or data objects upon which to wait.  

We ran into instances where we wanted to wait on a number of objects and hit bash limits on the number of characters that could be provided at the command line.  This new feature will allow us to get around that by simply specifying the longer list of jobs or data objects in a file instead of as command line arguments.

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -2409,6 +2409,12 @@ def close(args):
 
 def wait(args):
     had_error = False
+    # If only one path provided, check to see if it is a local file
+    # and if so gather actual paths on which to wait from the contents 
+    # of the file.
+    if len(args.path) == 1 and os.path.isfile(args.path[0]):
+        args.path = open(args.path[0]).read().strip().split('\n')
+
     for path in args.path:
         if is_job_id(path) or is_analysis_id(path):
             dxexecution = dxpy.get_handler(path)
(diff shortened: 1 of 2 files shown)


src/python/dxpy/scripts/dx.py
```diff
@@ -2409,6 +2409,12 @@ def close(args):
 
 def wait(args):
     had_error = False
+    # If only one path provided, check to see if it is a local file
+    # and if so gather actual paths on which to wait from the contents 
+    # of the file.
+    if len(args.path) == 1 and os.path.isfile(args.path[0]):
+        args.path = open(args.path[0]).read().strip().split('\n')
```
commandlinegirl: Please, try/catch here in case the file cannot be read due to insufficient permissions, etc. You can catch IOError like in this case: https://github.com/dnanexus/dx-toolkit/blob/c9135627a772a7a51c09624f433a39d6188250d1/src/python/dxpy/scripts/dx_build_app.py#L330-L334 and raise DXCLIError.

src/python/dxpy/scripts/dx.py
```diff
@@ -4965,7 +4971,7 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
 # wait
 #####################################
 parser_wait = subparsers.add_parser('wait', help='Wait for data object(s) to close or job(s) to finish',
-                                    description='Polls the state of specified data object(s) or job(s) until they are all in the desired state.  Waits until the "closed" state for a data object, and for any terminal state for a job ("terminated", "failed", or "done").  Exits with a non-zero code if a job reaches a terminal state that is not "done".',
+                                    description='Polls the state of specified data object(s) or job(s) until they are all in the desired state.  Waits until the "closed" state for a data object, and for any terminal state for a job ("terminated", "failed", or "done").  Exits with a non-zero code if a job reaches a terminal state that is not "done".  Can also provide a local file containing a list of data object(s) or job(s), one per line.',
                                     prog='dx wait',
                                     parents=[env_args])
 path_action = parser_wait.add_argument('path', help='Path to a data object or job ID to wait for', nargs='+')
```
commandlinegirl: Please, add "or a local file.." here since most people will just read (use as a reference) just the arguments. Thanks!

commandlinegirl:
I made the updates so that we can merge, cheers!:)

PR #355, opened 2018-06-06
PTFM-21023 dx watch job log relay test

ocbrewing:
This commit adds a test for dx watch to ensure websocket errors are handled
correctly from the apiserver route.
A recent change to the apiserver changes the error code
for closing the websocket from 1008 to 4008.  Despite that socket closing
code being different, the client error should continue to look the same.

### DIFF
--- a/src/python/test/test_dxclient.py
+++ b/src/python/test/test_dxclient.py
@@ -359,6 +359,28 @@ class TestDXClient(DXTestCase):
         shell.sendline("echo find projects | dx sh")
         shell.expect("project-")
 
+    def test_dx_watch_invalid_auth(self):
+        with without_auth():
+          with self.assertSubprocessFailure(stderr_regexp="PermissionDenied", exit_code=3):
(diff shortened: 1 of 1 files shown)


src/python/test/test_dxclient.py
```diff
@@ -359,6 +359,28 @@ def test_dx_shell(self):
         shell.sendline("echo find projects | dx sh")
         shell.expect("project-")
 
+    def test_dx_watch_invalid_auth(self):
```
ocbrewing: one thing I'm not sure of.. do I need to put some sort of skip on this test? I'm attempting to make sure that the environment is the same after this test as it was before, so I don't see why we should skip it - but maybe I'm missing something?

commandlinegirl: It looks that nothing needs to be skipped here. Usually we skip job runs or tests that create some global objects, orgs, etc. and only run them in special test suite environments.

PR #358, opened 2018-06-11
Add a script that attempts to autogenerate valid dxapp.json for a python app that uses argparse

jdidion:
The motivation for developing this script is that I have a python tool I want to appify, but it has dozens of command line parameters, and I did not want to have to type all of them into the app wizard.
The goal of this script is to generate a valid dxapp.py given a python application that uses argparse. It is likely that the user will still want to go in and tweak the results (e.g. provide better argument labels; correct data types that could not be guessed correctly), but it automates at least 80% of the effort.

This script works by subclassing ArgumentParser and then installing that subclass in the argparse module (i.e. monkey patching). The subclass overrides the parse_args method with one that loops through the configured actions, converts them to dxapp specs, adds all of the specs and other metadata to a dict, converts that dict to JSON, and then either writes it to a file or dumps it to stdout. This script could eventually be extended to support other command line parsing frameworks (e.g. Click, optparse).

This script works with target applications written in both python2 and python3 (which is why the shebang line is 'python' and not 'python2.7').


geetduggal:
Thanks!

PR #364, opened 2018-06-20
TIP-415: extend DXHTTPRequest; allow overriding of httpclient to send client certificate in request.

codecheetos:
required for TIP private routes for 'database-xxxx' operations. Changes in tip Apiserver client implementation overrides httpclient implementation with sslcontext. Keystore path, keystore password etc. passed as system property from hivemetastore and thriftserver.

@arunbhat @durae @geetduggal @wkoetke


arunbhat:
LGTM. @geetduggal this was needed for https://github.com/dnanexus/nucleus/pull/1796

PR #367, opened 2018-06-22
DEVEX-750 dx wdl integration

orodeh:
Integrating dxWDL into the dx-toolkit.

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -2613,6 +2613,47 @@ def build(args):
         print("Error: %s" % (e.message,), file=sys.stderr)
         err_exit()
 
+def build_pwf(args):
+    sys.argv = ['dx build_pwf'] + sys.argv[2:]
+    #args = parser_build_pwf.parse_args()
+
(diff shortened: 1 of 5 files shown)


src/python/dxpy/scripts/dx.py
```diff
@@ -2613,6 +2613,129 @@ def handle_arg_conflicts(args):
...
+            folder = d
+    if folder == "":
+        folder = '/'
+    if not folder.startswith('/'):
+        raise DXParserError('Error: folder must start with slash {}'.format(d))
+    return project + ":" + folder
+
+
+# Convert each unicode character to four hexadecimal digits
+def _unicodeToHex(buf):
```
geetduggal: Would be nice to know if there is a simpler way to handle unicode passed through to subprocess and then eventually execv and then in this case even into the JVM.   Seems like a lot of issues could happen along the way, so the hex encoding makes sense for now since you have control over both encoding and decoding across languages.  May want to add a comment to explain the reasoning here.

orodeh:
@geetduggal I think I fixed the issues. Please let me know if I can merge the branch. 

Thanks, 
      Ohad.

PR #368, opened 2018-06-26
Fix bash helpers tests

billagee:


orodeh:
This test is failing because the dxWDL runtime library is not installed on staging. @geetduggal, can you please reassign this task to me?

orodeh:
Meanwhile, I updated the dxWDL to version 0.68. This should clear this problem. dxWDL is now installed on staging us-east, as well as production.

PR #372, opened 2018-07-08
Make dx tool (& dxpy imports) usable with python 3

jtratner:
small fixes that allow `dx -h` to work in Python 3 and also a couple uses of the dxpy.api module as well.

Adding test cases or generation for Python 3 might be a little beyond me - but I figure this is a pretty straightforward change :)


jtratner:
Thanks for the quick review! :)

jtratner:
@commandlinegirl - btw, any interest in/thoughts about packaging a (experimental!) python 3 version of dx-toolkit? I haven't been able to compile the toolkit myself, but we're trying to move to be a Python 3-only shop by the end of this year.

jtratner:
is there a standard release schedule for dxpy? just wondering when this fix will go out.

PR #377, opened 2018-07-17
Remove unavailable documentation links

durae:
Private documentation can be found [here](https://github.com/dnanexus/nucleus/blob/master/apiserver/tip/Readme.md)


durae:
Thanks! I've checked it for Python. In all other SDKs it was just removal of commented lines.

PR #385, opened 2018-07-25
added logic to use aria for downloads, if preset, rather than wget as…

ifelsefi:
… it's much faster


orodeh:
@JusticeForMikeBrown thanks! that is a very nice addition. Indeed aria2c is faster than wget, but it is not always installed. We will incorporate this into dx-toolkit. We did want to clean it up a bit, and run it though our tests first.

PR #397, opened 2018-08-20
DEVEX-834 Upgrade Win installer to Python 2.7.15

billagee:
Bundling Python 2.7.15 for Windows upgrades setuptools, which allows us to `pip install cryptography<=2.2.2` on Windows.


geetduggal:
Thanks @billagee !

PR #401, opened 2018-09-07
SCI-1087 add stages to asset depends

msimbirsky:
In bundledDepends, you can specify to only use the bundle in certain stages of an app by supplying the "stages" argument. However, this argument is not checked for when converting from an asset to a bundle, so assetDepends does not support this. 

I've updated the asset handling to support this and added a test.


msimbirsky:
Tests run 👍

PR #402, opened 2018-09-07
DEVEX-851 - Fix python 3 dx upload issue

geetduggal:
Looks like the `headers` and `content-length` fields were duplicated when testing for python 2 vs. 3.  These lines simply check for python 3 and delete those fields when passing into urllib3.  Thanks to @jtratner for alerting us to this


commandlinegirl:
Thanks, Geet! I'll review it and also test it in my environment.

PR #403, opened 2018-09-10
Fix import dxpy cli

pkaleta:
I realize that Python 3 is not yet fully supported in dxpy, but there was a problem that seems to be fixed by this small PR.

Namely, the problem was the following error when using Python 3, dxpy and AWS Step Functions:

```
File "/var/task/basespace_sm/sequencing_handlers.py", line 7, in <module>
from dxpy import DXApplet
File "/var/task/dxpy/__init__.py", line 1002, in <module>
from .bindings import *
File "/var/task/dxpy/bindings/__init__.py", line 649, in <module>
from .dxfile import DXFile, DXFILE_HTTP_THREADS, DEFAULT_BUFFER_SIZE
File "/var/task/dxpy/bindings/dxfile.py", line 36, in <module>
from ..utils.resolver import object_exists_in_project
File "/var/task/dxpy/utils/resolver.py", line 34, in <module>
from ..cli import try_call, INTERACTIVE_CLI
File "/var/task/dxpy/cli/__init__.py", line 26, in <module>
INTERACTIVE_CLI = True if sys.stdin.isatty() and sys.stdout.isatty() else False
AttributeError: 'NoneType' object has no attribute 'isatty'
```

The problem is that according to the [docs](https://docs.python.org/3/library/sys.html#sys.__stdin__):

>nder some conditions stdin, stdout and stderr as well as the original values __stdin__, __stdout__ and __stderr__ can be None. It is usually the case for Windows GUI apps that aren’t connected to a console and Python apps started with pythonw.

Which I think is the case in AWS Step Functions too.

In this PR I first check that `sys.stdin` and `sys.stdout` are not None, before trying to use them. I've also added a unit test that fails without the fix from 6fdf569.

### DIFF
--- a/src/python/test/test_dxclient.py
+++ b/src/python/test/test_dxclient.py
@@ -76,6 +76,14 @@ def list_folder(proj_id, path):
     return output
 
 
+def test_dxpy_import_stdin_none(monkeypatch):
+    """
+    Ensure that importing dxpy works even if stdin is None.
+    """
+    monkeypatch.setattr(sys, 'stdin', None)
+    del sys.modules['dxpy.cli']
+    from dxpy.cli import INTERACTIVE_CLI
+
 
 class TestDXTestUtils(DXTestCase):
     @pytest.mark.TRACEABILITY_MATRIX
--- a/src/python/test/test_dxclient.py
+++ b/src/python/test/test_dxclient.py
@@ -76,13 +76,14 @@ def list_folder(proj_id, path):
     return output
 
 
-def test_dxpy_import_stdin_none(monkeypatch):
-    """
-    Ensure that importing dxpy works even if stdin is None.
-    """
-    monkeypatch.setattr(sys, 'stdin', None)
-    del sys.modules['dxpy.cli']
-    from dxpy.cli import INTERACTIVE_CLI
+class TestDXPYImport(object):
+    def test_dxpy_import_stdin_none(self, monkeypatch):
+        """
+        Ensure that importing dxpy works even if stdin is None.
+        """
+        monkeypatch.setattr(sys, 'stdin', None)
+        del sys.modules['dxpy.cli']
+        from dxpy.cli import INTERACTIVE_CLI
 
 
 class TestDXTestUtils(DXTestCase):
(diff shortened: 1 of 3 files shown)


src/python/test/test_dxclient.py
```diff
@@ -76,6 +76,14 @@ def list_folder(proj_id, path):
     return output
 
 
+def test_dxpy_import_stdin_none(monkeypatch):
```
commandlinegirl: Thank you for the contribution!

How do you run this test? Also, we put all the tests under Test classes, so you can for example put your unit tests under `TestDXTestUtils`.

pkaleta: I used this to run it:

```
% PYTHONPATH=src/python/test pytest src/python/test/test_dxclient.py::TestDXPYImport
================================================ test session starts =================================================
platform darwin -- Python 2.7.15, pytest-3.5.1, py-1.6.0, pluggy-0.6.0
rootdir: /Users/piotr/counsyl/dx-toolkit/src/python, inifile:
collected 1 item

src/python/test/test_dxclient.py .                                                                             [100%]

============================================== 1 passed in 0.57 seconds ==============================================
```

>Also, we put all the tests under Test classes, so you can for example put your unit tests under TestDXTestUtils.

Done. Moved it to `TestDXPYImport`, because it doesn't need `DXTestCase` or anything, just a simple Python object is enough. Also it doesn't seem like it is similar to other methods in `TestDXTestUtils`. Let me know what you think.

geetduggal: Thanks @pkaleta!  Did you commit the move and push to the branch?  (I don't see it at first glance :))

pkaleta: Good point @geetduggal :) Check out now.

commandlinegirl:
Thanks, Piotr! We use unittest, not pytest for testing so using monkeypatch might not work like that, let me check.

pkaleta:
The changes I'm proposing are in 1e3d5c5.

commandlinegirl:
Yes, thanks for rewriting the test with mock. We only use some pytest decorators. This LGTM.

PR #407, opened 2018-09-13
DEVEX-853 cat argument too long

orodeh:
A fix that created bash variables for each of a job's inputs, only if it isn't too long. The current limit is 16KB.

### DIFF
--- a/src/python/scripts/dx-print-bash-vars
+++ b/src/python/scripts/dx-print-bash-vars
@@ -39,5 +39,5 @@ for key, val in var_defs_hash.iteritems():
     if val_len <= (16 * 1024):
         print("export {}={}".format(key, val))
     else:
-        print("the variable {} is too long ({} bytes), omitting".format(key, val_len),
-              file=sys.stderr)
+        msg = "the variable {} is too long ({} bytes), omitting".format(key, val_len)
+        print("echo {}".format(msg))
--- a/src/python/scripts/dx-print-bash-vars
+++ b/src/python/scripts/dx-print-bash-vars
@@ -39,5 +39,5 @@ for key, val in var_defs_hash.iteritems():
     if val_len <= (16 * 1024):
         print("export {}={}".format(key, val))
     else:
-        msg = "the variable {} is too long ({} bytes), omitting".format(key, val_len)
+        msg = "\"the variable {} is too long ({} bytes), omitting\"".format(key, val_len)
         print("echo {}".format(msg))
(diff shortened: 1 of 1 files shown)


src/python/scripts/dx-print-bash-vars
```diff
@@ -35,4 +35,9 @@ job_input_file = file_load_utils.get_input_json_file()
 
 var_defs_hash = file_load_utils.gen_bash_vars(job_input_file)
 for key, val in var_defs_hash.iteritems():
-    print("export {}={}".format(key, val))
+    val_len = len(val)
+    if val_len <= (16 * 1024):
+        print("export {}={}".format(key, val))
+    else:
+        print("the variable {} is too long ({} bytes), omitting".format(key, val_len),
```
commandlinegirl: Hm, is this output later piped to `source`? In that case it'd better to add `echo` to the second print.

orodeh: That's a good point.

src/python/scripts/dx-print-bash-vars
```diff
@@ -35,4 +35,9 @@ job_input_file = file_load_utils.get_input_json_file()
 
 var_defs_hash = file_load_utils.gen_bash_vars(job_input_file)
 for key, val in var_defs_hash.iteritems():
-    print("export {}={}".format(key, val))
+    val_len = len(val)
+    if val_len <= (16 * 1024):
+        print("export {}={}".format(key, val))
+    else:
+        msg = "the variable {} is too long ({} bytes), omitting".format(key, val_len)
+        print("echo {}".format(msg))
```
commandlinegirl: I'd also surround it in brackets otherwise it'll error out on (), this should work 
`msg = "\"the variable {} is too long ({} bytes), omitting\"".format(key, val_len)`.

commandlinegirl: Yes, a quick test confirmation:

```
$ cat x.py 
key="a"
val_len=10
msg = "\"the variable {} is too long ({} bytes), omitting\"".format(key, val_len)
print("echo {}".format(msg))
$ python x.py > x.sh
$ source x.sh 
the variable a is too long (10 bytes), omitting
```

commandlinegirl: Or, it'd be even better to put the quotes in the echo print

orodeh:
A fix that creates a bash variable for a job inputs, only if it isn't too long. The current limit is 16KB.

PR #411, opened 2018-10-03
Add a convenience login script for a cloud workstation

geetduggal:
Pretty simple, but saves some time.  @jdidion @singerma @commandlinegirl interest


commandlinegirl:
That's useful! Thanks, @geetduggal!

ptn24:
How does this work? Just curious because I use cloud workstation at times

Damien-Black:
This is useful:

1. `unset DX_WORKSPACE_ID` allows you to change the project (`dx select` any project) without `dx-toolkit` reverting back to the `DX_WORKSPACE_ID`.
    * Before if you relied on relative paths when using `dx` commands it would alway change you current project back to `DX_WORKSPACE_ID`
2. `dx cd $DX_PROJECT_CONTEXT_ID:` changes project context to the root of the project that launched the cloud workstation job
3. `sudo PYTHONPATH=$PYTHONPATH -E bash` Puts you in a subshell with root access and configures `PYTHONPATH` so that `dxpy` still works as expected. This mimics how our worker run job scripts.


Basically if you run this script, you'll be able to interactively work from a cloud workstation instance:
* using relative paths. And not being surprised when you upload a file to the jobs container instead of the parent project.
* Without specifying sudo for almost every command

kpjensen:
@geetduggal since this is a bash script, not sure if we want to house it under `src/python`. What about under `bin/`?

src/python/scripts/dx-su-contrib
```diff
@@ -0,0 +1,3 @@
+unset DX_WORKSPACE_ID
```
kpjensen: `#!/bin/bash` shebang line?

geetduggal: Yes, thanks!

geetduggal:
> @geetduggal since this is a bash script, not sure if we want to house it under `src/python`. What about under `bin/`?

Done

jdidion:
This is awesome @geetduggal!

PR #416, opened 2018-10-11
Replace ws4py with websocket-client library in job_log_client

jdidion:
We received a request from Lilly to support proxies. The only place where this is an issue is job_log_client, which uses websockets to poll the log of a running job. Currently, we use ws4py, which does not have proxy support. This PR replaces ws4py with websocket-client, which does have proxy support.

Please code review. I would hold off on merging for now. I have provided Lilly with a build of this branch and they will test it for a while.


jdidion:
Thanks Geet. Michael Neylon from Lilly used this build today and verified it now works inside the Lilly environment. I am going to merge it in. But I will also add an issue to add tests for job_log_client, especially the reconnect logic.

PR #425, opened 2018-10-27
Adding "publish" option to dx-clone-asset

godotgildor:
This tool is mostly used for DNAnexus developers as they develop apps.  Once they clone the asset to new regions and they believe the asset is ready to be included with their app(let), they may want to copy the asset into the global asset projects.  This PR adds an undocumented option to copy the assets into these global asset projects.  For folks outside of DNAnexus, they won't be presented with this option when running `dx-clone-asset -h` since the option help is suppressed.  Moreover, if they do attempt to use it, the script will check to see if they have write access to the global asset projects.  Presumably they won't.  In this case the publish will become a no opt, simply stating that no global asset projects could be found.


commandlinegirl:
@godotgildor Is it something we still want to merge?

PR #430, opened 2018-10-31
Fix shell prompt in test_dx_ssh

billagee:


src/python/test/test_dxclient.py
```diff
@@ -903,7 +897,10 @@ def _test_dx_ssh(self, project, instance_type):
                 dx.expect("The job is running in terminal 1.", timeout=5)
                 # Check for terminal prompt and verify we're in the container
                 job_id = dxpy.find_jobs(name="sleep", project=project).next()['id']
-                dx.expect(("dnanexus@%s" % job_id), timeout=10)
+                # Expect the shell prompt - for example: dnanexus@job-xxxx:~⟫
+                dx.expect((job_id), timeout=10)
+                # NOTE: \u27EB is ⟫
+                dx.expect(u'\u27EB'.encode(sys_encoding), timeout=10, searchwindowsize=7000)
```
orodeh: This looks very fragile. 
Is there a way to check that the string `"dnanexus@%s" % job_id` is a substring of the line? 

I am now working through merging back the python-3 branch into master, and this kind of thing is going to get even trickier.

billagee: Good point about python 3; I'll take a look at simplifying this to avoid adding new string encoding gymnastics.

PR #434, opened 2018-11-02
Make DXProject.new be a classmethod.

pkaleta:
I think it makes more sense for the `DXProject.new` be a class method rather than instance method. Doesn't seem like you should need an instance of a project to create a new one.


jtratner:
thoughts on this PR @geetduggal ?

PR #441, opened 2018-11-09
Removing use of ubuntu 12.04 in test_dxclient.py

orodeh:


orodeh:
Turns out the `dx_ssh_test` in `test_dxclient.py` used Ubuntu 12.04. The test was actually relying on the particular output that bash has, on that platform. When you upgrade to 14.04, bash provides color coded output, and that broke the test.

PR #449, opened 2018-11-30
Fix Python virtualenv path in traceability runner

billagee:


billagee:
Thanks @orodeh! Yeah, the environment where this script is run still uses 2.7.

commandlinegirl:
@billagee Would it be possible to switch to py3?

billagee:
> @billagee Would it be possible to switch to py3?

Sure - I don't think the test suite that uses this shell script needs to stay on 2.7. I'll follow up with more info soon :)

PR #450, opened 2018-12-04
DEVEX-946 Fix writing to stdout for py2

commandlinegirl:


commandlinegirl:
Thanks! Yes, dxfile.read() will return bytes (according to the docstrings to [_read2](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/bindings/dxfile.py#L859)).

geetduggal:
Thanks @commandlinegirl !

PR #459, opened 2019-01-15
DEVEX-965 Fix file upload issue using api proxy in Python 3 environment

commandlinegirl:


orodeh:
Thanks for debugging this! 

If possible, two things would make it nicer:
1) Decoupling the python code from reliance on apiproxy
2) A test that would upload a file from a worker, reproducing the problem.

commandlinegirl:
Thanks for the review! I updated the code as per suggestions. I also added a test on the nucleus (it fails when using master but passes when using this branch). The change passes py2 and py3 branch suite tests.

commandlinegirl:
I like this solution more, too! Thanks

PR #467, opened 2019-01-29
TIP-1023: retry 503 just like 500

ghost:
For TIP, we want to treat 503 (ServiceUnavailable) errors just like 500 (InternalError), in terms of the retry with exponential backoff. Previously, 503 had an infinite retry every 60 seconds.

### DIFF
--- a/src/java/src/main/java/com/dnanexus/DXHTTPRequest.java
+++ b/src/java/src/main/java/com/dnanexus/DXHTTPRequest.java
@@ -279,7 +279,6 @@ public class DXHTTPRequest {
             // may update this to unconditionally retry if we can definitely determine that the
             // server never saw the request.
             boolean retryRequest = (retryStrategy == RetryStrategy.SAFE_TO_RETRY);
-            int retryAfterSeconds = 60;
 
             try {
                 // In this block, any IOException will cause the request to be retried (up to a
@@ -362,48 +361,21 @@ public class DXHTTPRequest {
                     throw DXAPIException.getInstance(errorType, errorMessage, statusCode);
                 } else {
                     // Propagate 500 error to caller
-                    if (this.disableRetry && statusCode != 503) {
+                    if (this.disableRetry) {
                         logError("POST " + resource + ": " + statusCode + " Internal Server Error, try "
                                 + String.valueOf(attempts + 1) + "/" + NUM_RETRIES
                                 + " Request ID: " +  requestId);
                         throw new InternalErrorException("Internal Server Error", statusCode);
                     }
-                    // If retries enabled, 500 InternalError should get retried unconditionally
+                    // If retries enabled, 500 InternalError should get retried unconditionally,
+                    // and so should 503 ServiceUnavailable (TIP-1023).
                     retryRequest = true;
-                    if (statusCode == 503) {
-                        Header retryAfterHeader = response.getFirstHeader("retry-after");
-                        // Consume the response to avoid leaking resources
-                        EntityUtils.consume(entity);
-                        if (retryAfterHeader != null) {
-                            try {
-                                retryAfterSeconds = Integer.parseInt(retryAfterHeader.getValue());
-                            } catch (NumberFormatException e) {
-                                // Just fall back to the default
-                            }
-                        }
-                        throw new ServiceUnavailableException("503 Service Unavailable", statusCode, retryAfterSeconds);
-                    }
                     throw new IOException(EntityUtils.toString(entity));
                 }
-            } catch (ServiceUnavailableException e) {
-                int secondsToWait = retryAfterSeconds;
-
-                if (this.disableRetry) {
-                    logError("POST " + resource + ": 503 Service Unavailable, suggested wait "
-                            + secondsToWait + " seconds" + ". Request ID: " +  requestId);
-                    throw e;
-                }
-
-                // Retries due to 503 Service Unavailable and Retry-After do NOT count against the
-                // allowed number of retries.
-                logError("POST " + resource + ": 503 Service Unavailable, waiting for "
-                        + Integer.toString(secondsToWait) + " seconds" + " Request ID: " +  requestId);
-                sleep(secondsToWait);
-                continue;
             } catch (IOException e) {
                 // Note, this catches both exceptions directly thrown from httpclient.execute (e.g.
                 // no connectivity to server) and exceptions thrown by our code above after parsing
-                // the response.
+                // the response. Also 503 ServiceUnavailable (TIP-1023).
                 logError(errorMessage("POST", resource, e.toString(), timeoutSeconds,
                         attempts + 1, NUM_RETRIES));
                 if (attempts == NUM_RETRIES || !retryRequest) {
@@ -414,9 +386,6 @@ public class DXHTTPRequest {
                             statusCode);
                 }
             }
-
-            assert attempts < NUM_RETRIES;
-            assert retryRequest;
 
             attempts++;
 


src/java/src/main/java/com/dnanexus/DXHTTPRequest.java
```diff
@@ -362,48 +361,21 @@ private ParsedResponse requestImpl(String resource, String data, boolean parseRe
                     throw DXAPIException.getInstance(errorType, errorMessage, statusCode);
                 } else {
                     // Propagate 500 error to caller
```
arunbhat: May be the comment also needs to be updated?

src/java/src/main/java/com/dnanexus/DXHTTPRequest.java
```diff
@@ -362,48 +361,21 @@ private ParsedResponse requestImpl(String resource, String data, boolean parseRe
...
-                    if (this.disableRetry && statusCode != 503) {
```
arunbhat: @geetduggal  or @commandlinegirl can you please review? Not sure why 503 was excluded..

arunbhat:
Not really sure if this change is needed since there already is a retry mechanism for 503 (with the retryAfterSeconds header). Are we sure we are not retrying? Can you please simulate a 503 (like return 503 on requests that are at a time that is divisible by 2?) in the api server and see what happens?

arunbhat:
May be the only problem to fix is in TIP-1081?

wkoetke:
@arunbhat , to answer your above questions:
1) "Are we sure we are not retrying?" (my answer: we ARE retrying already, just with different logic, an infinite number of retries every 60 seconds instead of exponential backoff for 6 attempts.
2) Can you please simulate a 503 (like return 503 on requests that are at a time that is divisible by 2?) in the api server and see what happens? (my answer: yes, already did that, that was my testing approach all along, both before and after my change. I simulated the infinite retries every 60 seconds before, and the exponential retry after, my change)

wkoetke:
> Please see my comments earlier. Not sure if this change is needed.

I think we should proceed with it. Exponential backoff is better than forcing it to retry and wait a full 60 seconds. This Java codepath is only used by TIP at this point as far as I know (not the cmd line dx toolkit invocations) so we can have our way with it to some extent.

wkoetke:
> May be the only problem to fix is in TIP-1081?

Ok, but let's not have the same discussion after implementation of 1080 where we say "maybe we shouldn't do this...500 was being excluded on purpose in the code". Either we want the change, or not.

src/java/src/main/java/com/dnanexus/DXHTTPRequest.java
wkoetke: The existing 503 logic seems fine if you imagine the typical cmd line dx-toolkit operation. But for TIP, it makes less sense, and it's better to quit after the small # of retries and let spark handle it. Just my opinion.

arunbhat:
> > Please see my comments earlier. Not sure if this change is needed.
> 
> I think we should proceed with it. Exponential backoff is better than forcing it to retry and wait a full 60 seconds. This Java codepath is only used by TIP at this point as far as I know (not the cmd line dx toolkit invocations) so we can have our way with it to some extent.

yep. @geetduggal  -> what do you think? Also @wkoetke -> could you please check what the python sdk does? Is it the same or different?

src/java/src/main/java/com/dnanexus/DXHTTPRequest.java
commandlinegirl: It was added at a client's request  https://jira.internal.dnanexus.com/browse/PTFM-15849. It was actually implemented by adding the ability to disable retries for all 5xx errors (https://github.com/dnanexus/dx-toolkit/commit/968fdd67433dcea7c6b7eb14ce2536991829b9ab) with some special handling of 503.

src/java/src/main/java/com/dnanexus/DXHTTPRequest.java
```diff
@@ -362,48 +361,21 @@ private ParsedResponse requestImpl(String resource, String data, boolean parseRe
...
                         logError("POST " + resource + ": " + statusCode + " Internal Server Error, try "
                                 + String.valueOf(attempts + 1) + "/" + NUM_RETRIES
                                 + " Request ID: " +  requestId);
                         throw new InternalErrorException("Internal Server Error", statusCode);
                     }
-                    // If retries enabled, 500 InternalError should get retried unconditionally
+                    // If retries enabled, 500 InternalError should get retried unconditionally,
+                    // and so should 503 ServiceUnavailable (TIP-1023).
                     retryRequest = true;
-                    if (statusCode == 503) {
```
commandlinegirl: Here, the response from the server was consume when the status code is 503. We are removing this functionality in this commit, do we want to remove it?

The thing is that if the header didn't contain (or contained malformed) "retry-after" we set retry to be the default 60 seconds. Maybe instead we should set it to exponential backoff here? We'd keep reading the "retry-after" header but at the same time we'd have more clever retry with exponential backoff.

commandlinegirl:
> > > Please see my comments earlier. Not sure if this change is needed.
> > 
> > 
> > I think we should proceed with it. Exponential backoff is better than forcing it to retry and wait a full 60 seconds. This Java codepath is only used by TIP at this point as far as I know (not the cmd line dx toolkit invocations) so we can have our way with it to some extent.
> 
> yep. @geetduggal -> what do you think? Also @wkoetke -> could you please check what the python sdk does? Is it the same or different?

I'm in favor of exponential backoff but would keep the "retry-after" header handling.

src/java/src/main/java/com/dnanexus/DXHTTPRequest.java
ptn24: Is the retry-after header always included?

ptn24: Also, what is the behavior in the Python bindings? We should synchronize the implementations. The support for disabling should not impact the default behavior right?

commandlinegirl: @ptn24 It's not always included and we handle that [here](https://github.com/dnanexus/dx-toolkit/commit/968fdd67433dcea7c6b7eb14ce2536991829b9ab#diff-6122e848fa81a7baae8083db2bb6c3bdR338)

commandlinegirl:
> > > > > Please see my comments earlier. Not sure if this change is needed.
> > > > 
> > > > 
> > > > I think we should proceed with it. Exponential backoff is better than forcing it to retry and wait a full 60 seconds. This Java codepath is only used by TIP at this point as far as I know (not the cmd line dx toolkit invocations) so we can have our way with it to some extent.
> > > 
> > > 
> > > yep. @geetduggal -> what do you think? Also @wkoetke -> could you please check what the python sdk does? Is it the same or different?
> > 
> > 
> > I'm in favor of exponential backoff but would keep the "retry-after" header handling.
> 
> @commandlinegirl , what do you mean by "keep the retry-after" header handling? What portion of it would you keep? How would keeping that also allow exponential backoff?

@wkoetke By "retry-after" handling I mean [the part](https://github.com/dnanexus/dx-toolkit/blob/72cb67af8a9c47570dab322cf571803ace81b1c3/src/java/src/main/java/com/dnanexus/DXHTTPRequest.java#L374:L383) where we get the response the server and check if there is a field "retry-after" time in the header set (which takes a precedence over wait time of the default 60s). I'd keep it all. But in the case when the response doesn't contain this field (or it's malformated) I wouldn't override it with 60s but with exponentially growing wait-time (like we do for other error codes) for a specified NUM_RETRIES (it would mean that that [this statement](https://github.com/dnanexus/dx-toolkit/blob/72cb67af8a9c47570dab322cf571803ace81b1c3/src/java/src/main/java/com/dnanexus/DXHTTPRequest.java#L397) won't hold which is OK; we could check how long we retry for 503 in dx-toolkit (and how it's generally implemented there, in dx-download-agent we retry for up to 30 min total)). 

In short, I'd keep the whole functionality as is except for defaulting to exp backoff instead of constant 60s.

PR #469, opened 2019-01-31
Support --phi flag for new projects

jtratner:
Right now I end up having users do:

```
dx api project new '{"name": "myname", "containsPHI": true}'
```

which is a little awkward to do.

Instead now you could do:

```
dx new project myname --phi
```

to have it include PHI.

Added a test case but wasn't sure how to run the tests.

In my hacky way of running it, got this to work:

```
$ pipenv install && pipenv run dx new project myname --phi
$ dx describe project-FV9Zvk80y1x2kzqP7Vy5Z89z
Result 1:
ID                  project-FV9Zvk80y1x2kzqP7Vy5Z89z
Class               project
Name                myname
Summary
Billed to           org-counsyl
Access level        ADMINISTER
Region              aws:us-east-1
Protected           false
Restricted          false
Contains PHI        true
Created             Thu Jan 31 11:37:09 2019
Created by          jtratner
Last modified       Thu Jan 31 11:37:09 2019
Data usage          0.00 GB
Sponsored data      0.00 GB
Storage cost        $0.000/month
Sponsored egress    0.00 GB used of 0.00 GB total
At spending limit?  false
Tags                -
Properties          -
downloadRestricted  false
defaultInstanceType "mem2_hdd2_x2"
```

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -4929,7 +4929,7 @@ parser_new_project.add_argument('--region', help='Region affinity of the new pro
 parser_new_project.add_argument('-s', '--select', help='Select the new project as current after creating',
                                 action='store_true')
 parser_new_project.add_argument('--bill-to', help='ID of the user or org to which the project will be billed. The default value is the billTo of the requesting user.')
-parser_new_project.add_argument('--phi', help='add PHI protection to project', default=False,
+parser_new_project.add_argument('--phi', help='Add PHI protection to project', default=False,
                                 action='store_true')
 parser_new_project.set_defaults(func=new_project)
 register_parser(parser_new_project, subparsers_action=subparsers_new, categories='fs')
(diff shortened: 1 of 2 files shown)


commandlinegirl:
Thanks! I'll run the test.

jtratner:
btw - where did the test end up going? just wondering best practice here

commandlinegirl:
I put it here: https://github.com/dnanexus/dx-toolkit/commit/80abc877a16272d718bca062205b03c47478773d

I changed the test to be more indirect - PHI projects tests are a bit tricky since the regular jenkins test user doesn't have permissions to create them.

Generally, tests that test `dx` CLI tools are in [test_dxclient.py](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/test_dxclient.py) while Python bindings and functions are tested in [test_dxpy](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/test_dxpy.py). 

You can run them with (how-to is at the top):

https://github.com/dnanexus/dx-toolkit/blob/master/build/run_python_integration_tests.py

So for example I run the test with:
```
build/run_python_integration_tests.py \
    --tests test_dxpy.TestDXClientNewProject.test_dx_create_new_project_with_phi
```

Issue #472, opened 2019-02-06
aria2c raising exception when running `dx download` on large files using instance type with 32 cores


acchiao:
When running `dx download` on large (~100 GB) files onto an instance with 32 cores using the `aria2c` downloader, it will raise an exception.

```
+ dx download '{"$dnanexus_link": "file-<redacted>"}' -o input
Exception: [AbstractOptionHandler.cc:69] errorCode=28 We encountered a problem while processing the option '--max-connection-per-server'.
  -> [OptionHandlerImpl.cc:187] errorCode=1 max-connection-per-server must be between 1 and 16.
Usage:
 -x, --max-connection-per-server=NUM The maximum number of connections to one
                              server for each download.
                              Possible Values: 1-16
                              Default: 1
                              Tags: #basic, #http, #ftp
Traceback (most recent call last):
  File "/usr/share/dnanexus/lib/python2.7/site-packages/dxpy/bindings/dxfile_functions.py", line 195, in _download_symbolic_link
    subprocess.check_call(cmd, stderr=subprocess.STDOUT)
  File "/usr/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
CalledProcessError: Command '[u'aria2c', u'--check-certificate=false', u'-s', '32', u'-x', '32', u'-o', u'input', u'-d', u'/', u'http://<redacted>']' returned non-zero exit status 28
Failed to call download: [u'aria2c', u'--check-certificate=false', u'-s', '32', u'-x', '32', u'-o', u'input', u'-d', u'/', u'http://<redacted>']
```

This is due to `aria2c` having a [hard coded limit of 16 possible connections](https://github.com/aria2/aria2/blob/8985d66e71f980e7d2765753800078f47761f1ba/src/OptionHandlerFactory.cc#L422) for a server. See issue [here](https://github.com/aria2/aria2/issues/580). It has been deemed a non-issue ([#648](https://github.com/aria2/aria2/pull/648#issuecomment-217971648)), but it is possible to edit the value and recompile ([#729](https://github.com/aria2/aria2/issues/729#issuecomment-242366689)).

The `--split` also seems to be restricted by the `--max-connection-per-server` option. See [here](https://aria2.github.io/manual/en/html/aria2c.html?highlight=split#cmdoption-s).

---

The line below causes `aria2c` to exceed the max limit of 16 connections when using an instance with 32 cores.

[...]

PR #473, opened 2019-02-06
DEVEX-1033 Limit aria2c connections to <= 16 per server for symlink downloads

kpjensen:
https://github.com/dnanexus/dx-toolkit/issues/472

### DIFF
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -184,7 +184,9 @@ def _download_symbolic_link(dxid, md5digest, project, dest_filename):
         cmd += ["-O", dest_filename, url]
     else:
         print("aria2c found in path so using that instead of wget \n")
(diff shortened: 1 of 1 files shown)


src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -184,7 +184,9 @@ def _download_symbolic_link(dxid, md5digest, project, dest_filename):
         cmd += ["-O", dest_filename, url]
     else:
         print("aria2c found in path so using that instead of wget \n")
-        cmd = ["aria2c", "--check-certificate=false", "-s", str(multiprocessing.cpu_count()), "-x", str(multiprocessing.cpu_count())]
+        # aria2c does not allow more than 16 connections per server
+        max_connections = 16 if multiprocessing.cpu_count() > 16 else multiprocessing.cpu_count()
```
commandlinegirl: Or maybe:

```suggestion
        max_connections = min(16, multiprocessing.cpu_count())
```

orodeh: How about using the python `min` function?  
`min(4, 6) = 4`

Issue #426, opened 2018-10-30
dxpy3 only allows utf-8 encoding


anujkumar93:
Currently, dxpy3 uses 'utf-8' while reading files. `data.decode("utf-8")` is invoked during `dxpy.DXFile.read()`. This does not allow other types of files (like gzip) to be read using dxpy. For example, the following fails for dxpy3:
```
>>> import dxpy
>>> dxpy.upload_local_file(filename='s_3_2126.bcl.gz', project='project-FK8pJ100x54yqG4x7XyK62v2')
<dxpy.bindings.dxfile.DXFile object at 0x1028592b0: file-FPG51X80x54zvv365P3k9ZBp (project-FK8pJ100x54yqG4x7XyK62v2)>
>>> dxpy.DXFile(dxid='file-FPG51X80x54zvv365P3k9ZBp', project='project-FK8pJ100x54yqG4x7XyK62v2').read()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/akumar/stor/.delete/lib/python3.6/site-packages/dxpy/bindings/dxfile.py", line 955, in read
    return data.decode("utf-8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 1: invalid start byte
```

anujkumar93:
@jtratner FYI

orodeh:
@anujkumar93, @jtratner I believe the issue goes deeper. With python-2 data read from a text file is a string, so is data read from a binary (e.g. gzipped) file. However, with python-3, these are two distinct types. Data read from a text file is a string, data read from a binary file is bytes. The distinction percolates throughout the dxpy3 code. Since I had to choose which kind of file to support in dxpy3, I chose text files.

It is worth thinking about your use case. Constructive suggestions welcome.

jtratner:
@orodeh To me, the underlying issue is that `DXFile` needs to support a binary `mode` argument (as well as an `encoding` argument), which would allow the user to specify how they want to handle their data. Currently only `r`, `w` and `a` are supported (i.e., only text modes).  (mode of a file pointer determines whether it produces text (AKA `str` / `unicode`) or `bytes`.)

It looks like the `write()` method is almost there, just the constructor of DXFile ought to take in an encoding argument (that defaults to `locale.getpreferredencoding(False)`) and also allow binary modes. 

I'd suggest making all the helper methods produce bytes that you then decode appropriately if a text pointer is required.

---

[...]

PR #474, opened 2019-02-06
DEVEX-917 Binary files on python3

orodeh:
Address issue https://github.com/dnanexus/dx-toolkit/issues/426. 

1. Support binary mode in file open. 
2. Update the CLI to make use of binary mode on python3. By default, we assume files are binary. This will work for ASCII inputs, and gzip/tar files, which we assume are the majority. If you need to use unicode, the `--textual` flag is provided.
3. Additional tests.

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -1930,7 +1930,7 @@ def cat(args):
 
         # We assume the file is binary, unless specified otherwise
         mode = "rb"
-        if args.textual is True:
+        if args.unicode_text is True:
             mode = "r"
         try:
             dxfile = dxpy.DXFile(entity_result['id'], mode=mode)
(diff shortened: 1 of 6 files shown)


src/python/dxpy/scripts/dx.py
```diff
@@ -1967,12 +1973,15 @@ def head(args):
...
-                        break
+                try:
+                    handler._read_bufsize = 1024*32
+                    for line in handler:
+                        print(line)
+                        counter += 1
+                        if counter == args.lines:
+                            break
+                except UnicodeDecodeError:
+                    sys.stdout.write("File contains binary data")
```
commandlinegirl: I was thinking we could add such a nicer printout for Py2 but let's not focus of the old (that's going away hopefully) too much!

PR #475, opened 2019-02-07
Attempt to correct aria2c output path

kpjensen:
Attempt to strip the path from the output filename for the call to aria2c when downloading symlinks.
```
root@kjdev:/dir0# dx download test.gz
# Generates a command with '-o'
aria2c -o /dir0/test13.gz
Download Results:
gid   |stat|avg speed  |path/URI
======+====+===========+=======================================================
9069c1|OK  |    60MiB/s|/dir0//dir0/test13.gz
```
The expected behavior would be to download test.gz to the current directory.
This seems fine but aria2c assumes `-d $(pwd)` if no directory argument is provided. 
```
root@kjdev:/dir0# aria2c -h |grep dir
 -d, --dir=DIR                The directory to store the downloaded file.
                              Possible Values: /path/to/directory
                              Default: /dir0
```
Thus the file is actually downloaded to `/dir0/dir0/test.gz`
```
root@kjdev:/dir0# tree /dir0/
/dir0/
└── dir0
    └── test.gz
```

After splitting the path and supplying both '-o' and '-d
```
root@kjdev:/dir0# dx download test.gz
Download Results:
gid   |stat|avg speed  |path/URI
======+====+===========+=======================================================
f7241b|OK  |    60MiB/s|/dir0/test.gz

# Now generates a command with '-o' and '-d' args
# aria2c -o test.gz -d /dir0

root@kjdev:/dir0# tree /dir0/
/dir0/
└── test.gz
```

### DIFF
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -191,7 +191,7 @@ def _download_symbolic_link(dxid, md5digest, project, dest_filename):
         # If '-d' arg not provided, aria2c uses current working directory
         cwd = os.getcwd()
         directory, filename = os.path.split(dest_filename)
-        directory = cwd if dirname in ["", cwd] else directory
+        directory = cwd if directory in ["", cwd] else directory
         cmd += ["-o", filename, "-d", directory, url]
 
     try:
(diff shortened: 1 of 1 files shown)


kpjensen:
Updated to convert the directory to absolute path.
```
root@kjdev:/dir0# dx download -o ../home/testing.gz test.gz
# Generated command
aria2c -o testing.gz -d /home
```

PR #478, opened 2019-02-08
DEVEX-1036 Add support for --ignore-reuse and --ignore-reuse-stage flags for workflows

commandlinegirl:
This commit adds an ignoreReuse option for building and running workflows. 

I looked into the posssibility of reusing "--ignore-reuse" (that we have for apps now) but couldn't really use it without breaking customers' runs. There is an open [issue](https://bugs.python.org/issue9338) with using nargs (argparse optionals with nargs='?', '*' or '+' can't be followed by positionals) which means we cannot use one param that accepts 0 or more arguments.

If you had any other ideas on how we could do that I'd be happy to change it - thanks!

### DIFF
--- a/src/python/dxpy/bindings/dxapp.py
+++ b/src/python/dxpy/bindings/dxapp.py
@@ -363,7 +363,6 @@ class DXApp(DXObject, DXExecutable):
                                           **kwargs)["id"])
 
     def _get_run_input(self, executable_input, **kwargs):
-        # May need to be changed when workflow apps are enabled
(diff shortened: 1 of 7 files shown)


commandlinegirl:
@kpjensen friendly ping re the review, thanks!:)

commandlinegirl:
@kpjensen Now I accept `--ignore-reuse` (boolean) for workflows as well, translating it to "*" before sending a request. I also added the mutually exclusive group for these three options.

PR #480, opened 2019-02-08
DEVEX-1038 dx symlink tests

orodeh:


orodeh:
Kurt, here are an initial test for symlinks. I already found a python3 bug with it. 

Thanks,
       Ohad.

PR #486, opened 2019-02-14
Skip symlinks test on mac

kpjensen:
Failing on mac due to missing `md5sum`.


kpjensen:
Installing `md5sum` and `wget` on Mac test node.

PR #491, opened 2019-02-21
DEVEX-1007 Add --instance-count to dx run

commandlinegirl:


src/python/test/test_dxclient.py
```diff
@@ -7293,114 +7523,6 @@ def test_build_app_with_resources(self):
             container_content = dxpy.api.container_list_folder(app_container, {"folder": "/"})
             self.assertIn(file_id, [item["id"] for item in container_content["objects"]])
 
-    @unittest.skipUnless(testutil.TEST_ISOLATED_ENV,
```
commandlinegirl: Moving this test to a separate class for Spark Apps

commandlinegirl:
@ocbrewing Could you have a look at it when you get a chance? 
@ptn24 I now enumerate different scenarios for merging requested and app's clusterSpecs to in the key function docstring.

src/python/dxpy/scripts/dx.py
```diff
@@ -2869,6 +2880,80 @@ def check_for_special_access(access_spec):
...
+    with the value the user passed to dx run for each entrypoint.
+
+    Instance count is currently the only clusterSpec's field the user is allowed to change
+    at runtime.
+
+    A few scenarios when requesting instance count for different entrypoints with dx run 
+    and the resulting merged systemRequirements (merged_cluster_spec):
+
+    Requested: {"*": 5}
+    App doc: {"main": "clusterSpec": {"initialInstanceCount": 7, bootstrapScript: "x.sh"},
```
ocbrewing: not sure that it matters here, but type, version, & ports are also fields that would need to be copied over. I assume those are handled correctly as well.

commandlinegirl: Yes, every other field is copied (in this deep copy [here](https://github.com/dnanexus/dx-toolkit/blob/53089e78f4f87fa951f3a5663ef74ad76fc53b88/src/python/dxpy/scripts/dx.py#L2943)), I'll add a note to the docstring

src/python/dxpy/scripts/dx.py
```diff
@@ -2869,6 +2880,80 @@ def check_for_special_access(access_spec):
...
+        for app_entrypoint, reqs in app_sys_reqs.items():
+            if "clusterSpec" in reqs:
+                merged_cluster_spec[app_entrypoint] = {"clusterSpec": copy.deepcopy(reqs["clusterSpec"])}
+                merged_cluster_spec[app_entrypoint]["clusterSpec"]["initialInstanceCount"] = requested_count
+
+    merged_cluster_spec = {}
+    entrypoint_to_instance_count = instance_count_to_sys_reqs(instance_count)
+
+    for entrypoint, requested_count in entrypoint_to_instance_count.items():
+        if entrypoint == "*":
```
ocbrewing: seems like you can't guarantee that '*' is going to run first or after something else, right? In the case that it runs first, this works great - but if it comes up later, it will overwrite another entrypoint that may have set some other value.  Not totally sure if I'm correct on this, but just raising that it may be an issue.

ptn24: I have not reviewed the implementation details yet, but we could either process `"*"` up front, or skip it and process it at the end. Might simplify the implementation

src/python/dxpy/scripts/dx.py
```diff
@@ -2803,12 +2848,13 @@ def run_body(args, executable, dest_proj, dest_path, preset_inputs=None, input_n
         "stage_instance_types": args.stage_instance_types,
         "stage_folders": args.stage_folders,
         "rerun_stages": args.rerun_stages,
+        "merged_cluster_spec": merged_cluster_spec,
```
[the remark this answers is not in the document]
commandlinegirl: No, this is an argument for [executable.run()](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/scripts/dx.py#L2709) and is later processed along other arguments into an API input here: https://github.com/dnanexus/dx-toolkit/blob/53089e78f4f87fa951f3a5663ef74ad76fc53b88/src/python/dxpy/bindings/dxapplet.py#L61

src/python/dxpy/scripts/dx.py
```diff
@@ -2775,13 +2775,58 @@ def run_batch_all_steps(args, executable, dest_proj, dest_path, input_json, run_
...
                                   **(args.instance_type or {}))
 
+    if args.sys_reqs_from_clone and not isinstance(args.instance_count, str):
+        args.instance_count = dict({fn: reqs['clusterSpec'] for fn, reqs in list(args.sys_reqs_from_clone.items())},
+                                  **(args.instance_count or {}))
+
+    # Since full clusterSpec must be passed to the API server, merge the cluster
+    # spec defined in the app with the instance_count by taking the app's spec
+    # and overwriting initialInstanceCount with args.instance_count.
+    merged_cluster_spec = {} # e.g. {'fn': {'clusterSpec': {initialInstanceCount: .., version: ..}}}
```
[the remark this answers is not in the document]
commandlinegirl: Hm, one issue with this approach is that we will be requesting even if the user wants to overwrite just one entrypoint. In the current one I iterate over the entrypoints that are requested to be modified and sent them with the API.

src/python/dxpy/scripts/dx.py
commandlinegirl: Thanks! I took it into account in the previous, slightly different implementation but then lost it when updating it. I now process if up front.

commandlinegirl:
@ptn24 friendly ping re the review, thanks!

src/python/dxpy/bindings/dxapplet.py
```diff
@@ -58,6 +58,17 @@ def _get_run_input_common_fields(executable_input, **kwargs):
         if kwargs.get('instance_type') is not None:
             run_input["systemRequirements"] = instance_type_to_sys_reqs(kwargs['instance_type'])
 
+        if kwargs.get('merged_cluster_spec'):
```
ptn24: This part mostly lg, but I do not like how we handle instance types and cluster specifications separately here. Consider merging the block at L58 with this block, using one of the following

```
# Option 1 -- use functions

def resolve_system_requirements_input(instance_type=None, cluster_specification=None):
    system_requirements = {}
    # Code
    return system_requirements

--//--

run_input["systemRequirements"] = resolve_system_requirements_input(instance_type=kwargs.get("instance_type"), cluster_specification=kwargs.get("merged_cluster_spec"))
```

or

```
# Option 2 -- use classes

class SystemRequirementsDict(object):
    def __init__(self, ...): ...
    @classmethod
    def from_instance_type(cls, instance_type=None):
        return cls(...)
    @classmethod
    def from_cluster_specification(cls, cluster_specification=None):
        return cls(...)

    def __add__(self, srd):
        """Add only, raise on collisions"""
        if not isinstance(srd, SystemRequirementsDict): raise
        # Code
        return SystemRequirementsDict(...)

    def add_override(self, srd)
        # Code
        return SRD(...)

    def as_dict(self): ...
--//--

run_input["systemRequirements"] = (SystemRequirementsDict.from_instance_type(...) + SRD.from_cluster_specification(...)).as_dict()
```

Might improve readability and unit-testability

src/python/dxpy/cli/parsers.py
```diff
@@ -312,6 +317,13 @@ def process_instance_type_arg(args, for_workflow=False):
             # is a string
             args.instance_type = _parse_inst_type(args.instance_type)
 
+def process_instance_count_arg(args):
+    if args.instance_count:
+        if not isinstance(args.instance_count, basestring):
+            args.instance_count = _parse_inst_count(args.instance_count[-1])
```
ptn24: Why `[-1]`?

src/python/dxpy/utils/__init__.py
```diff
@@ -301,6 +301,22 @@ def instance_type_to_sys_reqs(instance_type, entrypoint="*"):
     else:
         raise DXError('Expected instance_type field to be either a string or a dict')
 
+def instance_count_to_sys_reqs(instance_count, entrypoint="*"):
```
ptn24: If you use option 2 above, then you might be able to include this in it

src/python/dxpy/scripts/dx.py
```diff
@@ -2782,6 +2782,16 @@ def run_body(args, executable, dest_proj, dest_path, preset_inputs=None, input_n
         args.instance_type = dict({stage: reqs['instanceType'] for stage, reqs in list(args.sys_reqs_from_clone.items())},
                                   **(args.instance_type or {}))
 
+    if args.sys_reqs_from_clone and not isinstance(args.instance_count, str):
+        args.instance_count = dict({fn: reqs['clusterSpec'] for fn, reqs in list(args.sys_reqs_from_clone.items())},
```
ptn24: Consider `add_override` in option 2, if you go that route

src/python/dxpy/scripts/dx.py
```diff
@@ -2782,6 +2782,16 @@ def run_body(args, executable, dest_proj, dest_path, preset_inputs=None, input_n
...
+                                  **(args.instance_count or {}))
+
+    merged_cluster_spec = {}
+    if args.instance_count:
+        executable_describe = executable.describe()
+        app_sys_reqs = executable_describe['runSpec'].get('systemRequirements', {})
+        merged_cluster_spec = get_merged_cluster_spec(app_sys_reqs, args.instance_count)
```
ptn24: See `add_override`

src/python/dxpy/scripts/dx.py
```diff
@@ -2869,6 +2880,83 @@ def check_for_special_access(access_spec):
     else:
         run_batch_all_steps(args, executable, dest_proj, dest_path, input_json, run_kwargs)
 
+def get_merged_cluster_spec(app_reqs, instance_count):
```
ptn24: Any chance we could fold this into option 2?

src/python/dxpy/bindings/dxapplet.py
ptn24: As I reviewed, I think I started to lean towards option 2, but let me know your thoughts

src/python/dxpy/system_requirements.py
```diff
@@ -0,0 +1,182 @@
...
+System Requirements
++++++++++++++++++++
+
+A module containing utility methods useful for packing and unpacking
+system requirements.
+'''
+
+class SystemRequirementsDict(object):
+    def __init__(self, instance_type=None, cluster_spec=None):
+        self.instance_type = instance_type
```
ptn24: Might be good to document what `instance_type` and `cluster_spec` look like using small examples

src/python/dxpy/system_requirements.py
```diff
@@ -0,0 +1,182 @@
...
+            # By default, all entry points ("*") should use this instance type
+            return cls(instance_type={entrypoint: {"instanceType": instance_type_arg}})
+        elif isinstance(instance_type_arg, dict):
+            # instance_type is a map of entry point to instance type
+            return cls(instance_type={fn: {"instanceType": fn_inst} for fn, fn_inst in instance_type_arg.items()})
+        else:
+            raise DXError('Expected instance_type field to be either a string or a dict')
+
+    @classmethod
+    def from_cluster_spec(cls, app_sys_reqs, instance_count_arg):
```
ptn24: I think `app_sys_reqs` and `instance_count_arg` should generate separate SRDs, and `add_override` should be used to merge them

```
srd_default = SRD.from_sys_req(app_sys_reqs["systemRequirements"])
srd_runtime = SRD.from_instance_count(instance_count_arg)
srd_cluster_spec = srd_default.add_override(srd_runtime)
...
srd = srd_cluster_spec + srd_instance_type

--//--

class SRD(object):
    ...
    @classmethod
    def from_cluster_spec(cls, cluster_specs_by_entry): ...
    @classmethod
    def from_sys_req(cls, sys_req):
        # Resolve cluster_specs_by_entry from sys_req (might not be necessary)
        ...
    @classmethod
    def from_instance_count(cls, instance_count_arg, entry="*"):
        # Magic for resolving cluster_specs_by_entry
        ...
    def add_override(self, srd):
        # Magic for overriding self with srd
        ...
```

This way, the nasty logic is encapsulated in `add_override`, and we can test that extensively. Could help with readability too

WDYT?

src/python/test/test_dxclient.py
```diff
@@ -6958,6 +6959,244 @@ def _create_global_workflow(version):
...
+
+            self.assertEqual(azure_sys_reqs["main"]["clusterSpec"]["bootstrapScript"],
+                             "src/azure:westus_main_clusterBootstrap.py")
+            with open("cluster_app/src/azure:westus_main_clusterBootstrap.py") as f:
+                self.assertEqual(f.read(), bootstrap_code_azure)
+
+            # now rebuild with the result of `dx get` and verify that we get the same result
+            build_and_verify_bootstrap_script_inlined("cluster_app")
+
+    def test_get_cluster_spec_for_app_with_named_entrypoint(self):
```
ptn24: You probably already had this in mind, but I think these are unit tests that can live elsewhere. I think the tests in this file are integration tests right?

commandlinegirl: Yeah, but this deserves another ticket (we don't have a separate place for unit test atm).

ptn24: Not suggesting this per se, but I think some unit tests live in https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/test_dxpy_utils.py

src/python/dxpy/system_requirements.py
commandlinegirl: The instance_count is just a number (or dict) passed by the user, so I don't think it's possible to create a valid SRD from it on it's own (the invariant is that SRD.as_dict() can be a "systemRequirements" passed in the /run api call).

ptn24: You are mostly right about the invariant, but it can be interpreted more flexibly. `as_dict` just needs to provide a systemRequirements-like object that can be passed to the API

When instance_count is a number, what do we assume about it? We map it to `{"*": {"clusterSpec": "initialInstanceCount"}}`. We do not need to worry about the rest of `"*.clusterSpec"`

When instance_count is a dict (e.g., `{"*": 2, "foo": 4}`), we map it to `{"*": {"clusterSpec": {"initialInstanceCount": 2}}, "foo": {"clusterSpec": {"initialInstanceCount": 4}}}`

If the instance-count--based SRD was passed directly to the API without downstream changes, then the call would fail on the API side

So this is where `add_override` can handle adding the instance-count--based SRD to the default SRD

--//--

Alternatively you could start with the default SRD, and add a separate method for overriding based on the instance count

```
class SRD(object):
    @classmethod
    def from_sys_req(...): ...
    def override_instance_count(self, instance_count):
        # Nasty logic
```

In this design, the instance count is treated differently, and would never generate an SRD on its own

commandlinegirl:
I moved and updated the unit tests (aggregated into only three, I'll separate them out in the last step), updated SRD class and the code around it. Now running jenkins tests on the branch.

commandlinegirl:
@ptn24 py2 and py3 tests passed, if you could have a final look, thanks!

src/python/dxpy/scripts/dx.py
```diff
@@ -2803,6 +2804,21 @@ def run_body(args, executable, dest_proj, dest_path, preset_inputs=None, input_n
         args.instance_type = dict({stage: reqs['instanceType'] for stage, reqs in list(args.sys_reqs_from_clone.items())},
                                   **(args.instance_type or {}))
 
+    if args.sys_reqs_from_clone and not isinstance(args.instance_count, str):
+        # extract instance counts from cloned sys reqs and override them with args provided with "dx run"
+        args.instance_count = dict({fn: reqs['clusterSpec']['initialInstanceCount']
+                                        for fn, reqs in list(args.sys_reqs_from_clone.items()) if 'clusterSpec' in reqs},
+                                   **(args.instance_count or {}))
```
ptn24: Would this work if `args.instance_count` was just a string versus a dict? Would it be possible to use SRD here?

src/python/dxpy/system_requirements.py
```diff
@@ -0,0 +1,192 @@
...
+            raise DXError("Expected '_type' to be either 'all', 'clusterSpec', or 'instanceType'")
+
+        if _type == 'all':
+            return cls(system_requirements)
+
+        extracted = defaultdict(dict)
+        for entrypoint, req in system_requirements.items():
+            if _type in req:
+                extracted[entrypoint][_type] = req[_type]
+        return cls(dict(extracted))
```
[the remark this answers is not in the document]
commandlinegirl: I want to pass (and save in SRD.entrypoints) dict instead of collections.defaultdict that's why I'm returning `extracted` as dict.

src/python/dxpy/scripts/dx.py
commandlinegirl: If the args.instance_count is a string, the wildcard entry point will used and the count needs to overwrite values entry points (so we don't care about cloned vals).

Both args.instance_count and args.instance_type set here from cloning could be set as SRD but it'd require some adjustments down below. The SRD for instance types is created later, in `dxapplet.__get_run_input_common_fields()` so that `instance_type` can be parsed in one place regardless whether we're getting it from CLI args or from the `executable.run()` method. The `run(instance_count)` is not implemented yet since it requires again some more heavy lifting (since we require app description to process `instance-count` properly and the `run()` needs some more reorg for that purpose, to maybe not call another `describe` if it was done before).

ptn24:
```
>>> dict({"foo": 2}, **(None or {"bar": 4}))
{'foo': 2, 'bar': 4}
>>> dict({"foo": 2}, **("baz" or {"bar": 4}))
  File "<stdin>", line 1, in <module>
TypeError: type object argument after ** must be a mapping, not str
>>> dict({"foo": 2}, **({"foo": 4}  or {"bar": 4}))
{'foo': 4}
```

commandlinegirl: Right, but in the condition we exclude a string:
`... and not isinstance(args.instance_count, str)`

We want to exclude it since a string is where a wildcard '*' needs to be applied, so we don't care about whatever values are in the system requirements from the cloned job - we'll overwrite it anyway.

PR #492, opened 2019-02-22
Add samlIdP and jobReuse to  `dx update org-id`

kpjensen:
To allow users to use `dx update org` instead of requiring `dx api org-id update '{}'`


kpjensen:
I added a test for `--saml-idp`, but not for `--enable-job-reuse` since updating the `policies.jobReuse` field is restricted to billTos which have job reuse enabled in the backend.

src/python/test/test_dxclient.py
```diff
@@ -5750,6 +5750,15 @@ def get_name_and_policies(org_id=None):
         self.assertEqual(new_org_name, cur_org_name)
         self.assertEqual(new_org_policies, exp_org_policies)
 
+        # --saml-idp
+        proposed_idp = "samlprovider"
+        exp_org_policies = dict(cur_org_policies, restrictProjectTransfer=proposed_idp)
```
commandlinegirl: It looks like we don't need `exp_org_policies` for this particular test.

kpjensen: Thanks! Removed.

PR #496, opened 2019-02-27
dx org update args fix

kpjensen:
Fix failing `test_org_update`

### DIFF
--- a/src/python/dxpy/cli/org.py
+++ b/src/python/dxpy/cli/org.py
@@ -242,7 +242,7 @@ def _get_org_update_args(args):
     if args.saml_idp is not None:
         org_update_inputs["samlIdP"] = args.saml_idp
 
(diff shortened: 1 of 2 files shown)


src/python/dxpy/cli/org.py
```diff
@@ -242,14 +242,17 @@ def _get_org_update_args(args):
     if args.saml_idp is not None:
         org_update_inputs["samlIdP"] = args.saml_idp
```
commandlinegirl:
```suggestion
if args.enable_job_reuse and args.disable_job_reuse:
    raise DXCLIError("--enable-job-reuse and --disable-job-reuse cannot be used simultaneously")

```

kpjensen: This should be handled by argparse since they are mutually exclusive
`update_job_reuse_args = parser_update_org.add_mutually_exclusive_group(required=False)`

commandlinegirl: Ah, cool!

PR #501, opened 2019-03-11
SCI-1325 Add context manager functions from dx_utils.py

msimbirsky:
Added context manager functions from dx_utils.py


msimbirsky:
@jdidion I added your contextmanagers for `temp` and `fifo` but it seems like `temp` is just a re-implementation of `tempfile.TemporaryFile`. Are there situations where you would want to use this instead?

src/python/dxpy/sugar/context_handlers.py
```diff
@@ -0,0 +1,149 @@
...
+import os
+import sys
+import tempfile
+import subprocess
+import contextlib
+import shutil
+
+
+@contextlib.contextmanager
+def set_env(**environ):
```
jdidion: I suggest two changes:
1. Have this function take a dict arg, rather than **kwargs
2. Add a 'clear' kwarg that controls whether the existing environment is cleared or preserved when updating to the new environment.

src/python/dxpy/sugar/context_handlers.py
```diff
@@ -0,0 +1,149 @@
...
+    os.environ.update(environ)
+    try:
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(old_environ)
+
+@contextlib.contextmanager
+def cd(new_path=None, temp_dir=None):
+    """Context manager that changes the current working directory to `new_path`,
```
jdidion: I don't think temp_dir arg is needed. You could just have it work such that:
* If no argument is specified, create a temp dir (using tempfile.mkdtemp), cd to it and cd back and delete the directory before returning
* If an argument is specified and the directory doesn't exist, create it, cd to it, and cd back and delete the directory before returning
* If an argument is specified and the directory does exist, cd to the directory, cd to it, and cd back but don't delete the directory before returning.

You could have 'create' and 'cleanup' kwargs for the user to override the default behavior (if for example they don't want the tempdir to be deleted).

jdidion:
Actually, I remembered why I needed temp - see comment here: https://github.com/dnanexus/dx-toolkit/pull/511

msimbirsky:
I'm not following the comment:
```
The tmpfile contextmanager cannot be replaced by tempfile.TemporaryFile. It is needed in situations where the tempfile needs to be written from python and then closed and read from the subprocess.
```

Can you give an example of such a situation?

src/python/dxpy/sugar/context_handlers.py
msimbirsky: I think using **kwargs makes it more readable, since you can do:
`with set_env(PYTHONPATH=''): ...` or pass in a dictionary with multiple environment variables to update.

Having a "clear" or "overwrite" option is a good idea.
@Damien-Black Do you have an opinion on this?

src/python/dxpy/sugar/context_handlers.py
msimbirsky: I think "create" is redundant since a non-existent path would always have to be created, where as an existent path should be cd'd to without overwriting it. A "cleanup" command makes sense. We could replace `new_path` and `temp_dir` with just a `target_path` and `delete_on_exit` options. The only concern is user error if they cd to a root directory and then enable `delete_on_exit`.

msimbirsky:
I think the same function can be handled by: `tempfile.NamedTemporaryFile(mode="wt", delete=False)` but would need to test this depending on context.

src/python/dxpy/sugar/context_handlers.py
jdidion: That is why I don't like the delete_on_exit arg. IMO the context manager should only be able to delete a directory that it creates itself.

msimbirsky: That makes sense to me.

jdidion:
> I'm not following the comment:
> 
> ```
> The tmpfile contextmanager cannot be replaced by tempfile.TemporaryFile. It is needed in situations where the tempfile needs to be written from python and then closed and read from the subprocess.
> ```
> Can you give an example of such a situation?

Sure, have a look at transfers.py in #511 around line 326. Here I need to create a temp file to contain the names of the files to include in the tar. Then I need to close the file so when I call tar --files-from in the subprocess it can read from the file. But if I use tempfile.TemporaryFile, the file is deleted as soon as it's closed. If I were just reading to and writing from the file in python it would be fine since the file is opened with "w+" but it doesn't work in the case where the writer is python and the reader is a separate process.

If you don't think this is generally useful, I'm fine with not including it in context.py and just keeping it internal to transfers.py.

msimbirsky:
> > I'm not following the comment:
> > ```
> > The tmpfile contextmanager cannot be replaced by tempfile.TemporaryFile. It is needed in situations where the tempfile needs to be written from python and then closed and read from the subprocess.
> > ```
> > Can you give an example of such a situation?
> 
> Sure, have a look at transfers.py in #511 around line 326. Here I need to create a temp file to contain the names of the files to include in the tar. Then I need to close the file so when I call tar --files-from in the subprocess it can read from the file. But if I use tempfile.TemporaryFile, the file is deleted as soon as it's closed. If I were just reading to and writing from the file in python it would be fine since the file is opened with "w+" but it doesn't work in the case where the writer is python and the reader is a separate process.
> 
> If you don't think this is generally useful, I'm fine with not including it in context.py and just keeping it internal to transfers.py.

I don't think it's necessary in either module. In either case I will remove it here.

src/python/dxpy/sugar/context.py
```diff
@@ -0,0 +1,191 @@
...
+    os.environ.update(environ)
+    try:
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(old_environ)
+
+
+class cd:
+    """
```
jdidion: I'm curious why this one is done with a class vs @contextmanager decoration on function.

src/python/dxpy/sugar/context.py
```diff
@@ -0,0 +1,191 @@
...
+def fifo(name=None):
+    """
+    Create a FIFO, yield it, and delete it before exiting.
+    Args:
+        name: The name of the FIFO, or None to use a temp name.
+    Yields:
+        The name of the FIFO
+    """
+    if name is None:
+        temp = tempfile.NamedTemporaryFile(delete=False)
```
jdidion: What is the difference between this and tempfile.mkstemp()[1]?

src/python/dxpy/sugar/context.py
msimbirsky: It was done that way historically. I don't think rewriting this as a @contextmanager decoration would add much clarity as this class has explicit __enter__ and __exit__ functions.

src/python/dxpy/sugar/context.py
msimbirsky: No technical difference, NamedTemporaryFile calls mkstemp() on the backend.

src/python/dxpy/sugar/context_handlers.py
jdidion: This is the last sticking point for me - I think it's important to be able to choose whether you want to start from the current environment or have a completely fresh environment. But I'm also fine with finishing this PR and adding that in a separate PR.

msimbirsky: I think that's an important distinction. I'll add that in this PR.

msimbirsky:
@jdidion I went with your proposed method signature for set_env and added an override argument

src/python/dxpy/sugar/context_handlers.py
Damien-Black: This PR looks good. Once we add a `clear` option to `set_env` we should move to merge it in.

src/python/dxpy/sugar/context.py
```diff
@@ -60,12 +60,14 @@ def __exit__(self, type, value, traceback):
         dxpy.set_workspace_id(self.job_workspace_id)
 
 @contextlib.contextmanager
-def set_env(**environ):
+def set_env(environ, override=False):
```
msimbirsky: @Damien-Black "clear" option is added in this commit

Damien-Black:
@msimbirsky forgot to point out. Test in dx-toolkit repository are a [packaged separate](https://github.com/dnanexus/dx-toolkit/tree/master/src/python/test) from dxpy. There are some utility modules and functions useful for writing tests. I would give [dxpy_testutil.py](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/dxpy_testutil.py) a glance and see if you could re-use some of the functionality (common set-ups and tear-downs, etc...) here in your test.

src/python/test/sugar/test_context.py
```diff
@@ -67,3 +69,6 @@ def test_cd_targetpath(self):
             self.assertNotEqual(prev_dir, curr_dir, "Failed to change dirs")
 
         self.assertTrue(os.path.exists("/tmp/some_path"), "Temp path should still exist")
```
Damien-Black: Must have missed this in a prior commit, but this test will fail to execute on non-unix systems. I recommend using the more portable [`tempfile.gettempdir`](https://docs.python.org/2/library/tempfile.html#tempfile.gettempdir). Something like:


```python
prev_dir = os.getcwd()
temp_dir = os.path.join(tempfile.gettempdir(), "some_path")

with context.cd(temp_dir, cleanup=False):
            curr_dir = os.getcwd()
            self.assertNotEqual(prev_dir, curr_dir, "Failed to change dirs")

        self.assertTrue(os.path.exists(temp_dir), "Temp path should still exist")

os.remove(temp_dir)
```

jdidion:
@Damien-Black can you finish this up and tag me to review?

jdidion:
@Damien-Black I've made your suggested changes and also done code clean-up. Please re-review.

Damien-Black:
@jdidion Looks good to me. Let me know if a review is needed after the conflicts are resolved.

PR #502, opened 2019-03-11
TIP-1022: return modified date as millis from epoch

wkoetke:
The api route listFolder now returns modified date as millis from epoch, so we need to convert it back to readable format in the dx cmd. The apiserver PR is: https://github.com/dnanexus/nucleus/pull/2644

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -2622,19 +2622,24 @@ def list_developers(args):
     except:
         err_exit()
 
+def render_timestamp(epochSeconds):
+    # This is the format used by 'aws s3 ls'
+    return datetime.datetime.fromtimestamp(epochSeconds//1000).strftime('%Y-%m-%d %H:%M:%S')
+
 def list_database_files(args):
     try:
         results = dxpy.api.database_list_folder(
             args.database,
             input_params={"folder": args.folder, "recurse": args.recurse, "timeout": args.timeout})
-
         for r in results["results"]:
             if (args.csv == True):
+                dateStr = render_timestamp(r["modified"]) if r["modified"] != 0 else ''
                 print("{}{}{}{}{}".format(
-                    r["modified"], DELIMITER(","), r["size"], DELIMITER(","), r["path"]))
+                    dateStr, DELIMITER(","), r["size"], DELIMITER(","), r["path"]))
             else:
+                dateStr = render_timestamp(r["modified"]) if r["modified"] != 0 else '                   '
                 print("{}{}{}{}{}".format(
-                    r["modified"], DELIMITER(" "), str(r["size"]).rjust(12), DELIMITER(" "), r["path"]))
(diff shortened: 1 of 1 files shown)


src/python/dxpy/scripts/dx.py
```diff
@@ -2622,19 +2622,24 @@ def list_developers(args):
...
             input_params={"folder": args.folder, "recurse": args.recurse, "timeout": args.timeout})
-
         for r in results["results"]:
             if (args.csv == True):
+                dateStr = render_timestamp(r["modified"]) if r["modified"] != 0 else ''
                 print("{}{}{}{}{}".format(
-                    r["modified"], DELIMITER(","), r["size"], DELIMITER(","), r["path"]))
+                    dateStr, DELIMITER(","), r["size"], DELIMITER(","), r["path"]))
             else:
+                dateStr = render_timestamp(r["modified"]) if r["modified"] != 0 else '                   '
```
durae: Please remove some extra whitespaces:

```suggestion
                dateStr = render_timestamp(r["modified"]) if r["modified"] != 0 else ''
```

src/python/dxpy/scripts/dx.py
```diff
@@ -2622,19 +2622,24 @@ def list_developers(args):
...
+
 def list_database_files(args):
     try:
         results = dxpy.api.database_list_folder(
             args.database,
             input_params={"folder": args.folder, "recurse": args.recurse, "timeout": args.timeout})
-
         for r in results["results"]:
             if (args.csv == True):
+                dateStr = render_timestamp(r["modified"]) if r["modified"] != 0 else ''
```
durae: you can move this above condition as you need dateStr in both if and else. An I would use `date_str` as we are in python now :)

src/python/dxpy/scripts/dx.py
wkoetke: Done

src/python/dxpy/scripts/dx.py
wkoetke: Done

src/python/dxpy/scripts/dx.py
```diff
@@ -2622,19 +2622,24 @@ def list_developers(args):
...
         results = dxpy.api.database_list_folder(
             args.database,
             input_params={"folder": args.folder, "recurse": args.recurse, "timeout": args.timeout})
-
         for r in results["results"]:
+            date_str = render_timestamp(r["modified"]) if r["modified"] != 0 else ''
             if (args.csv == True):
-                print("{}{}{}{}{}".format(
-                    r["modified"], DELIMITER(","), r["size"], DELIMITER(","), r["path"]))
+		print("{}{}{}{}{}".format(
```
durae: Does this need one more space? Otherwise LGTM.

durae: Ok, it doesn't. Just github playing with my eyes.

PR #503, opened 2019-03-11
Change download to skip folder if unnecessary

geetduggal:
We would originally always do a check on subfolders for all paths passed in to download. This skips that check if we do not believe we are attempting to download a folder.

### DIFF
--- a/src/python/dxpy/cli/download.py
+++ b/src/python/dxpy/cli/download.py
@@ -169,7 +169,7 @@ def download(args):
 
         matching_folders = []
         # project may be none if path is an ID and there is no project context
-        if project is not None:
+        if project is not None and folderpath is not None:
             colon_pos = get_first_pos_of_char(":", path)
             if colon_pos >= 0:
                 path = path[colon_pos + 1:]


src/python/dxpy/cli/download.py
```diff
@@ -169,7 +169,7 @@ def download(args):
 
         matching_folders = []
         # project may be none if path is an ID and there is no project context
-        if project is not None:
+        if project is not None and folderpath is not None:
```
commandlinegirl: We initialize some variables in this block (e.g. `strip_prefix `) that are used later. It seems to be guarded though by errors thrown in lines 185 and 198, so it looks okay to me just not easy to read.

commandlinegirl:
Running jenkins tests suite on this branch.

commandlinegirl:
One test fails with this commit (`test_dx_upload_download()`). Looking into it.

commandlinegirl:
One solution is to skip this validation if `--lightweight` option is used (after [this](https://github.com/dnanexus/dx-toolkit/pull/519) is merged). With this option the conflict will not result in an error but the folders will be ignored and only files downloaded.

PR #507, opened 2019-03-25
Improve link handling

jdidion:
These changes are to add full JBOR support to several core dx-toolkit functions:

* dxlink: add a field parameter that can be used for creating a JBOR link to a specific field.
* is_dxlink: recognize links with 'job' keys (instead of 'id' key)
* get_dxlink_ids: support JBOR links

### DIFF
--- a/src/python/dxpy/bindings/dxdataobject_functions.py
+++ b/src/python/dxpy/bindings/dxdataobject_functions.py
@@ -34,12 +34,14 @@ from . import __dict__ as all_bindings
 from ..exceptions import DXError
 from ..compat import basestring
 
-def dxlink(object_id, project_id=None):
+def dxlink(object_id, project_id=None, field=None):
     '''
     :param object_id: Object ID or the object handler itself
     :type object_id: string or :class:`~dxpy.bindings.DXDataObject`
     :param project_id: A project ID, if creating a cross-project DXLink
     :type project_id: string
+    :param field: A field name, if creating a job-based object reference
+    :type field: string
 
     Creates a DXLink (a dict formatted as a symbolic DNAnexus object
     reference) to the specified object.  Returns *object_id* if it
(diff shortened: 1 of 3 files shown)


src/python/dxpy/bindings/dxdataobject_functions.py
```diff
@@ -34,26 +34,28 @@
...
     :type project_id: string
+    :param field: A field name, if creating a job-based object reference
+    :type field: string
 
     Creates a DXLink (a dict formatted as a symbolic DNAnexus object
     reference) to the specified object.  Returns *object_id* if it
     appears to be a DXLink already.
     '''
     if isinstance(object_id, DXDataObject):
         object_id = object_id.get_id()
```
orodeh: This part is confusing, the rest of the method is fairly readable. Why is dxlink returning a different type based on the content? It would be much more readable if dxlink always returned a dict (for example). 

Ways to address this are:
1) Add some documentation, explaining what is going on, and why. 
2) Splitting the code between the case where a bare bones object_id is returned, and where a dict is returned.

src/python/dxpy/bindings/dxdataobject_functions.py
```diff
@@ -80,28 +82,34 @@ def get_dxlink_ids(link):
...
-    if isinstance(link['$dnanexus_link'], dict):
+    if not is_dxlink(link):
+        raise DXError('Invalid link: %r' % link)
+    if isinstance(link['$dnanexus_link'], basestring):
+        return link['$dnanexus_link'], None
+    elif 'id' in link['$dnanexus_link']:
         return link['$dnanexus_link']['id'], link['$dnanexus_link'].get('project')
     else:
-        return link['$dnanexus_link'], None
+        return link['$dnanexus_link']['job'], link['$dnanexus_link']['field']
```
orodeh: Can you give an example of a dnanexus link that matches this case? (inlined as documentation)

src/python/dxpy/bindings/dxdataobject_functions.py
```diff
@@ -80,28 +82,34 @@ def get_dxlink_ids(link):
...
-        if isinstance(link['$dnanexus_link'], basestring):
-            link = link['$dnanexus_link']
-        else:
-            link = link['$dnanexus_link']['id']
-    class_name, _id = link.split("-", 1)
-    class_name = 'DX'+class_name.capitalize()
-    if class_name == 'DXGlobalworkflow':
-        class_name = 'DXGlobalWorkflow'
+        link = get_dxlink_ids(link)[0]
+    class_name = 'DX' + link.split("-", 1)[0].capitalize()
```
orodeh: If I understand the code right, the following comment would be helpful:
// Now we have an object-id, for example: "file-xxxx", "record-xxxx", ...

src/python/dxpy/bindings/dxdataobject_functions.py
```diff
@@ -80,28 +82,34 @@ def get_dxlink_ids(link):
...
+    if class_name not in all_bindings:
+        class_name = {
+            'DXGtable': 'DXGTable',
```
orodeh: We do not support GTables any more.

jdidion:
@orodeh please re-review. Thanks

PR #509, opened 2019-03-26
DEVEX 923 - Approved tools CLI

geetduggal:


geetduggal:
A version of this passed the nucleus MOAT tests and has undergone a bit of testing in a mini environment.  I ended up relying on the length of the ancestorSignatures list as that was the most reliable thing available to be at the point of the processing occurring at the base execution handler.

src/python/test/test_dxpy.py
```diff
@@ -197,6 +198,7 @@ def test_update_describe(self):
         self.assertEqual(desc["restricted"], True)
         self.assertEqual(desc["downloadRestricted"], True)
         self.assertEqual(desc["description"], "new description")
+        self.assertEqual(desc["allowedExecutables"][0], "applet-FGkbg180f5v2p1226jFqGJGy")
```
commandlinegirl: Is it an arbitrary ID? To emphasize it I'd write something like: "applet-123456..." 

AFAIR we check on the server side if the hash is correctly formatted but not if the executable exists? That is fine.

Minor: please add a test somewhere here for removing/nulling allowed executables list.

src/python/dxpy/scripts/dx.py
```diff
@@ -4592,6 +4592,7 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
 parser_update_project.add_argument('--containsPHI', choices=["true"],
                                    help="Flag to tell if project contains PHI")
 parser_update_project.add_argument('--bill-to', help="Update the user or org ID of the billing account", type=str)
+parser_update_project.add_argument('--allowed-executables', help='Executable ID(s) this project is allowed to run.  This operation overrides any existing list of executables.', type=str, nargs="+")
```
commandlinegirl: Maybe add here a note: "Passing null removes the list" or similar. Empty list means no executables are allowed to run?

PR #510, opened 2019-04-02
Hardcode python2.7 in environment

yesimon:
Since dx-toolkit is distributed as python2.7 only right now, hardcode it to 2.7 like in the other environment files.

### DIFF
--- a/environment
+++ b/environment
@@ -45,8 +45,7 @@ fi
 
 export DNANEXUS_HOME="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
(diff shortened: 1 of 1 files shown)


orodeh:
We are, albeit slowly, working towards python 3. The code has already been converted to work for 2.7 and 3.x. Currently, it is compiled and distributed only for 2.7, but you can download and compile it for version 3.x. 

To do this, do a `make python` from the top level directory, where python is your favorite python flavor.

yesimon:
Can you please update the binary distribution? It only contains 2.7 and breaks on any machine where `python` is python3. (at least on linux-64)

orodeh:
That is a very reasonable request. Can you please a new issue describing it? We want to move forward to python 3.

PR #511, opened 2019-04-02
SCI-1323, SCI-1324: Add sugar transfer and processing modules

jdidion:
This new branch is based on SCI-1321, rather than master.

* Add quote and makedirs to dxpy.compat
* Most of the changes suggested in #508 are implemented. Some exceptions:
    * The tmpfile contextmanager cannot be replaced by tempfile.TemporaryFile. It is needed in situations where the tempfile needs to be written from python and then closed and read from the subprocess.
    * I am still using dx commands for file upload and download when compression/decompression is involved. The python implementations of compression libraries are at least 2x slower than the system commands.
    * Regarding use of bash shell for subprocesses, see comment here: https://github.com/dnanexus/dx-toolkit/pull/508#discussion_r270569217
    * Regarding use of *args for passing commands, see comment here: https://github.com/dnanexus/dx-toolkit/pull/508#discussion_r269405402
    * For --buffer-size, I get the max value from project.describe(), but I also limit it to a max of 2GB - 1, due to a macOS bug that prevents read() from taking any value larger than that.

### DIFF
--- a/src/python/dxpy/sugar/processing.py
+++ b/src/python/dxpy/sugar/processing.py
@@ -69,13 +69,12 @@ class Processes:
         """
         The max return code of all the commands in the chain.
         """
+        if not self.was_run:
+            raise RuntimeError("Must call run() before returncode can be requested")
         if self._returncode is None:
-            returncodes = [
-                proc.returncode for proc in self._processes if proc.poll() is not None
-            ]
-            if returncodes:
-                # TODO: Is setting to max the right behavior?
-                self._returncode = max(returncodes)
+            # Set `self._returncode` to the returncode of the last process,
+            # to mimic behavior of `set -o pipefail`.
+            self._returncode = self._processes[-1].poll()
         return self._returncode
 
     def _init_stdout(self):
--- a/src/python/dxpy/sugar/processing.py
+++ b/src/python/dxpy/sugar/processing.py
@@ -72,9 +72,21 @@ class Processes:
         if not self.was_run:
             raise RuntimeError("Must call run() before returncode can be requested")
         if self._returncode is None:
-            # Set `self._returncode` to the returncode of the last process,
-            # to mimic behavior of `set -o pipefail`.
-            self._returncode = self._processes[-1].poll()
+            for proc in reversed(self._processes):
+                rc = proc.poll()
+                if rc is None:
+                    # The pipeline has not finished running
+                    break
+                if rc > 0:
+                    # A command finished with an error
+                    # Set `self._returncode` to the returncode of the last process
+                    # to finish with an error, to mimic behavior of `set -o pipefail`.
+                    self._returncode = rc
+                    break
+            else:
+                # None of the process return codes were None or non-zero, thus
+                # all must have been zero.
+                self._returncode = 0
         return self._returncode
 
     def _init_stdout(self):
--- a/src/python/dxpy/sugar/processing.py
+++ b/src/python/dxpy/sugar/processing.py
@@ -1,5 +1,6 @@
 from __future__ import print_function, unicode_literals, division, absolute_import
 import copy
+import errno
 import logging
 import shlex
 import subprocess
@@ -60,33 +61,31 @@ class Processes:
         self._processes = None
         self._output_handle = None
         self._returncode = None
-        # These will hold
-        self.out = None
-        self.err = None
+        self._out = None
+        self._err = None
 
     @property
     def returncode(self):
         """
-        The max return code of all the commands in the chain.
+        The return code of the last process to finish with an error, or 0. Is None
+        if the last process hasn't finished running.
         """
         if not self.was_run:
             raise RuntimeError("Must call run() before returncode can be requested")
         if self._returncode is None:
-            for proc in reversed(self._processes):
-                rc = proc.poll()
-                if rc is None:
-                    # The pipeline has not finished running
-                    break
-                if rc > 0:
-                    # A command finished with an error
-                    # Set `self._returncode` to the returncode of the last process
-                    # to finish with an error, to mimic behavior of `set -o pipefail`.
-                    self._returncode = rc
-                    break
-            else:
-                # None of the process return codes were None or non-zero, thus
-                # all must have been zero.
-                self._returncode = 0
+            # Set `self._returncode` to the returncode of the last process
+            # to finish with an error, to mimic behavior of `set -o pipefail`.
+            # Note that if any process other than the last has a return code of
+            # None, we ignore it.
+            self._returncode = self._processes[-1].poll()
+            if self._returncode == 0 and len(self._processes) > 1:
+                # The last process finished running without error, but check all
+                # the other processes for an error.
+                for proc in reversed(self._processes[-2::-1]):
+                    rc = proc.poll()
+                    if rc:
+                        self._returncode = rc
+                        break
         return self._returncode
 
     def _init_stdout(self):
@@ -133,7 +132,35 @@ class Processes:
         return open(tempfile.mkstemp()[1], "w")
 
     @property
-    def stdout(self):
+    def output(self):
+        """
+        The contents of the `stdout` stream of the last process in the chain.
+        Only available if `self.closed is True` and `self._stdout_type in
+        {_PIPE, _BUFFER}`.
+        """
+        if not (self.closed and self._stdout_type in {_PIPE, _BUFFER}):
+            raise RuntimeError(
+                "'output' is only available for closed processes using a pipe or "
+                "buffer for stdout."
+            )
+        return self._out
+
+    @property
+    def error(self):
+        """
+        The contents of the `stderr` stream of the last process in the chain.
+        Only available if `self.closed is True` and `self._stderr_type in
+        {_PIPE, _BUFFER}`.
+        """
+        if not (self.closed and self._stderr_type in {_PIPE, _BUFFER}):
+            raise RuntimeError(
+                "'error' is only available for closed processes using a pipe or "
+                "buffer for stderr."
+            )
+        return self._err
+
+    @property
+    def stdout_stream(self):
         """
         The `stdout` stream of the last process in the chain.
         """
@@ -142,7 +169,7 @@ class Processes:
         return self._stdout or self._processes[-1].stdout
 
     @property
-    def stderr(self):
+    def stderr_stream(self):
         """
         The `stderr` stream of the last process in the chain.
         """
@@ -170,7 +197,7 @@ class Processes:
         if self.capture_stderr:
             stderr.extend(self._stderr_buffers)
         if self._stderr_type in (_BUFFER, _PIPE):
-            stderr.append(self.err)
+            stderr.append(self.error)
         return stderr
 
     @property
@@ -249,6 +276,11 @@ class Processes:
             raise_on_error: Whether to raise a :class:`subprocess.CalledProcessError`
                 if the returncode was not 0.
         """
+        if not self.was_run:
+            raise RuntimeError("Cannot call block() until after calling run()")
+        if self.closed:
+            raise RuntimeError("Cannot call block() after calling close()")
+
         last_proc = self._processes[-1]
         try:
             out, err = (
@@ -256,15 +288,15 @@ class Processes:
                 for std in last_proc.communicate()
             )
             if self._stdout_type == _PIPE:
-                self.out = out
+                self._out = out
             if self._stderr_type == _PIPE:
-                self.err = err
+                self._err = err
         except ValueError:
             LOG.error("Error reading from stdout/stderr")
             pass
 
         if close:
-            self.close()
+            self._close_and_set_std()
 
         if raise_on_error:
             self.raise_if_error()
@@ -277,16 +309,18 @@ class Processes:
         Returns:
             True if processes were killed, else False.
         """
-        if self.was_run:
-            if not self.done:
-                for i, proc in enumerate(self._processes):
-                    try:
-                        proc.kill()
-                    except:
-                        LOG.exception(
-                            "Error killing running process command %s", self.cmds[i]
-                        )
-                return True
+        if self.was_run and not self.done:
+            for i, proc in enumerate(self._processes):
+                try:
+                    proc.kill()
+                except OSError as oserr:
+                    if oserr.errno == errno.ESRCH:
+                        # Ignore - process has already died
+                        pass
+                    LOG.exception(
+                        "Error killing running process command %s", self.cmds[i]
+                    )
+            return True
 
         return False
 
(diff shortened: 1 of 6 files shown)


src/python/dxpy/sugar/processing.py
```diff
@@ -0,0 +1,508 @@
...
+    def returncode(self):
+        """
+        The max return code of all the commands in the chain.
+        """
+        if self._returncode is None:
+            returncodes = [
+                proc.returncode for proc in self._processes if proc.poll() is not None
+            ]
+            if returncodes:
+                # TODO: Is setting to max the right behavior?
```
Damien-Black: I think returning the **first** failed command's error code is the correct thing to do. Mimicing the behaviour of `set -o pipefail`.

Devil's advocating a bit. If we wish to depart from `set -o pipefail` behaviour in this aspect then we should return the **first** failed error code In practice when piping, later failures are usually due to earliers ones.

```
$ dx describe file-xxxx --json | grep "name"
# DXCLIError: No matches found for file-xxxx
```

The return code in the normal `set -o pipefail` case is from the failed `grep`, 1. The error occured with the `dx` command, exit code 3.


Again, to be clear, I vote **return the error code of the last command** to keep with expectations set by `set -o pipefail`. Still show the failure message of the real failed command.

src/python/dxpy/sugar/processing.py
```diff
@@ -0,0 +1,508 @@
...
+        if value is None:
+            return None, _PIPE, subprocess.PIPE
+
+        std_type = _OTHER
+        if value is False:
+            value = None
+        elif value is True:
+            value = Processes._create_temp_outfile()
+            std_type = _BUFFER
+        elif isinstance(value, compat.basestring):
```
Damien-Black: I believe this is the correct check to perform here. @orodeh any thoughts/comments here? My thought's below:

I need to check if `dx-toolkit` documentation is outdated here. [In the docs](https://github.com/dnanexus/dx-toolkit/tree/a86d952d5142dc01a18d64f50d5049a0f7e39dbe/src/python#python-version-compatibility) we recommend

```python
from dxpy.compat import str
```

Then using `str` from that point onward. 

In this scenario a user is consuming our Class, so:
* in their python2 code they'll send over ASCII `str` (assuming they forget the __future__ imports)
* which would fail a `ininstance(value, str)` in our code since `str` == `unicode`

That's my reasoning why this is the preferred check here. Again @orodeh or @jdidion lmk if this is off.

jdidion: We need to handle both str and unicode in py2, so I think basestring is the right class to check against.

src/python/dxpy/sugar/processing.py
jdidion: Done

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+        This method is needed distinct from :class:`tempfile.TemporaryFile` in the
+        case where python needs to write to the file and then a subprocess needs
+        to read from the file.
+    """
+    path = tempfile.mkstemp(*args, **kwargs)[1]
+    try:
+        yield path
+    finally:
+        if os.path.exists(path):
+            os.remove(path)
```
msimbirsky: Note: This should be used from the context manager module, I am assuming it is only here temporarily and will be removed once the context managers module is merged in. In that case, I think it would be better to put it in a separate temporary context manager module in this PR and import it from there.

src/python/dxpy/sugar/processing.py
```diff
@@ -0,0 +1,519 @@
...
+        last_proc = self._processes[-1]
+        try:
+            out, err = (
+                "" if std is None else std.strip()
+                for std in last_proc.communicate()
+            )
+            if self._stdout_type == _PIPE:
+                self.out = out
+            if self._stderr_type == _PIPE:
+                self.err = err
```
msimbirsky: I don't like the use of .out and .err here. This is mostly aesthetic but the command call "run_cmd(cmd).out" seems pretty unintuitive to me. Perhaps we can add some helper functions, such as "get_stdout()" and "get_stderr()" and have those return .out and .err?

msimbirsky: This would make the command: `run_cmd(cmd).get_stdout()`

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+        kwargs["name"] = remote_filename
+    if not remote_folder:
+        remote_folder = "/"
+    elif not remote_folder.startswith("/"):
+        remote_folder = "/{}".format(remote_folder)
+    kwargs["folder"] = remote_folder
+    if project:
+        kwargs["project"] = project
+    LOG.info("Uploading file %s to %s", filename, remote_filename or filename)
+    handler = dxpy.upload_local_file(filename, **kwargs)
```
msimbirsky: As much as I like `remote_filename` and `remote_folder` as being more descriptive, I think the passed-in kwargs should be consistent with other dxpy functions, so should just be `name` and `folder` for the remote file/folder locations. To avoid confusion you could replace `filename` with `local_path`

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+        remote_filename (str): Optional, remote filename to upload file to.
+        remote_folder (str): Optional, remote folder to upload file to.
+        compression_level (int): Level of compression between 1 and 9 to compress
+            file to. Specify 1 for gzip --fast and 9 for gzip --best. If not
+            specified, --fast is assumed.
+        return_handler (bool): Whether to return a DXFile handler.
+        project (str): The project ID to upload to, if not the currently selected
+            project.
+        wait_on_close (bool): Whether to block until the file has closed.
+        method (str): Compression method; one of 'gz', 'bz2'.
```
msimbirsky: Method implies executable to me, either "gzip" or "bgzip". Perhaps replace `method` with `extension` and then use the appropriate exe from there.

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+        ]
+        if wait_on_close:
+            upload_cmd.append("--wait")
+        upload_cmd.append("-")
+        file_id = proc.chain_cmds([zip_cmd, upload_cmd], shell=True).out
+
+    return _wrap_file_id(file_id, return_handler)
+
+
+def archive_and_upload_files(
```
msimbirsky: `tar_and_upload` is more readable to me than `archive_and_upload`. Tarring a directory is more frequently used to refer to this operation in common context.

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+    max_part_size=None
+):
+    """
+    Archive and upload one or more files.
+
+    Shorthand for running 'tar', 'gzip', and 'dx upload' using subprocess on a
+    list of local files/directories.
+
+    Args:
+        filenames (str or list of str): = Local filenames or directories.
```
msimbirsky: Change "filenames" to `paths` to indicate the argument passed in can be a file or a directory

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+        remote_prefix (str) = Name to give to output tar archive. Must be provided
+            unless `filenames` is of length 1, in which case the prefix will be
+            the same as that of the single filename.
+        remote_folder (str): Optional, remote folder to upload file to.
+        compression_level (int) = Level of compression between 1 and 9 to
+            compress tar to. Specify 1 for gzip --fast and 9 for gzip --best.
+            If not specified, --fast is assumed. If None, no compression is
+            performed, i.e. the output is a .tar file.
+        chdir (bool): If a path, change to this directory; if None and filenames is
```
msimbirsky: The usage of `chdir` as a boolean is confusing here. Probably would be better to have the actual path to cd to be specified, much like passing a dir to the `-C` flag. This also allows for other paths to be specified under the `tar -C ` command.

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+
+    Args:
+        filenames (str or list of str): = Local filenames or directories.
+        remote_prefix (str) = Name to give to output tar archive. Must be provided
+            unless `filenames` is of length 1, in which case the prefix will be
+            the same as that of the single filename.
+        remote_folder (str): Optional, remote folder to upload file to.
+        compression_level (int) = Level of compression between 1 and 9 to
+            compress tar to. Specify 1 for gzip --fast and 9 for gzip --best.
+            If not specified, --fast is assumed. If None, no compression is
```
msimbirsky: A user could also naively specify `0` as a compression level for `None`.

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+        remote_folder = "/"
+    elif not remote_folder.startswith("/"):
+        remote_folder = "/{}".format(remote_folder)
+
+    if compression_level is not None and (
+        compression_level < 1 or compression_level > 9
+    ):
+        raise ValueError("Compression level must be between 1 and 9")
+
+    if compression_level:
```
msimbirsky: This should be an `elif` statement combined with the one above

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+    remote_file = _as_dxfile(remote_file, project)
+
+    if local_filename is None:
+        local_filename = SPECIAL_RE.sub("", remote_file.name)
+    if output_dir:
+        local_filename = os.path.join(output_dir, local_filename)
+
+    LOG.info("Downloading file %s to %s", remote_file.get_id(), local_filename)
+
+    cmd = ["dx", "download", remote_file.get_id(), "-o", local_filename]
```
msimbirsky: For simple download it would be better to run `dxpy.download_dxfile()`

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+        if not os.path.exists(local_filename):
+            raise ValueError(
+                "Expected file {} does not exist after untarring file {}",
+                local_filename, remote_file.get_id()
+            )
+        return [local_filename]
+    else:
+        # Otherwise return the list of files that were unpacked from the tar file
+        with open(file_list_filename, "rt") as inp:
+            return [os.path.join(output_dir, filename.rstrip()) for filename in inp]
```
msimbirsky: I like this additional return of all output files untarred, very useful

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+    if block:
+        LOG.info(
+            "Completed downloading file %s to %s", remote_file.get_id(), local_filename
+        )
+        return local_filename
+    else:
+        return result
+
+
+def _as_dxfile(fileobj, project=None):
```
msimbirsky: It seems to me this function duplicates dxpy.get_handler()

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+    # check that compression level is valid
+    if compression_level < 1 or compression_level > 9:
+        raise ValueError("Compression level must be between 1 and 9")
+
+    if not max_part_size or max_part_size <= 0:
+        max_part_size = _get_max_part_size(project)
+
+    if filename.endswith(ext):
+        cmd = [
+            "dx", "upload", "--brief", "--buffer-size", str(max_part_size),
```
Damien-Black: Not sure if I mentioned this in the previous PR.

Underneath, we try and perform uploads in parallel with each thread using the `buffer-size` that was set (code [permalink](https://github.com/dnanexus/dx-toolkit/blob/a86d952d5142dc01a18d64f50d5049a0f7e39dbe/src/python/dxpy/bindings/dxfile.py#L513)). We submit these to task to worker pools. Which AFAIK is unaware of the current memory on the worker. What can happen is:

```
num_active_threads * buffer-size > memory availble on machine
```

While your change is less prone failing to upload in `partLimits` I think you're more prone to making the worker going OOM.

My initial idea for a fix is:
1. Make the initial `buffer-size` calculation _slightly_ worker aware (i.e never choose a buffer size greater than the memory on the worker).
2. Modify writes to be aware of the pool and current memory on the machine.

Maybe there's a simpler fix.

src/python/dxpy/sugar/transfers.py
jdidion: Depends on whether you decide to include it in context.py or if it should remain private to this module. If you want me to keep it private, I'll prefix the function name with "_".

msimbirsky: I think a better implementation for this would be to use `tempfile.TemporaryNamedFile`, I'll add a comment below with what I mean.

src/python/dxpy/sugar/transfers.py
```diff
@@ -0,0 +1,952 @@
...
+    remote_path = "{}/{}.{}".format(remote_folder, remote_prefix, ext)
+    if project:
+        remote_path = "{}:{}".format(project, remote_path)
+
+    if not max_part_size or max_part_size <= 0:
+        max_part_size = _get_max_part_size(project)
+
+    with tmpfile() as names_file:
+        with open(names_file, "wt") as out:
+            out.write("\n".join(filenames))
```
msimbirsky: Replace this with:
```
with tempfile.NamedTemporaryFile(delete=False) as temp:
            with open(temp.name, 'wt') as out:
                     ...
           ...
           tar_cmd.extend(["cvf", "-", "--files-from", temp.name])
```

src/python/dxpy/sugar/processing.py
jdidion: I cleaned this up a bit, and changed to providing these values via the 'output' and 'error' properties rather than allowing direct access to the attributes.

src/python/dxpy/sugar/transfers.py
jdidion: Done

src/python/dxpy/sugar/transfers.py
jdidion: Changed it to 'compression_type'

src/python/dxpy/sugar/transfers.py
jdidion: Done

src/python/dxpy/sugar/transfers.py
jdidion: Done

src/python/dxpy/sugar/transfers.py
jdidion: Done

src/python/dxpy/sugar/transfers.py
jdidion: Oops - missed that. Changed it to call dxpy.get_handler unless fileobj is already a DXFile

src/python/dxpy/sugar/processing.py
```diff
@@ -0,0 +1,508 @@
...
+    def raise_if_error(self):
+        """
+        Raise an exception if the processes are finished running and the
+        return code is not 0.
+
+        Raises:
+            CalledProcessError
+        """
+        if self.done and not self.ok:
+            msg = "stderr from executed commands: {}".format(
```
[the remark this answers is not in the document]
jdidion: Done

src/python/dxpy/sugar/transfers.py
jdidion: Changed to using download_dxfile when block is True.

src/python/dxpy/sugar/transfers.py
jdidion: Cool. I used psutil to get the available memory, and have Uploader._submit pass in its _max_workers attribute to the upload function, and then use those together to limit max_part_size to at most available_memory / max_workers.

src/python/dxpy/sugar/transfers.py
jdidion: But then the file doesn't get deleted after running the tar command.

Damien-Black:
@jdidion I think I'm good with my part of the review. You can wrap up the PR when @msimbirsky gives the okay.

geetduggal:
@jdidion  Sorry I missed this PR!  Will try to keep better tabs in the future.

PR #512, opened 2019-04-03
DEVEX-1005 Decrease libcurl timeout

commandlinegirl:
A libcurl's option [CURLOPT_TIMEOUT](https://curl.haxx.se/libcurl/c/CURLOPT_TIMEOUT.html) was set to infinity for the cpp bindings http requests. This was causing UA to hang instead of timing out and retrying sending another request to the API server. This setting doesn't affect the libcurl session used for uploading chunk data [here](https://github.com/dnanexus/dx-toolkit/blob/0b85ce034e48f3a5c6f6e993b2f681af32d0848a/src/ua/chunk.cpp#L360). 

I set it to 15 minutes by default and 30 for PUT (potentially used by the binding's upload) though from a quick search I cannot find where HTTP_PUT might be used: https://github.com/dnanexus/dx-toolkit/search?q=HTTP_PUT&unscoped_q=HTTP_PUT

### DIFF
--- a/src/cpp/SimpleHttpLib/SimpleHttp.cpp
+++ b/src/cpp/SimpleHttpLib/SimpleHttp.cpp
@@ -276,9 +276,6 @@ namespace dx {
       // Make a copy of reqData, because read_callback (see HTTP_PUT case below) will modify it
       reqData_struct reqData_temp;
 
-      std::cout << "METHOD USED: " << method << " \n";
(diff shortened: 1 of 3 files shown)


commandlinegirl:
A few more comments in the JIRA ticket.

commandlinegirl:
Proper DEVEX ticket for this commit is DEVEX-1105

PR #514, opened 2019-04-03
DEVEX-1103 -Timeout and retry upload of file part

geetduggal:
This is a quick modification to allow for the timeout and retry of a file part upload.  It is still under testing.

### DIFF
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -39,6 +39,7 @@ from ..compat import open, USING_PYTHON2
 from ..exceptions import DXFileError, DXPartLengthMismatchError, DXChecksumMismatchError, DXIncompleteReadsError, err_exit
 from ..utils import response_iterator
 import subprocess
+from . import timeout
 
 def open_dxfile(dxid, project=None, mode=None, read_buffer_size=dxfile.DEFAULT_BUFFER_SIZE):
     '''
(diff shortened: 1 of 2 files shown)


src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -527,10 +528,26 @@ def report_progress(handler, num_bytes):
...
+            handler.write(buf,
+                        report_progress_fn=report_progress if show_progress else None,
+                        multithread=multithread,
+                        **remaining_kwargs)
+
+        for i in xrange(MAX_RETRIES_PART_UPLOAD):
+            try:
+                upload_part()
+                break
+            except Exception as e:
```
geetduggal: Catch the specific `TimeoutException` here instead of generic `Exception`

src/python/dxpy/bindings/timeout.py
```diff
@@ -0,0 +1,27 @@
...
+
+from functools import wraps
+import errno
+import os
+import signal
+
+class TimeoutError(Exception):
+    pass
+
+def timeout(seconds=10, error_message=os.strerror(errno.ETIME)):
```
commandlinegirl: According to https://docs.python.org/2/library/signal.html#signal.signal SIGALARM is not available for Windows:

```
On Windows, signal() can only be called with SIGABRT, SIGFPE, SIGILL, SIGINT, SIGSEGV,
SIGTERM, or SIGBREAK. A ValueError will be raised in any other case 
```

so we could catch it and simply return the function there (without the timeout handler).

According to https://stackoverflow.com/questions/8420422/python-windows-equivalent-of-sigalrm, it can be made work if python is installed with Cygwin on Windows, I’m not sure how we use/install Python there.

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -527,10 +528,26 @@ def report_progress(handler, num_bytes):
...
+                if i == MAX_RETRIES_PART_UPLOAD - 1:
+                    raise Exception("Tried too many times to upload file part.  There may be some issue with the network connectivity.")
+                else:
+                    print("Timeout exceeded uploading file part. Retrying upload.")
```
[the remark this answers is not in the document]
geetduggal: Done, thanks!

src/python/dxpy/bindings/timeout.py
geetduggal: Done

PR #515, opened 2019-04-04
 DEVEX-1106 Handling file arrays correctly

orodeh:


src/python/dxpy/utils/batch_utils.py
```diff
@@ -148,13 +151,14 @@ def batch_launch_args(executable, input_json, batch_tsv_file):
...
         if col_name == BATCH_ID:
             batch_index = i
-        elif (col_name in input_classes and
-              input_classes[col_name] != 'file'):
-            index_2_column[i] = col_name
-        elif (col_name in input_classes and
-              input_classes[col_name] == 'file'):
+        if col_name not in input_classes:
+            continue
+        if (input_classes[col_name] == 'file' or
```
commandlinegirl: Minor: this could be: 
```
if input_classes[col_name] in ['file', 'array:file']:
```

src/python/dxpy/utils/batch_utils.py
```diff
@@ -148,13 +151,14 @@ def batch_launch_args(executable, input_json, batch_tsv_file):
     for i, col_name in enumerate(header_line):
         if col_name == BATCH_ID:
             batch_index = i
-        elif (col_name in input_classes and
-              input_classes[col_name] != 'file'):
-            index_2_column[i] = col_name
-        elif (col_name in input_classes and
-              input_classes[col_name] == 'file'):
+        if col_name not in input_classes:
```
commandlinegirl: If you get a chance, could you add an example (in the comment at the top or, better yet, in `dev_wiki`) for generating and running batch file with `array:file` type?

orodeh: I added a test that makes sure writing square brackets is interpreted correctly arrays of files. This is the new test in `src/python/test/test_batch.py`.

I am not sure batch generation is able to create this for you. In any event, I didn't try it or fix it.

PR #518, opened 2019-04-12
TIP-1021: support for passing httpcontext to execute operation.

codecheetos:
Allows derived implementations to override with custom timeouts etc.

Apollo HTTP client for apiserver uses http-commons PoolingHttpClientConnectionManager which allows client to throw ConnectionTimeoutException if timeout exceeds. The timeout is passed by overridden method in derived instance of DXHttpRequest class.

See https://github.com/dnanexus/nucleus/pull/2773


arunbhat:
@commandlinegirl  please review

PR #519, opened 2019-04-24
DEVEX-1116 Reduce the number of API calls for dx download

commandlinegirl:
This reduces the # of calls from 5 (below) to 2 for one file download (3 if not counting this [PR](https://github.com/dnanexus/dx-toolkit/pull/503)).

- `findDataObjects()` already returns descriptions of the objects to be downloaded. The field "parts" needs to be included in the descriptions for file objects downstream (for which we were making another `describe` call). I now request "parts" to be included in file describes when `resolve_existing_path()` is called from `download()` functions.
- folder resolution is skipped [here](https://github.com/dnanexus/dx-toolkit/pull/503)
- I also skip a check on whether the file is actually in the project, if the project was explicitly named (it is skipped when `args.lightweight` is used)

In this example the calls  1, 2, 3 will not be made with this update:

image

Tests passed.

### DIFF
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -692,5 +692,10 @@ def download_folder(project, destdir, folder="/", overwrite=False, chunksize=dxf
                      ("" if remote_file['describe']['folder'] == "/" else remote_file['describe']['folder']),
                      remote_file['describe']['name'],
                      local_filename)
-        download_dxfile(remote_file['describe']['id'], local_filename, chunksize=chunksize, project=project,
-                        show_progress=show_progress, describe_output=remote_file['describe'], **kwargs)
+        download_dxfile(remote_file['describe']['id'],
+                        local_filename,
+                        chunksize=chunksize,
+                        project=project,
+                        show_progress=show_progress,
+                        describe_output=remote_file['describe'],
+                        **kwargs)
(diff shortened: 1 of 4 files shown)


src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -95,7 +95,7 @@ def new_dxfile(mode=None, write_buffer_size=dxfile.DEFAULT_BUFFER_SIZE, expected
 
 
 def download_dxfile(dxid, filename, chunksize=dxfile.DEFAULT_BUFFER_SIZE, append=False, show_progress=False,
-                    project=None, **kwargs):
+                    project=None, describe_output=None, **kwargs):
```
geetduggal: High level/clarification question.  For the new `describe_output` field in `download_dxfile`, for what code paths (if any) is the default value of None passed?

commandlinegirl: Whenever `download_dxfile()` is called (since it's a part of our API) without the `describe_output` parameter.

Internally it's used 
* in `dx get` when resources are downloaded [here](https://github.com/dnanexus/dx-toolkit/blob/98b7f0ae600147cf7dcae7ab710ef38f07458e9a/src/python/dxpy/utils/executable_unbuilder.py#L183) 
* in the [`download_all_inputs`](https://github.com/dnanexus/dx-toolkit/blob/ef5d9a3dad811dd6dc7bfa6f34eb9a4a6669985d/src/python/dxpy/bindings/download_all_inputs.py#L49) script

PR #523, opened 2019-05-16
DEVEX-1152 cryptography 2.3

kpjensen:


src/python/dxpy/__init__.py
```diff
@@ -185,6 +185,16 @@ def _match_hostname(cert, hostname):
 
 configure_urllib3()
 
+def configure_cryptography():
+    # Import inside of function since 'warnings' is imported globally in downloads.py and dxfile_functions.py
+    import cryptography
+    import warnings
+    from cryptography import utils
+    warnings.simplefilter('ignore', cryptography.utils.DeprecatedIn23)
```
commandlinegirl: Neat trick --> `simplefilter`

It looks like the function definition is different in Py2.7 and Py3.6:
https://docs.python.org/2.7/library/warnings.html#warnings.simplefilter
https://docs.python.org/3.6/library/warnings.html#warnings.simplefilter
Did you have a change to run it thru 3.6?

kpjensen: The function should only be called if Python version is <= 2.7.6, but I did check and the `simplefilter()` call succeeds in Py3, but `cryptography.utils.DeprecatedIn23` no longer exists in later cryptography versions. I added another check for its existence before filtering.

PR #525, opened 2019-05-22
Merge dxjava fork

orodeh:
This merges back changes made to the dxWDL fork of dxjava. There are three kinds of changes:
1. Differences between the scala/java compiler and the normal java compiler
2. Addition of support for proxies
3. Allowing disabling retries. This helps testing.

### DIFF
--- a/src/java/src/main/java/com/dnanexus/DXEnvironment.java
+++ b/src/java/src/main/java/com/dnanexus/DXEnvironment.java
@@ -604,7 +604,7 @@ public class DXEnvironment {
      *    https://localhost:3128
      *    https://dnanexus:welcome@localhost:3128
      */
-    public ProxyDesc parseProxyDefinition(String proxyDfn, String proxyMethod, String proxyDomain) {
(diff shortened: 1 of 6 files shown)


src/java/src/main/java/com/dnanexus/DXEnvironment.java
```diff
@@ -511,6 +597,56 @@ public int getSocketTimeout() {
...
     }
 
+    /**
+     * Parse the proxy definition. This could take one of several forms:
+     *    localhost:3128
+     *    http://localhost:3128
+     *    https://localhost:3128
+     *    https://dnanexus:welcome@localhost:3128
+     */
+    public ProxyDesc parseProxyDefinition(String proxyDfn, String proxyMethod, String proxyDomain) {
```
commandlinegirl: Does this have to be public?

orodeh: Correct, it doesn't have to be. I'll make it private.

PR #530, opened 2019-06-18
DEVEX-1205 xattr properties on dx-upload-all-outputs

kpjensen:
Add a new parameter `--xattr-properties` to `dx-upload-all-outputs`. If this argument is passed, read filesystem metadata and set `k,v` properties for each file uploaded, stripping the `user.` prefix from the key. Some additional checks may be required to make sure only user attributes are set as properties.

Add `xattr` as a dependency Linux builds of dx-toolkit, other distributions do not require it since this script only runs inside the DNAnexus execution environment.

### DIFF
--- a/src/python/scripts/dx-upload-all-outputs
+++ b/src/python/scripts/dx-upload-all-outputs
@@ -291,7 +291,7 @@ def create_entry(subdir_desc, local_path, target_path, basename):
             'fname': basename,
             'dxlink': None}
 
-def return_xatrr_as_properties(file):
+def return_xattr_as_properties(file):
     properties = {}
     for attribute in xattr.listxattr(file):
         if attribute.startswith('user.'):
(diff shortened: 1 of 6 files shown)


src/python/scripts/dx-upload-all-outputs
```diff
@@ -288,35 +291,42 @@ def create_entry(subdir_desc, local_path, target_path, basename):
             'fname': basename,
             'dxlink': None}
 
-def upload_one_file(entry, wait_on_close):
+def return_xatrr_as_properties(file):
+    properties = {}
+    for attribute in xattr.listxattr(file):
+        properties[attribute.lstrip('user.')] = xattr.getxattr(file, attribute)
```
asimenos: Why not add: `if attribute.startswith('user.')`

PR #533, opened 2019-06-24
Sci 1328 available memory

jdidion:
Renamed get_memory to available_memory.

I'm not sure where to put this function, so right now I'm putting it in sugar/__init__.py.

One thought is to have an environment.py or worker.py module with functions related to the worker runtime environment. But right now available_memory is the only such function.


src/python/dxpy/sugar/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+from dxpy.sugar.transfers import (
+    Uploader,
+    Downloader,
+    upload_file,
+    tar_and_upload_files,
+    download_file,
+)
+
+
+MEM_RE = re.compile(r"^MemTotal:[\s]*([0-9]*) kB")
```
Damien-Black: From the definitions and naming in `available_memory` I get the feeling you don't want the memory on the worker, but you want the free (unused) memory on the worker.

```suggestion
MEM_RE = re.compile(r"^MemFree:[\s]*([0-9]*) kB")
```

I'll add that I don't see this variable being used anywhere else. I think it's better to just write the regular expression in the `available_memory` function.

src/python/dxpy/sugar/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+MEM_CONVERSIONS = {
```
Damien-Black: TIL that `/proc/meminfo` stores memory in base 2 not base 10. The `kB` output misled me.

That being said, do we want to to use the same suffixes outputted when you read the file (`kB`). Or even do you want to coerce to `KiB`.

I'm still of the understanding that `KB` implies base-10. Either way the change would be minor and only impact readability of the code. Thoughts on the correct approach here?

src/python/dxpy/sugar/__init__.py
jdidion: I have a habit of defining regular expressions as constants to "cache" them, i.e. if available_memory is called in a loop, the re is only compiled once.

src/python/dxpy/sugar/__init__.py
jdidion: In the docstring under Returns I use  KiB, MiB, GiB. I'm not sure where else I'd use them here. But yes, we should use those same suffixes when printing the memory to the log.

src/python/dxpy/sugar/__init__.py
jdidion: Actually I think MemAvailable is the right one https://superuser.com/questions/980820/what-is-the-difference-between-memfree-and-memavailable-in-proc-meminfo

jdidion:
@Damien-Black made requested change. Please re-review.

src/python/dxpy/sugar/__init__.py
Damien-Black: Good catch

PR #536, opened 2019-06-28
Sci 1327 objects

jdidion:
Added module for dxpy object-related convenience methods.


src/python/dxpy/sugar/objects.py
```diff
@@ -0,0 +1,422 @@
...
+                "Folder {} does not exist in project {}".format(
+                    folder, project.get_id()
+                )
+            )
+        elif create:
+            LOG.info("Creating new folder %s in project %s", folder, project.get_id())
+            project.new_folder(folder, parents=True)
+            return []
+    elif exists is False:
+        raise dxpy.AppError(
```
Damien-Black: AppErrors are specific to execution environments (job environments). Here a `dxpy.DXSearchError` is preferred

```suggestion
        raise dxpy.DXSearchError(
```

src/python/dxpy/sugar/objects.py
```diff
@@ -0,0 +1,422 @@
...
+):
+    """Gets a project by name or ID. Creates a new project if there is no existing
+    project with the given name and `create is True`.
+
+    Args:
+        id_or_name: Project ID (project-XXX) or name. If this is a DXProject object,
+            it is validated and returned. Defaults to the PROJECT_CONTEXT_ID.
+        level: Minimum access level to search.
+        exists: Assert whether the project exists (True) or does not exist (False).
+        create: Whether to create the project if it does not exist.
```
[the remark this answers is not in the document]
jdidion: Yep - it's valid to create a project with the name "project-XXXXXXXX."

jdidion:
find_one_project etc are good in theory, but there are two problems:

1. There is no find_one_globalworkflow
2. The exception is the same whether there are too many or too few results, so I have to check the error message for whether it contains "more" or "none," which is clunky.

I changed the name of _create_one_handler to _find_one and based it off the _find_one code in search to make switching more seamless if we do it in future.

jdidion:
@Damien-Black please re-review

src/python/dxpy/sugar/objects.py
```diff
@@ -0,0 +1,439 @@
...
+            )
+        if next(candidates, None) is not None:
+            raise dxpy.DXSearchError(
+                "Expected one result, but found more: {}".format(str(kwargs))
+            )
+
+    return workflow
+
+
+def get_app_or_applet(app_or_applet_desc, project=None, **kwargs):
```
Damien-Black: Similar comment to `get_workflow`. This function won't be able to search for Apps.

src/python/dxpy/sugar/objects.py
```diff
@@ -0,0 +1,439 @@
...
+    return dxpy.find_one_app(
+        zero_ok=False,
+        more_ok=False,
+        name=app_desc,
+        return_handler=True,
+        **kwargs
+    )
+
+
+def get_applet(applet_desc, project=None, **kwargs):
```
Damien-Black: This functionality is already covered in `get_data_object `

src/python/test/sugar/test_objects.py
```diff
@@ -0,0 +1,211 @@
+from __future__ import print_function, unicode_literals, division, absolute_import
```
Damien-Black: I'll take a closer look at this module tomorrow and make a PR back.

AFAIK In `dx-toolkit`, the test don't assume any prior states and when they execute they'll lack access to projects or objects. I see that we use:

* [Temporary projects](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/dxpy_testutil.py#L143) when dealing with setting up temp projects.
* [DXTestCaseBuildWorkflows](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/dxpy_testutil.py#L463) to handle building test workflows when needed.

We should check with`dx-toolkit` maintainers with how they want these types of temporary resources handled.

src/python/dxpy/sugar/objects.py
```diff
@@ -0,0 +1,439 @@
...
+
+    project = get_project(project)
+    data_obj = None
+    search = True
+
+    if isinstance(data_obj_desc, dxpy.DXFile):
+        data_obj = data_obj_desc
+        search = False
+    elif resolver.is_data_obj_id(data_obj_desc):
+        data_obj = dxpy.get_handler(data_obj_desc, project)
```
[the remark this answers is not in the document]
jdidion: Good point. I will set classname=None in the signature, and default it to "file" when needed.

src/python/dxpy/sugar/objects.py
```diff
@@ -0,0 +1,439 @@
...
+        raise dxpy.DXSearchError("File {} does not to exist", data_obj_desc)
+    if data_obj and exists is False:
+        raise dxpy.DXSearchError(
+            "File {} exists but was expected not to exist", data_obj_desc
+        )
+
+    return data_obj
+
+
+def get_workflow(workflow_desc, project=None, **kwargs):
```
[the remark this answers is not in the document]
jdidion: Good point. I updated get_workflow to call get_data_object when searching for a project-based workflow, and get_app_or_applet to call get_data_object when searching for an applet. I also added some text to the docstrings explaining the use case for these methods. I also removed the kwargs from get_workflow and get_app_or_applet since the allowed parameters differ depending on whether it is project-based or global.

jdidion:
Thanks again, @Damien-Black. Please re-review.

jdidion:
@Damien-Black can you review this one as well? Would be great to finish up the initial release of sugar.

PR #546, opened 2019-07-17
Check if there is a newer version and if so, warn the user

jdidion:
Based on a suggestion from @yihhwang on the DevEx gchat channel.


jdidion:
Hopefully the code should only be run when the user runs the "dx" command with no arguments - that is what I intended. You may also want to look at caching the result for a set time period (e.g. 1 day).

PR #548, opened 2019-07-26
Sci 1986 fix filename regular expression

jdidion:
When the user does not provide a local filename to use with the dxpy.sugar.transfer.download* functions, we use the filename of the remote file, but we remove "unsafe" characters. The regular expression for identifying unsafe characters was missing '-'. The new regular expression is `r"[^\w-.]"`.

### DIFF
--- a/src/python/dxpy/sugar/transfers.py
+++ b/src/python/dxpy/sugar/transfers.py
@@ -64,7 +64,8 @@ from dxpy.sugar import processing as proc
 
 
 LOG = logging.getLogger()
-SPECIAL_RE = re.compile(r"[^\w.]")
+SPECIAL_RE = re.compile(r"[^\w-.]")
+"""Used based on recommendation here: https://superuser.com/a/748264/70028."""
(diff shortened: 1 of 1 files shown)


src/python/dxpy/sugar/transfers.py
```diff
@@ -64,7 +64,8 @@
 
 
 LOG = logging.getLogger()
-SPECIAL_RE = re.compile(r"[^\w.]")
+SPECIAL_RE = re.compile(r"[^\w-.]")
```
Damien-Black:
```suggestion
SPECIAL_RE = re.compile(r"[^\w\-\.]")
```

Only the `\-` is needed to not misinterpret the RE and a range [\w ... .], which is invalid and shouldn't compile. Added the `\.` just to future proof a bit if someone adds characters here.

jdidion: In an r-string, the escapes are redundant

PR #550, opened 2019-08-05
Sci 2012 bash template exit on command fails

Damien-Black:
**Motivation**

By default bash scripts continue execution even when they fail a command. As a result It's hard to detect failures in bash since they may pass without alerting users.

**Changes**

* This updates the dx-app-wizard templates used to include `set -e -o pipefail` preamble to bash scripts.

### DIFF
--- a/src/python/dxpy/templating/templates/bash/basic/dxapp.json
+++ b/src/python/dxpy/templating/templates/bash/basic/dxapp.json
@@ -2,6 +2,6 @@
   "runSpec": {
     "file": "src/DX_APP_WIZARD_NAME.sh",
     "distribution": "Ubuntu",
-    "release": "16.04"
+    "release": "14.04"
   }
 }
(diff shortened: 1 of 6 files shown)


Arkarachai:
@Damien-Black the set -e -o pipefail LGTM, but the Ubuntu 16.04 has already in placed. I'm just worried that there might be another branch. @kpjensen You can to 16.04 by default a few months back right? I file the request myself that time and I can see that the default from dx-app-wizard is 16.04

Damien-Black:
@Arkarachai you're right, in this  [commit](https://github.com/dnanexus/dx-toolkit/commit/018ec9d37b74faf2792908b5374cdfc695de9c34#diff-40b57273d29ca4bea451c4243a16e78c) we set the release to `16.04`.

The updates in the template are unneeded.

PR #552, opened 2019-08-29
TIP 1121: download database files using dx get

wkoetke:
Enhance 'dx get' to support database file download, essentially a very-simplified stripped down version of the 'dx download <file-id>' but for database (parquet) files instead of platform files:
- leverages much of the existing dx download codepath to perform the download
- invokes new route downloadFile (which is just a wrapper around readFile plus a few extra restrictions: must be internal user, must have license)
- only internal users can access for now, until proper billing is implemented
- 'filename' argument is the name of the file/directory object on S3 being downloaded, not a platform object


src/python/dxpy/cli/download.py
```diff
@@ -97,6 +98,41 @@ def download_one_file(project, file_desc, dest_filename, args):
     except:
         err_exit()
 
+def do_debug(msg):
+    logging.info(msg)
+
+# dest_filename = local file where downloaded file will go
+# src_filename = name of parquet file or folder being downloaded from database
+def download_one_database_file(project, file_desc, dest_filename, src_filename, file_status, args):
```
durae: shouldn't `file_desc` be `database_desc` ?

src/python/dxpy/scripts/dx.py
```diff
@@ -1821,6 +1825,47 @@ def get_workflow(entity_result, args):
     describe_output = entity_result['describe']
     dump_executable(dx_obj, output_path, omit_resources=True, describe_output=describe_output)
 
+def do_debug(msg):
+    logging.info(msg)
+
+def get_database(entity_result, args):
+    do_debug("dx.py#get_database - entity_result = {}".format(entity_result))
+    do_debug("dx.py#get_database - args = {}".format(args))
+    obj_name = entity_result['describe']['name']
```
durae: is `obj_name` used?

src/python/dxpy/bindings/dxdatabase.py
```diff
@@ -0,0 +1,339 @@
...
+    _class = "database"
+
+    _describe = staticmethod(dxpy.api.file_describe)
+
+    _http_threadpool_size = DXFILE_HTTP_THREADS
+    _http_threadpool = dxpy.utils.get_futures_threadpool(max_workers=_http_threadpool_size)
+
+    NO_PROJECT_HINT = 'NO_PROJECT_HINT'
+
+    def __init__(self, dxid=None, project=None, mode=None, read_buffer_size=DEFAULT_BUFFER_SIZE,
```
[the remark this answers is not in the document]
wkoetke: Well, it's intended to represent database data object. However, it's use cases are limited to downloading files (dx get) for now, so that representation is a little vague I admit. As for action items in the immediate term, @durae , I'll just remove any remaining unnecessary items like the ones you named, and I'll also refrain from adding stuff (e.g. databaseName) that would make sense but which aren't currently needed. Reasonable?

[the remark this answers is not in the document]
durae: yes, sounds alright for now. We can add it there later when needed for other database specific commands

src/python/dxpy/bindings/dxdatabase.py
```diff
@@ -0,0 +1,339 @@
...
+        return '{0:.2f} KiB'.format(B/KB)
+    elif MB <= B < GB:
+        return '{0:.2f} MiB'.format(B/MB)
+    elif GB <= B < TB:
+        return '{0:.2f} GiB'.format(B/GB)
+    elif TB <= B:
+        return '{0:.2f} TiB'.format(B/TB)
+
+def do_debug(msg):
+    logger.info(msg)
```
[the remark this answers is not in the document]
wkoetke: Yes, @durae , it's intended to be visible, provided the log level is set to debug. This is the correct pattern for dx toolkit that's used elsewhere. However, this line should be logger.debug, not logger.info, so I'll make that change.

src/python/dxpy/bindings/dxdatabase.py
```diff
@@ -0,0 +1,339 @@
...
+        # Initialize state
+        # self._pos = 0
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, traceback):
+        return
+
+    def __iter__(self):
```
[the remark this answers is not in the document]
wkoetke: Not necessary. Removed.

src/python/dxpy/bindings/dxdatabase.py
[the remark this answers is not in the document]
wkoetke: Done

src/python/dxpy/scripts/dx.py
wkoetke: Not used. Removed.

src/python/dxpy/bindings/dxdatabase.py
```diff
@@ -0,0 +1,339 @@
...
+            write buffer size will be constrained to be a multiple of
+            the allocation granularity)
+        :type file_is_mmapd: bool
+        """
+
+        DXDataObject.__init__(self, dxid=dxid, project=project)
+
+        # By default, a file is created in text mode. This makes a difference
+        # in python 3.
+        self._binary_mode = False
```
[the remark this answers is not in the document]
wkoetke: Not needed. Removed.

src/python/dxpy/bindings/dxdatabase.py
```diff
@@ -0,0 +1,339 @@
...
+                # logging.debug("Download URL unset or expired, requesting a new one")
+                if "timeout" not in kwargs:
+                    kwargs["timeout"] = FILE_REQUEST_TIMEOUT
+                do_debug("dxdatabase get_download_url - args = {}".format(args))
+                resp = dxpy.api.database_download_file(self._dxid, args, **kwargs)
+                do_debug("dxdatabase get_download_url - resp = {}".format(resp));
+                self._download_url = resp["url"]
+                self._download_url_headers = _validate_headers(resp.get("headers", {}))
+                if preauthenticated:
+                    self._download_url_expires = resp["expires"]/1000 - 60  # Try to account for drift
```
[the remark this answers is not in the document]
wkoetke: We don't use preauthenticated urls in readfile/downloadfile, so this unnecessary logic has been removed.

src/python/dxpy/scripts/dx.py
```diff
@@ -1821,6 +1825,44 @@ def get_workflow(entity_result, args):
...
+                                  args)
+    do_debug("dx.py#get_database - output_path = {}".format(output_path))
+    from dxpy.utils.executable_unbuilder import dump_executable
+    print("Downloading database files", file=sys.stderr)
+    dx_obj = dxpy.DXDatabase(obj_id)
+    describe_output = entity_result['describe']
+    do_debug("dx.py#get_database - dx_obj = {}".format(dx_obj))
+
+    # Call /database-xxx/listFolder to fetch database file metadata
+    list_folder_args = {"folder": args.filename, "recurse": args.recurse}
```
arunbhat: do we have a limit on the recurse or could the full database be downloaded?

wkoetke: Full database download is permitted. Shall I add a 'maxfilecount' argument to constrain this? The only downside I see is that it's kind of arbitrary, for example if you use maxfilecount of 1000 and there are 5000 files, which 1000 should be included...it would be in the order of the list returned by s3.listObjects.

arunbhat: hmm, how about a warning and only allow full database download if a "-f" argument is supplied. i am ok with downloading the whole thing since this a dnanexus only feature but don't want it to happen unknowingly..what do you think? @durae your thoughts?

durae: I would suggest a flag too (so it doesn't need to be run interactively), but we can't use `-f`. That's already reserved for `--overwrite`.

wkoetke: @durae and @arun, done. There is an --allow-all-files argument now which you need to specify if you omit the --filename arg. Otherwise, you get an error.

PR #561, opened 2019-10-17
Use a specific tag for samtools instead of :latest

billagee:
This is an attempt to avoid intermittent errors we see when pulling `quay.io/ucsc_cgl/samtools:latest` in tests.


orodeh:
I suggest getting rid of dx-docker altogether. Including the tests.

billagee:
Closing this since it looks like the tag change didn't help...

Issue #549, opened 2019-08-04
Running dx rm -r <project_id>:<file_id> warns me if I want to delete a project, not a file.


grepinsight:
On my local machine, I usually run `rm -rf <file or directory>` to delete a file or directory because that command works regardless of whether the target is a file or a directory. 

So I was very surprised to find that when  I ran `dx rm -r <project_id>:<file_id>` , I got the following warnning
```
  ===> dx rm -r project-<id-redacted>:file-<id-redacted>

===========================================================================
*     RED ALERT: Recursive deletion will remove all files in project!     *
*                                                                         *
*                  project-<id-redacted>                       *
*                                                                         *
*   Please issue 'dx rm -r --force' if you are sure you want to do this.  *
===========================================================================
(edited)

```
If I had an admin access to a particular project and I wasn’t careful, I would have blown up the whole project, because I might have issued `dx rm -rf <project_id>:<file_id> (!!!!)

Basically, if you specify a path with a  project:file_id syntax and a -r flag
it will try to delete the whole project even though the intention is to delete a file.

There’s a first safety check, but still people might inadvertently use dx rm -rf together.

The reason I am filing this issue is this:

__I think the behavior should be to delete a file, not a project if a file id is specified__
 

cc @jtratner

orodeh:
This warning is too strong. You *can* erase your project if you do:
```
dx rm -r project-xxxx: 
```

or, if you have selected it, then:
```
dx rm -r :
```

However, if you remove a single file, you are fine.

jtratner:
Could we fix the warning here then?

jtratner:
@kpjensen - looks like you fixed this? Can this ticket be closed? :)

PR #565, opened 2019-10-31
Fix project rm -r warning

kpjensen:
Only warn if attempting to delete an entire project, not a subfolder or file-id.
https://github.com/dnanexus/dx-toolkit/issues/549

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -858,7 +858,7 @@ def rm(args):
             try:
                 with nostderr():
                     project, folderpath, entity_results = resolve_existing_path(path, allow_mult=True, all_mult=args.all)
-                if folderpath == '/':
+                if folderpath == '/' and entity_results is None:
                     print("")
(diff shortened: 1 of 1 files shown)


src/python/dxpy/scripts/dx.py
```diff
@@ -858,7 +858,7 @@ def rm(args):
             try:
                 with nostderr():
                     project, folderpath, entity_results = resolve_existing_path(path, allow_mult=True, all_mult=args.all)
-                if folderpath == '/':
+                if folderpath == '/' and entity_results is None:
                     print("")
```
orodeh: does this additional check `entity_results = None` mean that the user is describing the root folder of the project, and nothing else?

kpjensen: Yes, if `entity_results` is something other than `None` then there is some object being described.

orodeh:
ok, I approve. But I am not longer the reviewer :-)

Can this be considered a tip?

PR #576, opened 2019-12-05
Allow disabling system exit on log client

jtratner:
@jdidion @geetduggal  - this PR adds some brief docs for the logclient (I may have formatted them wrong) + a way to disable the `SystemExit` that gets raised when a job is failed or terminated.

We've set something up to use this client to collect and store all logs from a job and it's quite confusing that it raises a SystemExit when you find a failed job.

### DIFF
--- a/src/python/dxpy/utils/job_log_client.py
+++ b/src/python/dxpy/utils/job_log_client.py
@@ -47,31 +47,30 @@ class DXJobLogStreamClient:
     ):
         """Initialize job log client.
 
-        :param job_id: dxid for a job (e.g. 'job-123' style)
+        :param job_id: dxid for a job (hash ID 'job-xxxx')
         :type job_id: str
         :param input_params: blob with connection parameters, should have keys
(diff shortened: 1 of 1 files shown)


src/python/dxpy/utils/job_log_client.py
```diff
@@ -43,8 +43,36 @@ class DXJobLogStreamingException(Exception):
...
+        :param msg_callback: single argument function that accepts a JSON blob with message
+        details. blob looks like
+        ``{"timestamp":1575465039481,"source":"APP","level":"STDOUT","job":"job-123","line":24,"msg":"success WfFragment"}``
+        where ``timestamp`` is milliseconds since epic and ``line`` is a sequence number.
+        :type msg_callback: callable
+        :param exit_on_failed: if True, will raise SystemExit with code of 3 if encountering a
+        failed job (this is the default behavior)
+        :type exit_on_failed: bool
+
+        NOTE: If you change the functionality of this client, please contact downstream users (including Myriad
```
jtratner: added this because there aren't tests and I'm hoping we can get a ping to make sure this still emits logs in the same way - I know it's a bit goofy :-/

PR #582, opened 2020-01-17
DEVEX-1416 py3 exec env support

kpjensen:
Python 3 execution environment support.
Add pip3 to execDepends.
Truncate joblog messages properly for streaming.

### DIFF
--- a/src/python/dxpy/dxlog.py
+++ b/src/python/dxpy/dxlog.py
@@ -82,15 +82,9 @@ class DXLogHandler(SysLogHandler):
         return message
 
     def is_resource_log(self, message):
-        is_resource_log_message = False
         if USING_PYTHON2:
-            if message.startswith(b"CPU: "):
-                is_resource_log_message = True
-        else:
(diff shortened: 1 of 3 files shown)


src/python/dxpy/dxlog.py
```diff
@@ -68,6 +69,29 @@ def encodePriority(self, record):
...
+                message = message[:8000] + "... [truncated]"
+        else:
+            # Trim bytes
+            encoded = message.encode('utf-8')
+            if len(encoded) > 8015:
+                # Ignore UnicodeDecodeError chars that could have been messed up by truncating
+                message = encoded[:8015].decode('utf-8', 'ignore') + "... [truncated]"
+        return message
+
+    def is_resource_log(self, message):
```
commandlinegirl: Maybe simpler would be:

```
    def is_resource_log(self, message):
        if USING_PYTHON2:
            return message.startswith(b"CPU: ")
        return message.startswith("CPU: ")
```

PR #585, opened 2020-01-28
Bump version of cryptography dependency

anujkumar93:
I received a security alert while using an older version of cryptography, and the source of this was due to an older dependency in dxpy. Also, other packages seem to have moved on to move their minimum version above 2.2.2, so this should fix dependency resolutions as well.


ying-w:
has there been any update to this?

kpjensen:
Released in v0.298.1

PR #586, opened 2020-01-30
DEVEX-1422 Fix tests to support Python3

commandlinegirl:
Passes the test suite except for the two sshing tests


commandlinegirl:
I am adding sorting to batch input generator since the tests relies on that.

PR #592, opened 2020-02-13
DEVEX-1383 asset builder for xenial v1

kpjensen:
Add `runSpecVersion` key to dxasset.json. If release is `xenial` and `runSpecVersion` is `"1"`, build the asset with the corresponding v1 asset builder app.

### DIFF
--- a/src/python/dxpy/asset_builder.py
+++ b/src/python/dxpy/asset_builder.py
@@ -222,6 +222,8 @@ def build_asset(args):
             app_run_result = dxpy.api.app_run(ASSET_BUILDER_PRECISE, input_params=builder_run_options)
         elif asset_conf['release'] == "14.04":
             app_run_result = dxpy.api.app_run(ASSET_BUILDER_TRUSTY, input_params=builder_run_options)
+        elif asset_conf['release'] == "16.04" and asset_conf['runSpecVersion'] == 1:
+            app_run_result = dxpy.api.applet_run('applet-applet-FjyxkZQ0x24p0vfpJVXKPYgx', input_params=builder_run_options)
         elif asset_conf['release'] == "16.04":
             app_run_result = dxpy.api.app_run(ASSET_BUILDER_XENIAL, input_params=builder_run_options)
 
--- a/src/python/dxpy/asset_builder.py
+++ b/src/python/dxpy/asset_builder.py
@@ -68,8 +68,7 @@ def validate_conf(asset_conf):
             "version": "0.0.1",
             "runSpecVersion": "1",
             "release": "16.04",
-            "version": 1,
-            "distribution": Ubuntu
+            "distribution": "Ubuntu"
             "execDepends":
                         [
                             {"name": "samtools", "package_manager": "apt"},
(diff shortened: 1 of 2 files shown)


src/python/dxpy/asset_builder.py
```diff
@@ -79,22 +82,30 @@ def validate_conf(asset_conf):
...
     if 'name' not in asset_conf:
         raise AssetBuilderException('The asset configuration does not contain the required field "name".')
-    # TODO: this default is not a good idea, and we will have to remove it once we ask customers to always provide release
-    if 'release' not in asset_conf:
-        asset_conf['release'] = "12.04"
-    elif asset_conf['release'] not in ['16.04', '14.04', '12.04']:
-        raise AssetBuilderException('The "release" field value should be either "12.04" (DEPRECATED), "14.04", "16.04".')
+
+    # Validate runSpec
+    if 'release' not in asset_conf or asset_conf['release'] not in ['16.04', '14.04', '12.04']:
```
commandlinegirl: That removes the ability to build assets for 12.04. Do we still allow building apps for 12.04? I would add to the title of this commit and to the CHANGELOG: "Remove ability to build assets for Ubuntu 12.04" or similar so that users know which older dxpy version to use if they really need to build 12.04 asset, e.g. to rebuild from source and repro old clinical trial results.

src/python/dxpy/asset_builder.py
```diff
@@ -79,22 +82,30 @@ def validate_conf(asset_conf):
...
+        raise AssetBuilderException('The "release" field value should be either "16.04", "14.04", or "12.04" (DEPRECATED)')
+    if 'runSpecVersion' in asset_conf:
+        if asset_conf['runSpecVersion'] not in ['0','1']:
```
commandlinegirl: Nit: add space after comma and set " instead of ' for strings (here and in similar places in this commit). Thanks!

src/python/dxpy/asset_builder.py
```diff
@@ -65,8 +66,10 @@ def validate_conf(asset_conf):
             "title": "A human readable name",
             "description": " A detailed description abput the asset",
             "version": "0.0.1",
-            "distribution": "Ubuntu",# (Optional)
-            "release": "12.04",
+            "runSpecVersion": "1",
+            "release": "16.04",
+            "version": 1,
```
commandlinegirl: This "version" repeats the "version": "0.0.1" in line 69, can be removed.

src/python/dxpy/asset_builder.py
```diff
@@ -65,8 +66,10 @@ def validate_conf(asset_conf):
...
+            "distribution": Ubuntu
```
commandlinegirl: Please enclose Ubuntu in "".

src/python/dxpy/asset_builder.py
kpjensen: It removes defaulting to 12.04, or defaulting to any release. It's still a valid release to build an asset.

commandlinegirl: Ah, right.

PR #603, opened 2020-03-26
DEVEX-1591: add dx-mount-all-inputs script (similar to dx-download-all-inputs)

wkoetke:
Add dx-mount-all-inputs script, which is similar to dx-download-all-inputs except that instead of downloading the input files, it mounts them using dxfuse. This helps the performance of apps like csv_loader and vcf_loader which deal with very large input files.

### DIFF
--- a/src/python/dxpy/bindings/mount_all_inputs.py
+++ b/src/python/dxpy/bindings/mount_all_inputs.py
@@ -143,6 +143,9 @@ def mount_all_inputs(exclude=None):
     with open(mount_manifest_file, 'w') as mfile:
         json.dump(files_manifest, mfile)
 
+    dxfuse_version = subprocess.check_output([dxfuse_cmd, "-version"])
+    print("Using dxfuse version " + str(dxfuse_version))
+
     uid = str(int(subprocess.check_output(["id", "-u"])))
     gid = str(int(subprocess.check_output(["id", "-g"])))
     print(subprocess.check_output([dxfuse_cmd, "-uid", uid, "-gid", gid, mount_dir, mount_manifest_file]))
--- a/src/python/dxpy/bindings/mount_all_inputs.py
+++ b/src/python/dxpy/bindings/mount_all_inputs.py
@@ -18,8 +18,8 @@ from __future__ import print_function
 import subprocess
 import json
 import os
-import dxpy
 from dxpy.utils import file_load_utils
+from dxpy.exceptions import err_exit
 
 def _build_mount_manifest(to_mount):
     files_list = []
@@ -39,6 +39,15 @@ def _build_mount_manifest(to_mount):
     print(files_manifest)
     return files_manifest
 
+def _which(program):
+    def is_exe(fpath):
+        return os.path.isfile(fpath) and os.access(fpath, os.X_OK)
+
+    for path in os.environ["PATH"].split(os.pathsep):
+        exe_file = os.path.join(path, program)
+        if is_exe(exe_file):
+            return exe_file
+    return None
 
 def _gen_helper_dict(filtered_inputs):
     '''
@@ -109,7 +118,9 @@ def mount_all_inputs(exclude=None):
     home_dir = os.environ["HOME"]
     mount_dir = os.path.join(home_dir, "in")
     mount_manifest_file = os.path.join(home_dir, "mount-manifest.json")
-    dxfuse_cmd = "/usr/bin/dxfuse"
+    dxfuse_cmd = _which("dxfuse")
+    if dxfuse_cmd is None:
+        err_exit("dxfuse is not installed on this system")
 
     subprocess.check_output(["mkdir", mount_dir])
 
--- a/src/python/dxpy/bindings/mount_all_inputs.py
+++ b/src/python/dxpy/bindings/mount_all_inputs.py
@@ -1,0 +1,147 @@
+# Copyright (C) 2014-2020 DNAnexus, Inc.
+#
+# This file is part of dx-toolkit (DNAnexus platform client libraries).
+#
+#   Licensed under the Apache License, Version 2.0 (the "License"); you may not
+#   use this file except in compliance with the License. You may obtain a copy
+#   of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#   License for the specific language governing permissions and limitations
+#   under the License.
+
+from __future__ import print_function
+import subprocess
+import json
+import os
+import dxpy
+from dxpy.utils import file_load_utils
+
+def _build_mount_manifest(to_mount):
+    files_list = []
+    for file_rec in to_mount:
+        file_entry = {}
+        file_handler = file_rec['handler']
+        file_entry['proj_id'] = file_handler.get_proj_id()
+        file_entry['file_id'] = file_rec['src_file_id']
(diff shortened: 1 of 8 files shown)


wkoetke:
Per discussion with @orodeh , I need to make additional commit to determine which dxfuse command should be run rather than making assumptions about its name, location, or availability. Stay tuned for that...

src/python/dxpy/bindings/mount_all_inputs.py
```diff
@@ -0,0 +1,147 @@
...
+        file_entry['file_id'] = file_rec['src_file_id']
+        file_name = file_rec['trg_fname']
+        k = file_name.rfind("/")
+        file_parent = file_name[:k]
+        file_entry['parent'] = "/in/" + file_parent
+        files_list.append(file_entry)
+    files_manifest = {}
+    files_manifest['Files'] = files_list
+    print("File mount manifest:")
+    print(files_manifest)
```
arunbhat: Can you please give an example of this for the csv loader with the etl config file specified?

commandlinegirl:
@wkoetke Thanks for the PR! I'm in meetings this afternoon but will review either in between or this evening.

src/python/dxpy/bindings/mount_all_inputs.py
```diff
@@ -0,0 +1,147 @@
...
+        file_handler = file_rec['handler']
+        file_entry['proj_id'] = file_handler.get_proj_id()
+        file_entry['file_id'] = file_rec['src_file_id']
+        file_name = file_rec['trg_fname']
+        k = file_name.rfind("/")
+        file_parent = file_name[:k]
+        file_entry['parent'] = "/" + file_parent
+        files_list.append(file_entry)
+    files_manifest = {}
+    files_manifest['Files'] = files_list
```
commandlinegirl: It seems the _files_ key in the manifest is not capitalized (according to the documentation: https://github.com/dnanexus/dxfuse/blob/master/doc/Internals.md#manifest). Or is it working with 'Files' as well?

src/python/scripts/dx-mount-all-inputs
```diff
@@ -0,0 +1,72 @@
...
+    {"$dnanexus_link": "file-xxxx"}
+
+and filename INPUT.TXT, then the linked file will be mounted into the
+path:
+
+    $HOME/in/FOO/INPUT.TXT
+
+If an input is an array of files, then all files will be placed into
+numbered subdirectories under a parent directory named for the
+input. For example, if the input key is FOO, and the inputs are {A, B,
```
commandlinegirl: When multiple files in the array have the same name, e.g. {A, A, A}.vcf, will the target directory structure look the same?

src/python/scripts/dx-mount-all-inputs
```diff
@@ -0,0 +1,72 @@
...
+parser.add_argument('--except',
+                    help=fill('Do not mount the input with this name. (May be used multiple times.)',
+                              width_adjustment=-20),
+                    action="append",
+                    dest="exclude",
+                    default=[])
+
+args = parser.parse_args()
+
+dxpy.mount_all_inputs(exclude=args.exclude)
```
commandlinegirl: Would it be possible to add a test for it? It could go to [test_dxpy.py](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/test_dxpy.py). It'd, for example, build an applet in a new project (built with "temporary_project"), running only dx-mount-all-inputs and listing (asserting) the mounted files.

src/python/dxpy/bindings/mount_all_inputs.py
wkoetke: 'Files' works, yes.

wkoetke:
> Per discussion with @orodeh , I need to make additional commit to determine which dxfuse command should be run rather than making assumptions about its name, location, or availability. Stay tuned for that...

This is now done.

src/python/dxpy/bindings/mount_all_inputs.py
```diff
@@ -0,0 +1,147 @@
...
+
+    This allows using shell globbing (FOO/*/*.vcf) to get all the files in the input
+    order and prevents issues with files which have the same filename.'''
+
+    print("Mounting inputs...")
+
+    home_dir = os.environ["HOME"]
+    mount_dir = os.path.join(home_dir, "in")
+    mount_manifest_file = os.path.join(home_dir, "mount-manifest.json")
+    dxfuse_cmd = "/usr/bin/dxfuse"
```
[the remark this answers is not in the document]
wkoetke: It won't be 'provided' (b/c we can't provide it for all platforms/versions/environments), but in the code we look for a valid available executable (and throw an error if there isn't one). Same pattern used elsewhere in the toolkit (e.g. looking for 'wget'):

https://github.com/dnanexus/dx-toolkit/pull/603/commits/7e89c201e6d6a36b850dfef74b9b15cc0ec6b144

src/python/dxpy/bindings/mount_all_inputs.py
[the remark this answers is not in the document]
wkoetke: Done

src/python/dxpy/bindings/mount_all_inputs.py
```diff
@@ -0,0 +1,147 @@
...
+    files_list = []
+    for file_rec in to_mount:
+        file_entry = {}
+        file_handler = file_rec['handler']
+        file_entry['proj_id'] = file_handler.get_proj_id()
+        file_entry['file_id'] = file_rec['src_file_id']
+        file_name = file_rec['trg_fname']
+        k = file_name.rfind("/")
+        file_parent = file_name[:k]
+        file_entry['parent'] = "/" + file_parent
```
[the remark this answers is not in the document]
wkoetke: Done

src/python/scripts/dx-mount-all-inputs
wkoetke: @commandlinegirl , I added an app test for dx-mount-all-inputs similar to the 'file_load' category for the dx-download-all-inputs script. Needs to be run in an app, rather than a unit level test, because dx-mount-all-inputs (just like dx-download-all-inputs) throws an error if not run from inside a job.

src/python/scripts/dx-mount-all-inputs
wkoetke: Those same file names end up getting mounted in unique locations, just like dx-download-all-input downloads them into unique locations. So for an array of files that get mounted it would look like this:

/home/dnanexus/in/csv/0/newdata1.csv
/home/dnanexus/in/csv/1/newdata1.csv
/home/dnanexus/in/csv/2/newdata2.csv

src/python/dxpy/bindings/mount_all_inputs.py
```diff
@@ -0,0 +1,155 @@
...
+    to_mount = []
+    for ival_list in inputs.values():
+        to_mount.extend(ival_list)
+
+    files_manifest = _build_mount_manifest(to_mount)
+    with open(mount_manifest_file, 'w') as mfile:
+        json.dump(files_manifest, mfile)
+
+    uid = str(int(subprocess.check_output(["id", "-u"])))
+    gid = str(int(subprocess.check_output(["id", "-g"])))
```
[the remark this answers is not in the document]
wkoetke: Done

src/python/test/file_mount/basic/dxapp.json
```diff
@@ -0,0 +1,20 @@
+{ "name": "basic",
+  "title": "basic",
+  "summary" : "basic file mount",
+  "runSpec": {
+    "file": "run.sh",
+    "interpreter": "bash",
+    "distribution": "Ubuntu",
+    "release": "14.04"
```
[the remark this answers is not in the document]
wkoetke: Done

PR #604, opened 2020-04-01
DEVEX-1386 'python3' support for dx-app-wizard

geetduggal:
Adds support for 'interpreter':'python3' with 'version':'1', Ubuntu 16.04 in the toolkit. This will be replacing the old Python 2 as the default.  Tested manually the creation, building, and running of both bash and Python 3 apps.  Added tests app wizard testing script that is also run for traceability testing.

I believe documentation of the python3 support at large is a separate ticket, so this will be the only PR for this ticket.

### DIFF
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -230,7 +230,7 @@ def _check_syntax(code, lang, temp_dir, enforce=True):
     """
     # This function needs the language to be explicitly set, so we can
     # generate an appropriate temp filename.
-    if lang == 'python2.7' or 'python3' in lang':
+    if lang == 'python2.7' or 'python3' in lang:
         temp_basename = 'inlined_code_from_dxapp_json.py'
     elif lang == 'bash':
         temp_basename = 'inlined_code_from_dxapp_json.sh'
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -230,12 +230,12 @@ def _check_syntax(code, lang, temp_dir, enforce=True):
     """
     # This function needs the language to be explicitly set, so we can
     # generate an appropriate temp filename.
-    if lang == 'python2.7':
+    if lang == 'python2.7' or lang == 'python3':
         temp_basename = 'inlined_code_from_dxapp_json.py'
     elif lang == 'bash':
         temp_basename = 'inlined_code_from_dxapp_json.sh'
     else:
-        raise ValueError('lang must be one of "python2.7" or "bash"')
+        raise ValueError('lang must be one of "python2.7", "python3", or "bash"')
     # Dump the contents out to a temporary file, then call _check_file_syntax.
     with open(os.path.join(temp_dir, temp_basename), 'w') as ofile:
         ofile.write(code)
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -230,7 +230,7 @@ def _check_syntax(code, lang, temp_dir, enforce=True):
     """
     # This function needs the language to be explicitly set, so we can
     # generate an appropriate temp filename.
-    if lang == 'python2.7' or lang == 'python3':
+    if lang == 'python2.7' or 'python3' in lang':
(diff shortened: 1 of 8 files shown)


emiloslavsky:
don't we also need to modify the templates? e.g. https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/templating/templates/python/basic/test/test.py#L52

emiloslavsky:
Also I wonder if DEVEX-1524 is still broken.

commandlinegirl:
@geetduggal Would you also update [this](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/utils/executable_unbuilder.py#L100) line.

It should include "python3" as well.

commandlinegirl:
As well as in the script generating dxapp.json: https://github.com/dnanexus/dx-toolkit/blob/813ac1357062441d436268e6076a20ecaa2530c9/src/python/scripts/dx-generate-dxapp#L319

geetduggal:
Thanks @emiloslavsky !  Yes, I think the templates are a great point.  Looking at them, it seems like there may be multiple issues, also related to Python 3.

commandlinegirl:
@geetduggal Could you also add python3 to this condition: 

https://github.com/dnanexus/dx-toolkit/blob/42fa7faac36b4c8e7de94417ee0a65a57021707a/src/python/dxpy/scripts/dx_build_app.py#L233

geetduggal:
> Requesting the updates in the interpreter choices (see my other comments).
> 
> Also, the templates need to updated:
> 
> ```
> dx-app-wizard --help
> (...)
>   --template {basic,parallelized,scatter-process-gather}
>                         Execution pattern of your app
> ```
> 
> The code for these templates is here: https://github.com/dnanexus/dx-toolkit/tree/master/src/python/dxpy/templating/templates
> 
> The question is how often these templates are even used? Perhaps we could deprecate them for Python3.

Thanks @commandlinegirl !  I discussed this with @emiloslavsky this afternoon and we decided if its a quick fix then we will keep the templates.  It turned out to be a pretty quick one and all three templates work as expected with a Python 3 compile of the toolkit.  So we'll leave it in for the moment.

geetduggal:
> @geetduggal Could you also add python3 to this condition:
> 
> https://github.com/dnanexus/dx-toolkit/blob/42fa7faac36b4c8e7de94417ee0a65a57021707a/src/python/dxpy/scripts/dx_build_app.py#L233

Thanks!  I updated it.  We may also need separate ticket(s) for other scripts like these

geetduggal:
Addressed comments inline.
I also manually verified all apps for the templates are created with 'python3' and 'version' 1
dx-app-wizard integration tests also pass with my modifications and addition to the tests

geetduggal:
@commandlinegirl  @emiloslavsky I've completed this pass addressing the review.  Please re-review at your leisure.

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -230,12 +230,12 @@ def _check_syntax(code, lang, temp_dir, enforce=True):
     """
     # This function needs the language to be explicitly set, so we can
     # generate an appropriate temp filename.
-    if lang == 'python2.7':
+    if lang == 'python2.7' or lang == 'python3':
```
commandlinegirl, line 233: Actually, it could cover other "python3" version added down the road (including python3.5), so how about:

```suggestion
    if lang == 'python2.7' or 'python3' in lang':
```

commandlinegirl:
@geetduggal Thanks! I think we also have to update prints in each of the python template tests, (with print import from future), e.g. [here](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/templating/templates/python/basic/test/test.py#L38).

geetduggal:
> @geetduggal Thanks! I think we also have to update prints in each of the python template tests, (with print import from future), e.g. [here](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/templating/templates/python/basic/test/test.py#L38).

Thanks @commandlinegirl I updated it. Also, since dxpy already imports `print_function` from `__future__`, anyplace else where we import that that also imports dxpy may be redundant.

geetduggal:
@commandlinegirl please re-review at your convenience

PR #607, opened 2020-04-08
docs: python 3.3, 3.4 -> 3.5. py2.7 supports ends on March 1, 2021

emiloslavsky:


jdidion:
I think you should explicitly say what the minimum python version is. Even just "Python 3.5+".

PR #609, opened 2020-04-09
DEVEX-1497 Error out for dx run --clone with analysis and executable

commandlinegirl:
cc @emiloslavsky

### DIFF
--- a/src/python/dxpy/scripts/dx.py
+++ b/src/python/dxpy/scripts/dx.py
@@ -3277,6 +3277,11 @@ def run(args):
                     }
                 }
         else:
+            if args.executable != "":
+                error_mesg = "Workflow executable (\"{}\") cannot be provided when re-running an analysis with 'dx run --clone'. ".format(args.executable)
+                error_mesg += "You can instead run 'dx run --clone {}', with other optional CLI arguments, to re-run the previously run analysis.".format(args.clone)
+                err_exit(exception=DXParserError(error_mesg),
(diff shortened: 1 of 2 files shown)


src/python/dxpy/scripts/dx.py
```diff
@@ -4727,7 +4732,7 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
                         action='append', type=str)
 
 parser_run.add_argument('-h', '--help', help='show this help message and exit', nargs=0, action=runHelp)
-parser_run.add_argument('--clone', help=fill('Job or analysis ID or name from which to use as default options (will use the exact same executable ID, destination project and folder, job input, instance type requests, and a similar name unless explicitly overridden by command-line arguments)', width_adjustment=-24))
+parser_run.add_argument('--clone', help=fill('Job or analysis ID or name from which to use as default options (will use the exact same executable ID, destination project and folder, job input, instance type requests, and a similar name unless explicitly overridden by command-line arguments. When using an analysis with --clone a workflow executable cannot be overriden and should not be provided)', width_adjustment=-24))
```
geetduggal, line 4735:
```suggestion
parser_run.add_argument('--clone', help=fill('Job or analysis ID or name from which to use as default options (will use the exact same executable ID, destination project and folder, job input, instance type requests, and a similar name unless explicitly overridden by command-line arguments. When using an analysis with --clone a workflow executable cannot be overriden and should not be provided.)', width_adjustment=-24))
```

PR #619, opened 2020-05-09
APPS-45 Add --brief option to dx build

commandlinegirl:
Added this option for dx build apps, still need to add it for workflows.

I'm wondering if there a better way to temporarily set logging to, say, ERROR, so that warnings and info are not displayed with the `--brief` option.

WIP


commandlinegirl:
> LGTM. I will open an additional PR to fix gnureadline requirements install.

Thank you! I will remove the gnureadline fix from this PR then so that it's not mixed up with dx build fixes.

PR #624, opened 2020-05-21
QE-496: Upload agent scenario

allaco-dn:
- Upload agent example scenario added
- expect_error if not present by default is False

### DIFF
--- a/src/python/test/behave_tests/features/dx.feature
+++ b/src/python/test/behave_tests/features/dx.feature
@@ -4,8 +4,8 @@ Feature: 1: dx commands example
   Scenario: 1010_CLI new project test
     Given CLI login as user qa_admin_org1
     When CLI create new project
-      | name      | --bill-to     | expect_error |
-      | [default] | org-test_org1 | no           |
+      | name      | --bill-to     |
+      | [default] | org-test_org1 |
     Then login as user qa_admin_org1
     And add qa_tester_orgs123 user as ADMINISTER to project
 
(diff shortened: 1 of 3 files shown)


src/python/test/behave_tests/features/dx.feature
```diff
@@ -20,7 +20,6 @@ Feature: 1: dx commands example
   Scenario: 1030_CLI file upload test, project select on login
     Given CLI login as user qa_admin_org1
     When CLI create new project
-    # TODO: remove expect error (by default no)
```
avigriyanova, line 23: expect_error is still there, is it according to plan? :D

src/python/test/behave_tests/features/upload_agent.feature
```diff
@@ -0,0 +1,11 @@
+@upload_agent
+Feature: 2: Upload agent example
+
+  Scenario: 2000_upload file with upload agent
+    Given CLI login as user qa_admin_org1
+    And CLI create new project
+      | name      | --select |
+      | [default] |          |
+    And generate file big 2 Bytes named zzz.file
```
avigriyanova, line 9: why not `generate big 2 Bytes file named zzz.file`?

src/python/test/behave_tests/features/dx.feature
allaco-dn: Yes, to show that both are supported :D But it is not mandatory

src/python/test/behave_tests/features/upload_agent.feature
allaco-dn: It is an existing step, phrased like that...

avigriyanova: weird by ok :D I guess it's not worth to fix it now

PR #625, opened 2020-05-29
TIP-1447: clear error message when no response entity available

wkoetke:
When no response entity available, make this more clear in the error message. Previously, it would just show as " IOException: . "


arunbhat:
LGTM. But Ola or Geet should approve!

PR #628, opened 2020-07-21
DEVEX-1732 Ubuntu 20.04 build targets and py 3.8

kpjensen:
Add build targets for dx-java-bindings debian package in prep for Ubuntu 20.04 AEE and tarball with python 3. 

Remove old script `dx_build_report_html.py` and associated tests. Update test applets to provide instance type in dxapp.json.

Fix api wrapper generators for python 3. Restrict requests to < 2.24.0 as it breaks some SSL tests.


commandlinegirl:
LGTM, thanks. 

Re Python3.8, one more change we'll need to make in dxpy - I got this warning today when running dxpy on Py3.7:

```
  /Users/ozalcman/Library/Python/3.7/lib/python/site-packages/dxpy/compat.py:22: 
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated,
and in 3.8 it will stop working
    from collections import MutableMapping
```

commandlinegirl:
Thanks!

PR #638, opened 2020-08-28
DEVEX-1769 Hide help output of `dx notebook`

kpjensen:
Add back removed `dx notebook` and `dx loupe-viewer` commands, but hide the subparsers from `dx` help output.


src/python/dxpy/scripts/dx.py
```diff
@@ -3880,6 +3880,9 @@ def __call__(self, parser, namespace, values, option_string):
 class DXArgumentParser(argparse.ArgumentParser):
     def _print_message(self, message, file=None):
         if message:
+            if message.startswith("usage: dx [-h] [--version] command") and "dx: error: argument command: invalid choice:" in message:
```
emiloslavsky, line 3883: with python3, is message guaranteed to be a string?

kpjensen: I assumed it is since the message is decoded if `USING_PYTHON2`. Works locally with Python 3.8, but I'm running it through the jenkins test suite.

PR #654, opened 2020-12-09
APPS-89 detach mode for dx run

mhrvol:
Is there some "whats new" file?

Added support for --detach while using dx run


src/python/dxpy/scripts/dx.py
```diff
@@ -4881,6 +4884,7 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
                         help=fill('Print help and examples for how to specify inputs',
                                   width_adjustment=-24),
                         action=runInputHelp, nargs=0)
+parser_run.add_argument('--detach', help=fill("New job will not start as subjob if run inside of a different job.", width_adjustment=-24), action='store_true')
```
commandlinegirl, line 4887: I think we can use a 'detached job' expression, so for example: "New job will start as a detached job if run inside of a different job", to tell the user what the new job *will* be launched as.

src/python/dxpy/bindings/dxapplet.py
```diff
@@ -158,7 +161,7 @@ def _get_cleanup_keys(self):
     def run(self, executable_input, project=None, folder=None, name=None, tags=None, properties=None, details=None,
             instance_type=None, stage_instance_types=None, stage_folders=None, rerun_stages=None, cluster_spec=None,
             depends_on=None, allow_ssh=None, debug=None, delay_workspace_destruction=None, priority=None,
-            ignore_reuse=None, ignore_reuse_stages=None, extra_args=None, **kwargs):
+            ignore_reuse=None, ignore_reuse_stages=None, extra_args=None, detach=None, **kwargs):
```
commandlinegirl, line 164: Minor: we tend to keep the extra_args=None, **kwargs at the end of the params list, so please move detach=None before extra_args, thanks!

src/python/dxpy/toolkit_version.py
```diff
@@ -1 +1 @@
-version = '0.304.0'
+version = '0.304.1'
```
commandlinegirl, line 1: Afair this will be updated automatically at the release.

commandlinegirl (changes requested):
Minor issues mostly.

commandlinegirl:
> Is there some "whats new" file?

Yes, you can add a new item to [CHANGELOG](https://github.com/dnanexus/dx-toolkit/blob/master/CHANGELOG.md#unreleased), thanks!

Also, please change the commit/PR title from Apps-89 to APPS-89 so that it's picked up by JIRA.

> Added support for --detach while using dx run

src/python/dxpy/scripts/dx.py
mhrvol: I did not want to use the word "detached" to describe what "detach" means.  But sure, I'll update it

src/python/dxpy/bindings/dxapplet.py
mhrvol: fixed

src/python/dxpy/toolkit_version.py
mhrvol: removed this change

src/python/dxpy/bindings/dxapplet.py
```diff
@@ -222,8 +227,8 @@ def run(self, executable_input, project=None, folder=None, name=None, tags=None,
                                         debug=debug,
                                         delay_workspace_destruction=delay_workspace_destruction,
                                         priority=priority,
-                                        extra_args=extra_args)
-
+                                        extra_args=extra_args,
+                                        detach=detach)
```
commandlinegirl, line 231: Here `detach` also should move before extra_args. And also please move it in the docstrings.

src/python/dxpy/scripts/dx.py
```diff
@@ -4881,6 +4884,7 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
                         help=fill('Print help and examples for how to specify inputs',
                                   width_adjustment=-24),
                         action=runInputHelp, nargs=0)
+parser_run.add_argument('--detach', help=fill("New job will start as a detached job if run inside of a different job.", width_adjustment=-24), action='store_true')
```
commandlinegirl, line 4887: So maybe: "as a detached job (not a subjob) if run ..".

commandlinegirl (changes requested):

src/python/dxpy/scripts/dx.py
emiloslavsky:
```suggestion
parser_run.add_argument('--detach', help=fill("When invoked from a job, detaches the new job from the creator job so the new job will appear as a typical root execution.  Setting DX_RUN_DETACH environment variable to 1 causes this option to be set by default.", width_adjustment=-24), action='store_true')
```

emiloslavsky (changes requested):
What about (from https://confluence.internal.dnanexus.com/display/TTN/Detached+Jobs)
```
If `dx run` fails dues to lack of org.licenseFeatures.detachedJob feature, the error message should tell the user how to disable the global "always detach jobs" configuration if that configuration is set.
```

src/python/dxpy/bindings/dxapplet.py
```diff
@@ -192,6 +195,8 @@ def run(self, executable_input, project=None, folder=None, name=None, tags=None,
         :type ignore_reuse_stages: list
         :param extra_args: If provided, a hash of options that will be merged into the underlying JSON given for the API call
         :type extra_args: dict
+        :param detach: If provided, job will not start as subjob if run inside of a different job.
```
commandlinegirl, line 198: Please change detach location here in docstrings, too.

mhrvol: fixed

commandlinegirl (commented):
Looks good but could you add a simple test to test_dx_client.py?

src/python/dxpy/scripts/dx.py
```diff
@@ -2869,6 +2870,11 @@ def run_one(args, executable, dest_proj, dest_path, input_json, run_kwargs):
                 else:
                     ssh_args = parser.parse_args(['ssh', dxexecution.get_id()])
                 ssh(ssh_args, ssh_config_verified=True)
+    except PermissionDenied as e:
+        if run_kwargs.get("detach") and os.environ.get("DX_RUN_DETACH") == "1" and "detachedJob" in e.msg:
+            print("Unable to start detached job in given project. "
+                  "To disable running detached job by default unset the environment variable DX_RUN_DETACH ('unset DX_RUN_DETACH')")
```
emiloslavsky, line 2876:
```suggestion
                  "To disable running jobs as detached by default, please unset the environment variable DX_RUN_DETACH ('unset DX_RUN_DETACH')")
```

src/python/dxpy/scripts/dx.py
```diff
@@ -2869,6 +2870,11 @@ def run_one(args, executable, dest_proj, dest_path, input_json, run_kwargs):
...
+        err_exit()
```
emiloslavsky, line 2877: I'd `raise(e)` to show the actual error that mentions detachedJob license.

emiloslavsky (changes requested):
minor wording change

src/python/dxpy/scripts/dx.py
mhrvol: The actual error is shown as well, as err_exit() shows the error. 

Do you want me to create a new, different exception than the one that really occured?

emiloslavsky: In that case, no need to `raise(e)`.  Thanks Marek!

src/python/test/test_dxclient.py
```diff
@@ -963,6 +965,64 @@ def _test_dx_ssh(self, project, instance_type):
...
+                dx.sendline("dx run {} --yes --detach".format(sleep_applet2))
+                time.sleep(1)
+                job2 = next(dxpy.find_jobs(name="sleep2", project=self.project), None)
+                print(job2)
+                self.assertTrue(job_id in job2['detachedFrom'])
+                dx.sendline("exit")
+                dx.expect("bash running", timeout=10)
+                dx.sendcontrol("c")  # CTRL-c
+                dx.expect("[exited]")
+                dx.expect("dnanexus@job", timeout=10)
```
commandlinegirl, line 1024: Thanks for adding the test! It's going to be super useful (the pexpect experience) for our other interactive apps. 

Please:
* terminate both jobs at the end, with something like: run("dx terminate " + job_id)
* remove print

commandlinegirl (approved):
LGTM, with the two updates to test that I mention in the comments. Thanks!

src/python/dxpy/scripts/dx.py
mhrvol: I changed it to raise(e)

but note that this change will not affect anything as err_exit() part of the code calls this:
`    exc = exception if exception is not None else sys.exc_info()[1]
    if isinstance(exc, SystemExit):
        raise exc`
        
+ does some other code that should not affect anything in this case.

src/python/test/test_dxclient.py
mhrvol: As discussed, I'll leave this test out for now - it will take longer than expected

commandlinegirl: SG, thanks!

commandlinegirl (approved):

emiloslavsky (approved):

PR #655, opened 2020-12-09
DEVEX-1779 Fix symlink download error 

mhrvol:


src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -179,6 +179,22 @@ def _download_symbolic_link(dxid, md5digest, project, dest_filename):
                                             duration=6*3600,
                                             project=project)
 
+    def call_cmd(cmd, retry=True):
```
commandlinegirl, line 182: Our convention elsewhere in the toolkit is to name it `ok_to_retry`.

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -179,6 +179,22 @@ def _download_symbolic_link(dxid, md5digest, project, dest_filename):
...
+            if aria2c_exe is not None:
+                print("Downloading symbolic link with aria2c")
+            else:
+                print("Downloading symbolic link with wget")
+            subprocess.check_call(cmd, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            msg = ""
+            if e and e.output:
+                msg = e.output.strip()
+            if e.returncode == 22 and retry:  # hotfix, DEVEX-1779
```
commandlinegirl, line 193: In case the error persists we want to break from retries after a few retries, one example of retries with exponential backoff is [here](https://github.com/dnanexus/dx-toolkit/blob/c4906fe48f757260dd528efe9bf1d50f26076a46/src/python/dxpy/__init__.py#L782).

commandlinegirl (changes requested):
Please run the simlink tests with Py2 and Py3 (e.g. in jenkins branches).

src/python/dxpy/bindings/dxfile_functions.py
mhrvol: I've implemented the "exponential backoff" from the other comment. Is this fine?

src/python/dxpy/bindings/dxfile_functions.py
mhrvol: Thanks

src/python/dxpy/bindings/dxfile_functions.py
commandlinegirl: Looks good, I'd only set the max_retries to e.g. 4.

commandlinegirl:
Looks good, thanks! Did you have a chance to run a test on this on jenkins (or locally?)

mhrvol:
> Looks good, thanks! Did you have a chance to run a test on this on jenkins (or locally?)

https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/11/

Symlink tests passed. Although this is by no means proof that the error will not reappear.

commandlinegirl (approved):
LGTM, thanks!

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -179,6 +181,24 @@ def _download_symbolic_link(dxid, md5digest, project, dest_filename):
                                             duration=6*3600,
                                             project=project)
 
+    def call_cmd(cmd, max_retries=2, num_attempts=0):
```
XingGemini, line 184: Should you define constant values as DEFAULT_MAX_RETRIED and DEFAULT_NUM_ATTEMPS instead of hardcoding them?  

I don't know too much about engineering practice.  @commandlinegirl

XingGemini (commented):
LGTM.  Have a question about general practice.

src/python/dxpy/bindings/dxfile_functions.py
commandlinegirl: Please also address this before merging. @mhrvol

src/python/dxpy/bindings/dxfile_functions.py
commandlinegirl: Thanks @XingGemini  num_attempts is recursively updated value of the current attempt number and initialized at 0 so it's very local to this function. max_retries could be extracted but it's also rather local to this particular inner function so IMHO can stay as is.

XingGemini: Fair enough.  Thanks for the clarification.

XingGemini:
LGTM.  Feel free to merge.

src/python/dxpy/bindings/dxfile_functions.py
mhrvol: I checked email with this notification after I checked/merged this PR. There is another PR with this change. Feel free to merge after approving.

PR #656, opened 2020-12-11
TIP-1872 Update jackson libraries to work well with Spark 3.0

sushanthdn:


arunbhat (changes requested):
Should we do this after we go to Spark 3.0 OR is this independent?

sushanthdn:
> Should we do this after we go to Spark 3.0 OR is this independent?

I have built the jar locally and added it as part of nucleus repo, so ideally there is no dependency as such but its good to keep both consistent.

PR #661, opened 2021-01-07
APPS-330 keep-open option for dx build

mhrvol:


src/python/dxpy/scripts/dx.py
```diff
@@ -4463,6 +4463,9 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
 # --region
 app_options.add_argument("--region", action="append", help="Enable the app in this region. This flag can be specified multiple times to enable the app in multiple regions. If --region is not specified, then the enabled region(s) will be determined by 'regionalOptions' in dxapp.json, or the project context.")
 
+# --keep-open
+build_parser.add_argument('--keep-open', help=fill("Do not close workflow after building it. Does not affect app(let)s.", width_adjustment=-24), action='store_true')
```
commandlinegirl, line 4467: --> Does not apply to app(let)s and global workflows.

commandlinegirl: Also, we throw errors about conflicts in the args here: https://github.com/dnanexus/dx-toolkit/blob/c4ca39b24564dbf77d83278125f3def86788830f/src/python/dxpy/scripts/dx.py#L2567. Please add for app(let)s and g workflows. Thanks!

commandlinegirl (changes requested):
Minor comments.

src/python/dxpy/scripts/dx.py
mhrvol: I will throw the error regarding global workflow problem here.

However how to make sure that applet is built and not workflow? If used has only dxworkflow in current directory and  runs `dx build`, workflow will be built, even without --workflow switch.

My initial idea was to just ignore the --keep-open flag in case applet/app being built.

src/python/dxpy/workflow_builder.py
```diff
@@ -453,7 +454,7 @@ def _build_underlying_workflows(enabled_regions, json_spec, args):
     try:
         for region, project in projects_by_region.items():
             json_spec['project'] = project
-            workflow_id = _build_regular_workflow(json_spec)
+            workflow_id = _build_regular_workflow(json_spec, args.keep_open)
```
commandlinegirl, line 457: Here we don't have to pass args.keep_open since we require that the underlying workflows in the global workflow are closed so the arg will always be false.

commandlinegirl (commented):
One update request.

src/python/dxpy/scripts/dx.py
commandlinegirl: Thanks!

src/python/dxpy/workflow_builder.py
mhrvol: fixed

commandlinegirl (approved):
LGTM, thanks!

PR #663, opened 2021-01-20
xattr import fix / test fixes

kpjensen:
Wrap `xattr` import in a try/catch in `dx-upload-all-outputs`. Was causing some test failures. 

Upgrade some of the dxclient test applets to Ubuntu 20.04 and Python 3. 

Fix a few dxclient workflow tests which require waiting for analysis execution and stage doc creation.


commandlinegirl (approved):
LGTM.

PR #670, opened 2021-02-22
APPS-361 - search for reference genome files proj using region

mhrvol:
Updated reference Genome files project search - searching in a certain region only

### DIFF
--- a/src/python/dxpy/cli/exec_io.py
+++ b/src/python/dxpy/cli/exec_io.py
@@ -140,8 +140,8 @@ def interactive_help(in_class, param_desc, prompt):
                 query_project = dxpy.WORKSPACE_ID
             elif opt_num == 1:
                 region = None
-                if dxpy.PROJECT_CONTEXT_ID:
-                    region = dxpy.describe(dxpy.PROJECT_CONTEXT_ID).get("region")
+                if dxpy.WORKSPACE_ID:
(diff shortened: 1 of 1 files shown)


commandlinegirl (approved):

src/python/dxpy/cli/exec_io.py
```diff
@@ -139,7 +139,10 @@ def interactive_help(in_class, param_desc, prompt):
             if opt_num == 0:
                 query_project = dxpy.WORKSPACE_ID
             elif opt_num == 1:
-                query_project = dxpy.find_one_project(name="Reference Genome Files", public=True, billed_to="org-dnanexus_apps", level="VIEW")['id']
+                region = None
+                if dxpy.PROJECT_CONTEXT_ID:
+                    region = dxpy.describe(dxpy.PROJECT_CONTEXT_ID).get("region")
```
commandlinegirl, line 144:
```suggestion
                if dxpy.WORKSPACE_ID:
                    region = dxpy.describe(dxpy.WORKSPACE_ID).get("region")
```

Actually, shouldn't it be `dxpy.WORKSPACE_ID`? if opt_num 0 is chosen, the current project is searched ('List and choose from available data in the current project'). If option 1 - we need the region from the current. project

commandlinegirl (changes requested):

PR #673, opened 2021-04-02
DEVEX-1894 Connection (socket) leak in DxFile

artbasov:
Hi, this is a follow-up to Case #00026075.

Generally, the sole problem is in the constant creation of `HttpClient` overall and specifically whenever `FileApiInputStream` empties its buffer (see [DXFile:577](https://github.com/dnanexus/dx-toolkit/blob/master/src/java/src/main/java/com/dnanexus/DXFile.java#L577) for ex).
This leads to the creation of a new connection pool and leasing a single connection. Then without closing the `HttpClient` we just leave it to GC to cleanup. Since the default protocol for `HttpClient` is `HTTP 1.1` and [it requires any connection to be persistent by default](https://tools.ietf.org/html/rfc2616#section-8.1.2) we can't rely on the server or client to terminate TCP session unless we provide `Connection: close` header explicitly and we can't rely on any timeouts (connection/read/write) due to its defaults of infinity.

It seems to me that just sharing a single `HttpClient` for the entire SDK is the most sensible way to handle this.
Essentially, this PR just moves the creation of `HttpClient` from `DXHTTPRequest` to the `DXEnvironment` and since the environment is involved in every request anyway, it might as well manage the client.

This change renders some of the getters of `DXEnvironment` (timeouts, proxy) unused, but I didn't remove them for backward compatibility. You guys should know better if they could be removed.

A couple of questions:
- This change constrains every HTTP request to a single connection pool. Which is basically an introduction of a bottleneck into the library. Due to a large number of use-cases from a mobile app just checking job statuses once in X min to 72 core servers utilizing full bandwidth to download files from DNA nexus as fast as possible I think we need to introduce a way for users to fine-tune how a connection pool or the entire HttpClient should behave. There are a few ways I know of:
  - expose `http.connection-manager.max-per-host` and `http.connection-manager.max-total` [HttpClient settings](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/examples/org/apache/http/examples/client/ClientConfiguration.java#L198) 
  - allow them to provide their own `HttpClient` instance which they could tune however they like. This also allows sharing the same `HttpClient` between multiple `DXEnvironments` (for ex: ETL between 2 projects)

### DIFF
--- a/src/java/src/main/java/com/dnanexus/DXEnvironment.java
+++ b/src/java/src/main/java/com/dnanexus/DXEnvironment.java
@@ -41,7 +41,7 @@ import com.google.common.base.Preconditions;
  * Immutable class storing configuration for selecting, authenticating to, and communicating with a
  * DNAnexus API server.
  */
-public class DXEnvironment {
+public class DXEnvironment implements AutoCloseable {
 
     /**
      * Builder class for creating DXEnvironment objects.
@@ -723,6 +723,11 @@ public class DXEnvironment {
      */
     public int getConnectionTimeout() {
         return this.connectionTimeout;
+    }
+
+    @Override
+    public void close() throws IOException {
+        httpclient.close();
     }
 
     private static final Logger LOG = LoggerFactory.getLogger(DXEnvironment.class);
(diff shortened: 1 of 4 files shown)


kpjensen:
Looks good to me. I will open a follow-up PR for configurable settings of `http.connection-manager.max-per-host` and `http.connection-manager.max-total` similar to `socketTimeout`.

kpjensen (approved):

PR #677, opened 2021-04-13
Remove very old/optional dependencies

### LINKED ISSUE #676: Do not depend on gnureadline
`gnureadline` is an optional dependency anyways, and only used for a subset of dxpy commands.  It's annoying to get `gnureadline` installed on certain machines, so it'd be nice to remove the dependency. (can always include it in the toolkit download).

jtratner:
1. Remove windows-only colorama dependency (that seems like it doesn't
matter regardless?). This version of colorama is from 2014. Makes dxpy
unable to work with other packages.
2. Remove pyreadline and gnureadline dependency (both repos say
generally not needed) - codebase is designed to not have readline
installed and still work.

While readline may still be necessary on OSX, dxpy prints out that it's
not available and all usages of readline are within try/except so it
should not actually break anything.

closes #676

### DIFF
--- a/src/python/dxpy/cli/exec_io.py
+++ b/src/python/dxpy/cli/exec_io.py
@@ -37,7 +37,10 @@ from ..compat import input, str, shlex, basestring, USING_PYTHON2
 try:
     import gnureadline as readline
 except ImportError:
-    import readline
+    try:
+        import readline
+    except ImportError:
+        pass
 ####################
 # -i Input Parsing #
 ####################
--- a/src/python/requirements.txt
+++ b/src/python/requirements.txt
@@ -4,6 +4,3 @@ python-dateutil>=2.5
 psutil>=3.3.0
 requests>=2.8.0,<2.24.0
 cryptography>=2.3,<3.0
-gnureadline==8.0.0; sys_platform == "darwin" and python_version < "3.9"
-pyreadline==2.1; sys_platform == "win32" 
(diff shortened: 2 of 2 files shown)


emiloslavsky:
We have lots of osx users and some windows users, so I'd prefer not to remove these packages as they can be useful.
We'll look into upgrading colorama to the latest version.   
colorama does not have any dependencies in its requirements.txt - how does it make dxpy not work with other packages?

geetduggal:
@emiloslavsky , @kpjensen perhaps it's possible, then, to provide an additional package like dxpy-core/lite that can be installed via pip?  I like the idea of it working 'out of the box' in general, but would like to address this use case if possible as I can't imagine @jtratner  is the first to run into this use case.

wormsik:
We have recently fully removed `gnureadline` dependency and also bumped all dependencies to the most recent versions.

There is still one exception as `websocket-client` still contains one major unresolved bug but we plan to update as soon as it will be finally resolved.

Regarding system-specific dependencies - unfortunately, we are not able to fully remove them as dx-toolkit is used on various platforms and we need to ensure we can provide the same user experience on all of them. In case you would experience any issues with dependency resolution in Poetry (which seems to handle platform markers incorrectly), you can add `dxpy` with `--platform <linux/darwin>` (https://python-poetry.org/docs/cli/#add) argument which would restrict dxpy only for your target environment.

PR #688, opened 2021-05-03
APPS-496 Cost limit added to dx run

mhrvol:


commandlinegirl (changes requested):
Please check JIRA ticket for comments.

src/python/dxpy/scripts/dx.py
```diff
@@ -4902,6 +4903,8 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
                                               "new job will appear as a typical root execution. Setting DX_RUN_DETACH "
                                               "environment variable to 1 causes this option to be set by default.",
                                               width_adjustment=-24), action='store_true')
+parser_run.add_argument('--cost-limit', help=fill("Maximum cost of the job before termination.",
```
commandlinegirl, line 4906: It can only be passed to applets (jobs) or workflows (analyses) as well?

mhrvol: It should be passed to both. 

It is set in `_get_run_input_common_fields` method, which should be called from workflows and app(let)s

mhrvol: I can do some more testing, just ot make sure

r-i-v-a (approved):

PR #689, opened 2021-05-04
PTFM-24836 Read DX_WATCH_PORT from env

kpjensen:
Set `dxpy.WATCH_PORT` if `DX_WATCH_PORT` is an env variable set in the execution environment.


ebuzek (approved):
looks good to me

PR #695, opened 2021-06-18
APPS-646 dx archive/unarchive

YuxinShi0423:


src/python/dxpy/scripts/dx.py
```diff
@@ -5636,6 +5855,94 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
...
+    description=fill('Requests for the specified set files or for the files in a single specified folder to be unarchived on the platform.' +
+    'The requested copy will eventually be transitioned over to the live state while all other copies will move over to the archival state.') + 
+    '''
+
+EXAMPLES
+
+  ''' + 
+  fill('# unarchive 3 files in project "FirstProj" with project id project-B0VK6F6gpqG6z7JGkbqQ000Q', width_adjustment=-2, subsequent_indent='  ') + 
+  '''
+  $ dx archive  FirstProj:file-B0XBQFygpqGK8ZPjbk0Q000Q FirstProj:/path/to/file1 project-B0VK6F6gpqG6z7JGkbqQ000Q:/file2
```
emiloslavsky, line 5913:
```suggestion
  $ dx unarchive  FirstProj:file-B0XBQFygpqGK8ZPjbk0Q000Q FirstProj:/path/to/file1 project-B0VK6F6gpqG6z7JGkbqQ000Q:/file2
```

src/python/dxpy/scripts/dx.py
```diff
@@ -5636,6 +5855,94 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
...
+parser_unarchive_output.add_argument('-q', '--quiet', help='Do not print extra info messages', action='store_true')
+parser_unarchive_output.add_argument(
+    '-n','--dry-run', dest='dry_run',
+    help=fill('(Default) if --dry-run is not specified,',width_adjustment=-24) + '\n' + fill('"Tagged <> files for unarchival, totalling <> GB, costing <>"',width_adjustment=-24,initial_indent='  ')  + '\n' + 
+    fill('if --dry-run is specified,',width_adjustment=-24) + '\n' +
+    fill('"Would tag <> files for unarchival, totalling <> GB, costing <>"' , width_adjustment=-24,initial_indent='  '), 
+    default=False, action='store_true')
+
+parser_unarchive.add_argument('-y', '--yes', dest='confirm', help='Do not ask for confirmation',action='store_false')
+parser_unarchive.add_argument('--no-recurse', dest='recurse',help=fill('When `path` refers to a single folder, this flag causes only files in the specified folder and not its subfolders to be archived.This flag has no impact when `path` input refers to a collection of files.', width_adjustment=-24), action='store_false')
```
emiloslavsky, line 5937:
```suggestion
parser_unarchive.add_argument('--no-recurse', dest='recurse',help=fill('When `path` refers to a single folder, this flag causes only files in the specified folder and not its subfolders to be unarchived.This flag has no impact when `path` input refers to a collection of files.', width_adjustment=-24), action='store_false')
```

src/python/dxpy/scripts/dx.py
```diff
@@ -5636,6 +5855,94 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
...
+
+parser_unarchive.add_argument(
+    'path', 
+    help=fill('List of path(s) that should be unarchived. May refer to a single folder or specify one or more files in a single project.', width_adjustment=-24),
```
emiloslavsky, line 5941:
```suggestion
    help=fill('May refer to a single folder or specify one or more files in a single project.', width_adjustment=-24),
```

src/python/dxpy/scripts/dx.py
```diff
@@ -5636,6 +5855,94 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
...
+                                               parents=[all_arg],
+                                               prog='dx unarchive'
+                                               )
+
+parser_unarchive.add_argument('--rate', help=fill('The speed at which all files in this request are unarchived.', width_adjustment=-24) + '\n'+ fill('- Azure regions: {Expedited, Standard}', width_adjustment=-24,initial_indent='  ') + '\n'+ 
+fill('- AWS regions: {Expedited, Standard, Bulk}', width_adjustment=-24,initial_indent='  '), choices=["Expedited", "Standard", "Bulk"], default="Standard")
+
+parser_unarchive_output = parser_unarchive.add_argument_group(title='Output').add_mutually_exclusive_group()
+parser_unarchive_output.add_argument('-q', '--quiet', help='Do not print extra info messages', action='store_true')
+parser_unarchive_output.add_argument(
```
emiloslavsky, line 5929: should contain
```
If -q option is specified, prints nothing. Otherwise,   if --dry-run is not specified, prints "Tagged <> files for unarchival, totalling <> GB, costing <format the value returned by api call using project's currency setting>" and
    if --dry-run is  specified, prints "Would tag <> files for unarchival, totalling <> GB, costing <>"
```

src/python/dxpy/scripts/dx.py
```diff
@@ -3787,6 +3787,225 @@ def publish(args):
...
+                    print("     Unexpected error: {}".format(format_exception(e)))
+                failed_path[project_id] = format_exception(e)
+                had_error = True
+                continue
+            if res["files"] == 0:
+                print("Failed request: {} is not unarchiving any file.".format(request_input))
+    
+    if not args.quiet:
+        if args.dry_run:
+            print(f'Would tag {total["files"]} files for unarchival, totalling {total["size"]} GB, costing {total["cost"]}')
```
emiloslavsky, line 3999: from spec <format the value returned by api call using project's currency setting>

emiloslavsky (changes requested):

jtratner:
small thought: why not drop the `--dry-run` flag, have `-y --yes` flag for both and then if you run either `archive` or `unarchive`, run `--dry-run` first if `-y` isn't specified and then prompt?  That way default is to do nothing, and normal path is interactive prompt:

```
$ dx archive project-123:/folder
Would archive 256 files totalling 500GB - continue? [Y/n] Y
Archived 256 files.
```

```
$ dx unarchive project-123:/folder
Would unarchive 256 files at a cost of $X - continue? [Y/n] Y
Unarchived 256 files...
```

or:

```
dx archive -y project-123:/folder
Archived 256 files
```

YuxinShi0423:
> small thought: why not drop the `--dry-run` flag, have `-y --yes` flag for both and then if you run either `archive` or `unarchive`, run `--dry-run` first if `-y` isn't specified and then prompt? That way default is to do nothing, and normal path is interactive prompt:
> 
> ```
> $ dx archive project-123:/folder
> Would archive 256 files totalling 500GB - continue? [Y/n] Y
> Archived 256 files.
> ```

Currently only the unarchive has the option "dry run", so it could not work for archival. However by default all the files to be archived/unarchived (including their current status) would be displayed before prompting (if not "-y" or "-n") and sending the request, so at least the user knows what it would try to process.

src/python/dxpy/scripts/dx.py
```diff
@@ -5636,6 +5801,104 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
...
+#####################################
+                               
+parser_archive = subparsers.add_parser(
+    'archive', 
+    help='Requests for the specified set files or for the files in a single specified folder to be archived on the platform', 
+    description=
+    '''
+Requests for the specified set files or for the files in a single specified folder to be archived on the platform. 
+For each file, if this is the last copy of a file to have archival requested, the full archival of the object will be triggered. 
+Otherwise, the files will be marked in an archival state denoting that archival has been requested.
```
emiloslavsky, line 5815:
```suggestion
Otherwise, the files will be placed in an archival state denoting that archival has been requested.
```

YuxinShi0423: The description here is the same as the doc [https://documentation.dnanexus.com/developer/api/data-containers/projects#api-method-project-xxxx-archive](url). Shall I update the documentation as well?

jtratner:
Even if API doesn't have `--dry-run` in it, you can still fake out `--dry-run` as you've said, more about consistency of the CLI command that you're presenting others

YuxinShi0423 (commented):
Tests went well locally on staging using the same script running on Jenkins. But not on Jenkins dx-toolkit release board since it needs an authorized user. @kpjensen any idea about that?

geetduggal:
> small thought: why not drop the `--dry-run` flag, have `-y --yes` flag for both and then if you run either `archive` or `unarchive`, run `--dry-run` first if `-y` isn't specified and then prompt? That way default is to do nothing, and normal path is interactive prompt:
> 
> ```
> $ dx archive project-123:/folder
> Would archive 256 files totalling 500GB - continue? [Y/n] Y
> Archived 256 files.
> ```
> 
> ```
> $ dx unarchive project-123:/folder
> Would unarchive 256 files at a cost of $X - continue? [Y/n] Y
> Unarchived 256 files...
> ```
> 
> or:
> 
> ```
> dx archive -y project-123:/folder
> Archived 256 files
> ```

FWIW, I like this idea since archiving is a rather large operation.  Just starting my review now, so not sure if we incorporated it, but just wanted to give my $.02 for @emiloslavsky

src/python/dxpy/scripts/dx.py
```diff
@@ -3787,6 +3787,117 @@ def publish(args):
...
+
+    # resolve paths
+    target_project = None    
+    target_folder = None
+    target_files = []
+    for path in args.path:
+        try: 
+            project_id, folderpath, file_results = resolve_existing_path(path, expected_classes=None, allow_mult=True, all_mult=args.all)
+        except ResolutionError as e:
+            err_exit('Could not resolve "' + path + '": ' + e.msg, code=3, arg_parser=parser_archive)
```
geetduggal, line 3801: (Minor: we may want to move towards using other formatting methods / string interpolation a little more as it tends to be more robust to type issues and easier to deal with more complex formatting requirements)

src/python/dxpy/scripts/dx.py
```diff
@@ -3787,6 +3787,117 @@ def publish(args):
     except:
         err_exit()
 
+def archive(args):
+    dx_archive_errors = [InvalidState, ResourceNotFound, PermissionDenied]
+
+    # resolve paths
+    target_project = None    
+    target_folder = None
+    target_files = []
```
geetduggal, line 3796: We should think about the implications of two things if we already haven't (apologies in advanced if I missed a convo on this earlier): 

- What are practical limits to the length of `target_files` if any?
- What will be the system behavior if a user e.g. writes a shell script around `dx archive` on a metric-ton of files?  Should we design some 'in-product guidance' around that case? (@emiloslavsky  interest)

src/python/dxpy/scripts/dx.py
```diff
@@ -3787,6 +3787,117 @@ def publish(args):
...
+    for path in args.path:
+        try: 
+            project_id, folderpath, file_results = resolve_existing_path(path, expected_classes=None, allow_mult=True, all_mult=args.all)
```
geetduggal, line 3799: If there are many calls to this this will certainly be expensive.  We may want to avoid calling this procedure at all and be a bit more conservative about the requirements of what is in `args.path`.  For example, we may require that this list contains either valid folder paths--as verified by a single call made at the beginning to `project-xxxx/describe` or a string of the form `project-xxxx:file-yyyy` which can be validated by a single bulk call, e.g. to `findDataObjects` or something like that.  This approach would minimize API calls made to the platform by performing bulk operations when necessary and minimize load to the platform.  Another approach would be to separate out the arguments into two groups: folders and files.  Files must always be accompanied with a 'project' to help us minimize load.  For this feature in particular I think it's important to think about the scale of the inputs because my feeling is that users tend to like to archive lots of files at a time.  It's natural to do that by folder or project, but I can see users providing lots of files in some cases and IMHO and we should either (1) address this in the design as a limit we impose and error out on the client side or (2) protect ourselves via the method described above.

src/python/test/test_dxclient.py
```diff
@@ -10254,7 +10254,191 @@ def test_update_app(self):
...
+        dxpy.api.project_remove_folder(cls.proj_unarchive_id, {"folder": cls.rootdir,"recurse":True})
+        rm_project(cls.proj_archive_id)
+        rm_project(cls.proj_unarchive_id)
 
+    @classmethod
+    def gen_uniq_fname(cls):
+        cls.counter += 1
+        return "file_{}".format(cls.counter)
+  
+    def test_archive_files(self):
```
geetduggal, line 10287: May also want to add a test case that mixes in files and folders to ensure there's no unexpected behavior when you have a mix of types.

geetduggal (commented):
Nice job @YuxinShi0423 on your work on this first bigger task :-).  I have a few comments mostly related to making sure we are being as friendly to the backend as possible. We can discuss more in our standup tomorrow.

src/python/dxpy/scripts/dx.py
jtratner, line 3801: Just nit here
```suggestion
            err_exit(f'Could not resolve {repr(path)}:  {type(e).__name__}:{e}', code=3, arg_parser=parser_archive)
```

jtratner: tho that's only if python 3.6+

YuxinShi0423:
> FWIW, I like this idea since archiving is a rather large operation. Just starting my review now, so not sure if we incorporated it, but just wanted to give my $.02 for @emiloslavsky

@geetduggal  I had codes that listing the current status of requested files before doing anything (might be something useful for archiving since it does not have dry-run)

src/python/dxpy/scripts/dx.py
YuxinShi0423: > What are practical limits to the length of target_files if any? Should we design some 'in-product guidance' around that case? (@emiloslavsky interest)
The limit is defined in the API config, which is 1000 files now. For archival the default limit is hardcoded as 1000 and going over the limit would get us an apierror "InvalidInput". For unarchival

src/python/dxpy/scripts/dx.py
YuxinShi0423: For now it allows either one folder path or array of files in one project, so I guess it would not be so expensive? 
If it is a folder path it is passed directly to the server after being resolved without extra API call; and if we have files, resolving the path calls 'describe' on each (But hopefully the user would paste 1000 files here :D ) 

Also using`resolve_existing_path` here could help if there is more than one file with the same name. In that case the user can choose whether to pick some of them.

src/python/test/test_dxclient.py
```diff
@@ -10254,7 +10254,191 @@ def test_update_app(self):
...
+        self.assertEqual(dxpy.describe(fid_subdir1)["archivalState"],"archived")
+        self.assertEqual(dxpy.describe(fid_subdir2)["archivalState"],"archived")
+
+    def test_archive_invalid_paths(self):
+        # mixed file and folder path        
+        fname1 = self.gen_uniq_fname()
+        fid1 = create_file_in_project(fname1, self.proj_archive_id,folder=self.rootdir)
+        
+        with self.assertSubprocessFailure(stderr_regexp="Expecting either a single folder or a list files for each request", exit_code=3):
+            run("dx archive {}:{} {}:{}".format(self.proj_archive_id,fid1,self.proj_archive_id,self.rootdir))
```
YuxinShi0423, line 10369: @geetduggal is this the test case you suggested?

src/python/dxpy/scripts/dx.py
emiloslavsky: * project-xxxx/archive (https://documentation.dnanexus.com/developer/api/data-containers/projects#api-method-project-xxxx-archive) does not  have a documented limit on the number of files.  If there's a limit, let's document it in project-xxxx/archive and in `dx archive` 
* Re "For unarchival, there is no default limit. And it would only unarchive the first N files and leave the rest untouched without raising an error" - not sure I'm following.  What is `N`?

src/python/dxpy/scripts/dx.py
emiloslavsky: If user supplies paths as `project-xxxx:file-yyyy` or `file-yyyy` (in which case project must be set), can we skip resolution entirely and pass paths as specified by the user to project-xxxx/archive?   We can also recommend this input format for bulk archiving calls.

src/python/test/test_dxclient.py
emiloslavsky: currently user can supply one of
* a list of file paths
* a single folder

src/python/dxpy/scripts/dx.py
YuxinShi0423: Sorry I was lazy copypasting the format from other resolving error messages. I will change that to .format string.

src/python/dxpy/scripts/dx.py
YuxinShi0423: > * project-xxxx/archive (https://documentation.dnanexus.com/developer/api/data-containers/projects#api-method-project-xxxx-archive) does not have a documented limit on the number of files.  If there's a limit, let's document it in project-xxxx/archive and in `dx archive`
The limit is coded in API server config or hardcoded in the code. I will put it in the documentation and in the cli help.
> * Re "For unarchival, there is no default limit. And it would only unarchive the first N files and leave the rest untouched without raising an error" - not sure I'm following.  What is `N`?
The 'N' is the number above (the one in config; if none then 1000)

src/python/dxpy/scripts/dx.py
YuxinShi0423: I checked the nucleus code and with the given project id
- if 'files', it will check if all inputs are valid IDs and if all are in that project
- if 'folder', it will check if the folder is in that project

My guess is we can simply prepare the inputs into 1 project-xxxx, and a list of file-xxxx or one clean/folder/path. However if proj-xxxx:/path/to/file is given we still need to resolve it.

YuxinShi0423:
> small thought: why not drop the `--dry-run` flag, have `-y --yes` flag for both and then if you run either `archive` or `unarchive`, run `--dry-run` first if `-y` isn't specified and then prompt? That way default is to do nothing, and normal path is interactive pro

As @jtratner suggested, now for both archive and unarchive, if "--y" is specified, we can first "dry run" it:
- archive: display length of input file array / total counts of files found in input folderpath
- unarchive: do dry run and show the results

Then ask for user to confirm the request

@jtratner @geetduggal @emiloslavsky feel free to comment

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,155 @@ def publish(args):
...
+                if len(path) > 2: # ignore invalid path
+                    continue
+
+                # if the input path is in the following format:
+                #   file-xxxx
+                #   /folderpath/filename or /folderpath/
+                # set the target project to be the current one
+                elif len(path) == 1:
+                    if not target_project:
+                        target_project = dxpy.WORKSPACE_ID
```
geetduggal, line 3834: `dxpy.WORKSPACE_ID` should only exist if the call is made from a running job (http://autodoc.dnanexus.com/bindings/python/current/dxpy.html).  In general, the file could not exist in the current project (`dxpy.PROJECT_CONTEXT_ID`).  My feeling is it may be best in this case to require a project ID and error out in case one isn't provided.  It's not the 'classic' way we have done things, but it is a safer, best practice that we can begin enforcing more.  We have a design project planned this quarter for how we can do this more hoistically.  @emiloslavsky interest.

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,155 @@ def publish(args):
...
+                    # name is given
+                    elif path[-2]:
+                        project_results = dxpy.find_projects(name=path[-2],describe=True)
+                        if project_results:
+                            picked_project = paginate_and_pick(project_results, 
+                                                                (lambda result: ':'.join([result['describe']['name'],result['id']])),
+                                                                )['id']
+                    # empty string: current project
+                    elif path[-2] == '':
+                        picked_project = dxpy.WORKSPACE_ID
```
geetduggal, line 3852: Same note as above.

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,155 @@ def publish(args):
...
+        elif target_files: 
+            request_input = {"files": target_files}
+        elif target_folder:
+            request_input = {"folder": target_folder, "recurse":args.recurse}
+        else:
+            err_exit("No input file/folder is found in project {}".format(target_project), 
+                            code=3)
+        return request_input
+
+    def get_archival_paths(args):
```
geetduggal, line 3816: For procedures like this, it might be nice to do some more detailed unit testing to confirm various corner cases and general expected behavior.   While I wouldn't make it a 'requirement', it would be much appreciated by myself and future us 😃

geetduggal (changes requested):
Hi @YuxinShi0423 , thanks for all your great work and I think this is getting closer to the finish lines.  Just a handful of comments.

src/python/dxpy/scripts/dx.py
YuxinShi0423: Fixed.

src/python/dxpy/scripts/dx.py
YuxinShi0423: More tests have been added :)

YuxinShi0423: Updates on getting project id: 
- If the __target project__ is not set, check the path and validate its project by calling "describe", set it as the target project
- if after processing there is neither target folder nor target files, reset the target project to NONE, so the project in the next path is resolved  
- Otherwise, keep the target project id&name
- Parse the project id/name from following paths without resolving it, and compare to the target (assuming they are all in the same project)

src/python/dxpy/scripts/dx.py
geetduggal: @emiloslavsky can you help us make a product-level decision here based on the convo above?  Currently Yuxin defaults to the project context (which makes sense under certain assumptions), but we have an alternative approach that I described above to error out if it's not provided.  There are certainly pros and cons to both and just want to make sure you have the final say there.

geetduggal: Actually, nevermind I think I see what you're doing @YuxinShi0423 but would like to clarify with you.  I'll try to make it to the standup tomorrow to discuss more on the tail end of the standup.   Just want to make sure we're on the same page. Cheers!

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,177 @@ def publish(args):
...
+        confirm_msg="\n".join(["{:>10}: {}".format(k,v) for k,v in request_input.items()])
+        print(confirm_msg)
+        
+        if mode == "archival":
+            counts = len(set(target_files)) or len(list(dxpy.find_data_objects(project=target_project,folder=target_folder,classname="file",recurse=args.recurse)))
+            print('Will request in total {} file(s) for archival in {}'.format(counts,target_project))
+        elif mode == "unarchival":
+            dryrun_request_input = dict(**request_input, dryRun=True)
+            dryrun_res = send_archive_request(target_project, dryrun_request_input, request_func)
+            print('Would tag {} file(s) for unarchival, totalling {} GB, costing ${}'.format(dryrun_res["files"], dryrun_res["size"],dryrun_res["cost"]/1000))
```
emiloslavsky, line 3943:
```suggestion
            print('Will tag {} file(s) for unarchival in {}, totalling {} GB, costing ${}'.format(dryrun_res["files"], target_project, dryrun_res["size"],dryrun_res["cost"]/1000))
```

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,177 @@ def publish(args):
...
+    # ask for confirmation if needed
+    if args.confirm and INTERACTIVE_CLI:
+        print("Will send the request:")       
+        print("{:>10}: {}".format("project",target_project))
+        confirm_msg="\n".join(["{:>10}: {}".format(k,v) for k,v in request_input.items()])
+        print(confirm_msg)
+        
+        if mode == "archival":
+            counts = len(set(target_files)) or len(list(dxpy.find_data_objects(project=target_project,folder=target_folder,classname="file",recurse=args.recurse)))
+            print('Will request in total {} file(s) for archival in {}'.format(counts,target_project))
```
emiloslavsky, line 3939:
```suggestion
            print('Will tag {} file(s) for archival in {}'.format(counts,target_project))
```

emiloslavsky: In case of a folder with a large number of files, can we just get the count of files?

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,177 @@ def publish(args):
...
+    elif mode == "unarchival":
+        archival_options = {"rate": args.rate}
+        request_func = dxpy.api.project_unarchive        
+    request_input.update(archival_options)         
+    
+    # ask for confirmation if needed
+    if args.confirm and INTERACTIVE_CLI:
+        print("Will send the request:")       
+        print("{:>10}: {}".format("project",target_project))
+        confirm_msg="\n".join(["{:>10}: {}".format(k,v) for k,v in request_input.items()])
```
emiloslavsky, line 3934: I don't think this is needed

emiloslavsky (changes requested):
mostly minor text updates.

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,158 @@ def publish(args):
...
+        
+        paths = [split_unescaped(':', path, include_empty_strings=True) for path in args.path]
+        possible_projects = set()
+        possible_objects = set()
+        for p in paths:
+            if len(p)>2: 
+                err_exit("Path '{}' is invalid. Please check the inputs or check --help for example inputs.".format(p), code=3)
+            elif len(p) == 2:
+                possible_projects.add(p[0])
+            possible_objects.add(p[-1])
```
YuxinShi0423, line 3832: Updates on getting the paths: 

- parse all paths and throws errors in there is an invalid one like "proj-xxx:mistyped:filename"
- get one single project ID as the target. If a path is given as ":file-id", assume the project to be the current project context. Return NONE if we could not find one (will throw an error in later validation).
- gather objects in the target project, resolve filenames if needed

jtratner:
What happens if you have multiple projects listed? Throw an error?

Do folders get resolved to individual files? Listed in folders key? What
about if there is both a file AND a folder at the same name in DNAnexus?

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,158 @@ def publish(args):
...
+                        choice = pick(["{} ({})".format(result['describe']['name'], result['id']) for result in project_results], allow_mult=False)
+                        proj = project_results[choice]['id']
+                except:
+                    err_exit("Cannot find project with name {}".format(proj), code=3)
+
+            if target_project and proj!= target_project:
+                err_exit("All paths must refer to files/folder in a single project, but two project ids: '{}' and '{}' are given. ".format(
+                                target_project, proj), code=3)
+            else:
+                target_project = proj
```
YuxinShi0423, line 3858: @jtratner 
- multiple projects would raise an error
- we keep extending the file array from the inputs until there is one input seen as a folder path
- later there is a validation check: if both file arrays and a folder path are non empty, it would raise an error

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,158 @@ def publish(args):
...
+                err_exit("Path '{}' is invalid. Please check the inputs or check --help for example inputs.".format(p), code=3)
+            elif len(p) == 2:
+                possible_projects.add(p[0])
+            possible_objects.add(p[-1])
+
+        # get project ID
+        for proj in possible_projects:
+            # is project ID
+            if is_container_id(proj):
+                pass
```
emiloslavsky, line 3838: should be an error

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,158 @@ def publish(args):
...
+    
+    # ask for confirmation if needed
+    if args.confirm and INTERACTIVE_CLI:
+        
+        if mode == "archival":
+            if target_files:
+                counts = len(target_files)
+                print('Will tag {} file(s) for archival in {}'.format(counts,target_project))
+            else: 
+                print('Will tag file(s) for archival in folder {}:{} {}recursively'.format(target_project, target_folder, 'non' if not args.recurse else ''))
```
emiloslavsky, line 3920:
```suggestion
                print('Will tag file(s) for archival in folder {}:{} {}recursively'.format(target_project, target_folder, 'non-' if not args.recurse else ''))
```

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,158 @@ def publish(args):
     except:
         err_exit()
 
+def archive(args):
```
emiloslavsky, line 3788: Optional: rename to archive_unarchive

src/python/dxpy/scripts/dx.py
YuxinShi0423: We have a use case like this listed in the CLI 
`    # archive all files recursively in project-B0VK6F6gpqG6z7JGkbqQ000Q
    $ dx archive project-B0VK6F6gpqG6z7JGkbqQ000Q:/
`
in this case `proj = project-B0VK6F6gpqG6z7JGkbqQ000Q` which is a container id.

emiloslavsky: Sorry - I though is_container_id only tested for 'container-xxxx' only, but I see that it also tests for project-xxxx.

src/python/dxpy/scripts/dx.py
```diff
@@ -3785,6 +3785,158 @@ def publish(args):
...
+        target_files = set()
+        
+        paths = [split_unescaped(':', path, include_empty_strings=True) for path in args.path]
+        possible_projects = set()
+        possible_objects = set()
+        for p in paths:
+            if len(p)>2: 
+                err_exit("Path '{}' is invalid. Please check the inputs or check --help for example inputs.".format(p), code=3)
+            elif len(p) == 2:
+                possible_projects.add(p[0])
```
emiloslavsky, line 3831:
```suggestion
                possible_projects.add(p[0])
            elif len(p) == 1:
              possible_projects.add('')
```

emiloslavsky (changes requested):
dx cp and dx ls seem to resolve project-unqualified input files using the PROJECT_CONTEXT_ID.   To be consistent and minimize surprises for dx users, let's resolve project-unqualified paths in `dx archive` and `dx unarchive` to be in PROJECT_CONTEXT_ID.

src/python/dxpy/scripts/dx.py
```diff
@@ -5634,6 +5786,104 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
...
+Otherwise, the file will be marked in an archival state denoting that archival has been requested.
+'''.format(BOLD('the specified set files'), BOLD('a single specified folder'), BOLD('ONE project')) +
+'''
+The input paths should be either 1 folder path or up to 1000 files, and all path(s) need to be in the same project. 
+To specify which project to use, prepend the path or ID of the file/folder with the project ID or name and a colon. 
+
+EXAMPLES:
+
+    # archive 3 files in project "FirstProj" with project id project-B0VK6F6gpqG6z7JGkbqQ000Q
+    $ dx archive FirstProj:file-B0XBQFygpqGK8ZPjbk0Q000Q FirstProj:/path/to/file1 project-B0VK6F6gpqG6z7JGkbqQ000Q:/file2
```
emiloslavsky, line 5809:
```suggestion
    $ dx archive FirstProj:file-B0XBQFygpqGK8ZPjbk0Q000Q FirstProj:/path/to/file1 project-B0VK6F6gpqG6z7JGkbqQ000Q:/file2
    
    # archive 2 files in the currently selected project. Specifying file ids saves time by avoiding file name resolution.
    $ dx select FirstProj
    $ dx archive file-A00000ygpqGK8ZPjbk0Q000Q file-B00000ygpqGK8ZPjbk0Q000Q
    
```

src/python/dxpy/scripts/dx.py
```diff
@@ -5634,6 +5791,114 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
...
+
+EXAMPLES:
+
+    # archive 3 files in project "FirstProj" with project ID project-B0VK6F6gpqG6z7JGkbqQ000Q
+    $ dx archive FirstProj:file-B0XBQFygpqGK8ZPjbk0Q000Q FirstProj:/path/to/file1 project-B0VK6F6gpqG6z7JGkbqQ000Q:/file2
+    
+    # archive 2 files in current project. 
+    # (selecting the project ahead and specifying file IDs saves time by avoiding path resolution.)
+    $ dx select FirstProj
+    $ dx archive file-A00000ygpqGK8ZPjbk0Q000Q file-B00000ygpqGK8ZPjbk0Q000Q
```
YuxinShi0423, line 5819: @emiloslavsky suggestions are added

src/python/dxpy/scripts/dx.py
```diff
@@ -5805,8 +5810,13 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
 
 EXAMPLES:
 
-    # archive 3 files in project "FirstProj" with project id project-B0VK6F6gpqG6z7JGkbqQ000Q
+    # archive 3 files in project "FirstProj" with project ID project-B0VK6F6gpqG6z7JGkbqQ000Q
     $ dx archive FirstProj:file-B0XBQFygpqGK8ZPjbk0Q000Q FirstProj:/path/to/file1 project-B0VK6F6gpqG6z7JGkbqQ000Q:/file2
+    
+    # archive 2 files in current project. 
+    # (selecting the project ahead and specifying file IDs saves time by avoiding path resolution.)
```
emiloslavsky, line 5817:
```suggestion
     # archive 2 files in a selected project. Specifying file ids saves time by avoiding file name resolution.
```

src/python/dxpy/scripts/dx.py
```diff
@@ -3832,9 +3832,14 @@ def get_archival_paths(args):
             possible_objects.add(p[-1])
 
         # get project ID
+        if not possible_projects:
+            if not dxpy.PROJECT_CONTEXT_ID:
```
emiloslavsky, line 3836: What happens with 
```
 dx select p2; 
 dx archive p1:f1 f2
``` 
Do we error out because f2 should be resolved as p2:f2 and p2 != p1?

emiloslavsky (changes requested):

src/python/dxpy/scripts/dx.py
YuxinShi0423: Yes. In this scenario, since we have project id given in some paths, here the `possible_projects = {'f1',''}` and `''` is later interpreted as p2.

emiloslavsky: my tests show that the above test case needs fixing.

src/python/dxpy/scripts/dx.py
YuxinShi0423: Fixed: `:file-xxxx` and `file-xxxx` are now assumed to be in selected project.

emiloslavsky (approved):

emiloslavsky:
@geetduggal - "dx archvie/unarchive" looks good to me - please take another look!

YuxinShi0423:
@geetduggal @kpjensen @jtratner please feel free to comment. Thanks!

geetduggal (approved):

geetduggal:
@YuxinShi0423 I know we designed it to not accept multiple project IDs.  Can you confirm either from your code or a test that:

```
dx select project-yyyy
dx archive project-xxxx:/foo bar
```
where bar is a file name in project-yyyy errors out?

Also, what is the behavior when running this command

```
dx select project-xxxx
dx archive foo foo
```
where project-xxxx contains both a folder named foo and a file named foo?

YuxinShi0423:
> @YuxinShi0423 I know we designed it to not accept multiple project IDs. Can you confirm either from your code or a test that:
> 
> ```
> dx select project-yyyy
> dx archive project-xxxx:/foo bar
> ```
> 
> where bar is a file name in project-yyyy errors out?

```
➜  dx-toolkit git:(APPS-646-dx-archive/unarchive) ✗ dx select yshi_NLR_source
Selected project yshi_NLR_source
➜  dx-toolkit git:(APPS-646-dx-archive/unarchive) ✗ dx archive yshi_NLR_dest:/test_smallfile test_smallfile
All paths must refer to files/folder in a single project, but two project ids: 'project-G46qvz80y9B3jv84BfFfYZPg' and 'project-G45QZPpK7GkK3kz04JF31Z6Z' are given.
```
> 
> Also, what is the behavior when running this command
> 
> ```
> dx select project-xxxx
> dx archive foo foo
> ```
> 
> where project-xxxx contains both a folder named foo and a file named foo?

```
➜  dx-toolkit git:(APPS-646-dx-archive/unarchive) ✗ dx mkdir test_smallfile
➜  dx-toolkit git:(APPS-646-dx-archive/unarchive) ✗ dx archive test_smallfile test_smallfile               

Tagged 1 file(s) for archival in project-G46qvz80y9B3jv84BfFfYZPg
```
since when resolving I first assume it to be a file

YuxinShi0423:
@geetduggal but there is a bug that just occurs to me:
```
➜  python git:(APPS-646-dx-archive/unarchive) ✗ dx archive -y test_smallfile :/test_smallfile/
Expecting either a single folder or a list files for each API request
➜  python git:(APPS-646-dx-archive/unarchive) ✗ dx archive -y test_smallfile :/test_smallfile 
Will tag 1 file(s) for archival in project-G46qvz80y9B3jv84BfFfYZPg
Confirm all paths? [Y/n]: n
```
```
➜  python git:(APPS-646-dx-archive/unarchive) ✗ dx archive -y test_smallfile /test_smallfile/ 
Will tag file(s) for archival in folder project-G46qvz80y9B3jv84BfFfYZPg:/test_smallfile recursively
Confirm all paths? [Y/n]: n
➜  python git:(APPS-646-dx-archive/unarchive) ✗ dx archive -y test_smallfile /test_smallfile/
Expecting either a single folder or a list files for each API request
```
even when the call is exactly the same, the results are different, because the order of objects are random in the set `possible_objects`. Since once we get a folder, the rest of the objects would be skipped, sometimes we have `target_files` and `target_folder` but sometimes only the `target_folder`. 

In this case breaking the loop once a folder is found might not be a good choice. To change that we can either do some kind of sorting or go through all paths without breaking on found folder. 

Any suggestions? I will modify my test cases as well since they did pass :(

geetduggal:
> @geetduggal but there is a bug that just occurs to me:
> 
> ```
> ➜  python git:(APPS-646-dx-archive/unarchive) ✗ dx archive -y test_smallfile :/test_smallfile/        
> Expecting either a single folder or a list files for each API request
> ➜  python git:(APPS-646-dx-archive/unarchive) ✗ dx archive -y test_smallfile :/test_smallfile/
> Will tag file(s) for archival in folder project-G46qvz80y9B3jv84BfFfYZPg:/test_smallfile recursively
> Confirm all paths? [Y/n]: n
> ```
> 
> ```
> ➜  python git:(APPS-646-dx-archive/unarchive) ✗ dx archive -y test_smallfile /test_smallfile/ 
> Will tag file(s) for archival in folder project-G46qvz80y9B3jv84BfFfYZPg:/test_smallfile recursively
> Confirm all paths? [Y/n]: n
> ➜  python git:(APPS-646-dx-archive/unarchive) ✗ dx archive -y test_smallfile /test_smallfile/
> Expecting either a single folder or a list files for each API request
> ```
> 
> even when the call is exactly the same, the results are different, because the order of objects are random in the set `possible_objects`. Since once we get a folder, the rest of the objects would be skipped, sometimes we have `target_files` and `target_folder` but sometimes only the `target_folder`.
> 
> In this case breaking the loop once a folder is found might not be a good choice. To change that we can either do some kind of sorting or go through all paths without breaking on found folder.
> 
> Any suggestions? I will modify my test cases as well since they did pass :(

Thanks @YuxinShi0423! At this point these are more-or-less corner cases that I wouldn't want to have block getting the functionality out for integration testing with the rest of the toolkit tests on staging.  My suggestion is to address these corner cases in a separate PR and discuss the behavior with @emiloslavsky.  This will allow the core functionality to go out with the toolkit release @kpjensen will do and we can discuss this separately.  Spiritually, from my perspective there are two broad ways to handle these kinds of cases:

1.  Error early (e.g. in the Client).  Advantage: you can provide more detailed explanation to the user as to what went wrong.  Disadvantages: user may need duplicate logic across clients if working e.g. in a different language
2. Error deep (e.g. in the API).  Advantage: more unified behavior across clients. Disadvantages: may not always be applicable/easy to deliver clear message to client, may end up in a weird state if API isn't idempotent depending on how 'deep' you throw an error. 

Practically, most things end up being a mix. In this case, since we're already doing a bit of error checking  on the client side and an important goal is to protect the user against accidental issues when causing the API directly, I think simply addressing the cases in the most simple way possible on the client side would be appropriate.

emiloslavsky:
The following should give an error as at most one folder can be given as an input argument:
- `dx archive folder1  file1` 
- `dx archive folder1 folder2` 
-`dx upload x; dx mkdir x; dx archive x x/` (note the trailing slash indicating a folder).
@YuxinShi0423 please update the pull request with the fix

YuxinShi0423:
1. equivalent folder
```
➜ dx archive -y :/foo/ :foo/ /foo/ foo/
Will tag file(s) for archival in folder project-G4B26p005pGF60Jj4K5vf8Q0:/foo recursively
➜ dx archive -y foo/ foo// //foo/
Will tag file(s) for archival in folder project-G4B26p005pGF60Jj4K5vf8Q0:/foo recursively
```
2. folders (regardless of files)
```
➜ dx archive -y foo/ foo bar/ bar
Only one folder is allowed for each request. Please check the inputs or check --help for example inputs.
```
3. file & folder
```
➜ dx archive -y foo foo/                                                  
Expecting either a single folder or a list of files for each API request
➜ dx archive -y foo/ foo         
Expecting either a single folder or a list of files for each API request
```
4. equivalent filename
```
dx archive -y foo :foo /foo :/foo
Will tag 1 file(s) for archival in project-G4B26p005pGF60Jj4K5vf8Q0
```
5. Non-recursive search: need to declare path if filename is the same
```
# we have 3 files: /bar and 2 /bar/bar
➜ dx archive -y bar -a
Will tag 1 file(s) for archival in project-G4B26p005pGF60Jj4K5vf8Q0
➜ dx archive -y bar/ -a
Will tag file(s) for archival in folder project-G4B26p005pGF60Jj4K5vf8Q0:/bar recursively
➜ dx archive -y bar/bar -a
Will tag 2 file(s) for archival in project-G4B26p005pGF60Jj4K5vf8Q0
➜ dx archive -y bar bar/bar -a
Will tag 3 file(s) for archival in project-G4B26p005pGF60Jj4K5vf8Q0
➜ dx archive -y bar bar/bar   
0) bar (file-G4B3j3805pGP17bx4JkQkqPp)
1) bar (file-G4B3j2005pG0B28F4K007Q3z)
Pick a numbered choice or "*" for all: 
```

src/python/dxpy/scripts/dx.py
```diff
@@ -3823,17 +3829,26 @@ def get_archival_paths(args):
...
-                err_exit("Path '{}' is invalid. Please check the inputs or check --help for example inputs.".format(p), code=3)
+                err_exit("Path '{}' is invalid. Please check the inputs or check --help for example inputs.".format(":".join(p)), code=3)
             elif len(p) == 2:
                 possible_projects.add(p[0])
             elif len(p) == 1:
                 possible_projects.add('')
             
-            possible_objects.add(p[-1])
+            obj = p[-1]
+            if obj[-1] == '/':
```
emiloslavsky, line 3845: Can you make resolution of path work in way that `dx ls` resolves it's paths? 
```
$ dx ls -la x\\/
closed  2021-08-12 15:23:39 0 bytes   x/ (file-G4BYkyQ0Fgk8672vB7fKjqPJ)
$ dx archive x\\/
path ['x\\/']
Failed request: {'folder': '/', 'recurse': True, 'allCopies': False}
```

src/python/dxpy/scripts/dx.py
```diff
@@ -5634,6 +5802,112 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
...
+    parents=[all_arg],
+    prog='dx unarchive')
+
+parser_unarchive.add_argument('--rate', help=fill('The speed at which all files in this request are unarchived.', width_adjustment=-24) + '\n'+ fill('- Azure regions: {Expedited, Standard}', width_adjustment=-24,initial_indent='  ') + '\n'+ 
+fill('- AWS regions: {Expedited, Standard, Bulk}', width_adjustment=-24,initial_indent='  '), choices=["Expedited", "Standard", "Bulk"], default="Standard")
+
+parser_unarchive.add_argument('-q', '--quiet', help='Do not print extra info messages', action='store_true')
+parser_unarchive.add_argument(
+    '-y','--confirm', dest='confirm',
+    help=fill('Do not ask for confirmation. If not specified, dry run will be skipped' , width_adjustment=-24),
```
emiloslavsky, line 5899: should say 'Do not ask for confirmation.'

emiloslavsky: Also, should be -y or --yes, and should have 'store_false' as in 
https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/scripts/dx.py#L5002

```
parser_rmproject.add_argument('-y', '--yes', dest='confirm', help='Do not ask for confirmation', action='store_false')
```

emiloslavsky: similarly for archive

YuxinShi0423: Fixed. Thanks for catching that error :)

src/python/dxpy/scripts/dx.py
YuxinShi0423:
```
➜  dx archive x\\/       
Will tag 1 file(s) for archival in project-G4B26p005pGF60Jj4K5vf8Q0
Confirm all paths? [Y/n]: n
➜  dx archive x/  
Will tag file(s) for archival in folder project-G4B26p005pGF60Jj4K5vf8Q0:/x recursively
Confirm all paths? [Y/n]: n
```

emiloslavsky (approved):

XingGemini:
@YuxinShi0423 @eugenemi  Congrats.

PR #696, opened 2021-06-28
APPS-650 Ensure parts are in the completed state when waiting on upload

mhrvol:
In case any part is not in successful state after upload, file will be re-uploaded.

I was not able to reproduce this exact issue. Note that this does not solve problem with timeouts (slow internet connection) && cancellations - in these cases, file will be still stuck in open state, as 'close file' is not called at all.


src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -537,33 +500,89 @@ def report_progress(handler, num_bytes):
...
+        if show_progress:
+            sys.stderr.write("\n")
+            sys.stderr.flush()
+        try:
+            handler.wait_until_parts_uploaded()
+            if filename is not None:
+                fd.close()
+
+            break
+        except DXError:
```
mhrvol, line 570: Whats happening here?

In case a part did not finish successfully - it is not in completed state, we will try to reupload the file. I chose 2 retries.

r-i-v-a (approved):

src/python/dxpy/bindings/__init__.py
```diff
@@ -657,6 +657,33 @@ def _wait_on_close(self, timeout=3600*24*1, **kwargs):
             i += 1
             elapsed += wait
 
+    def _wait_until_parts_uploaded(self, timeout=60, **kwargs):
```
emiloslavsky, line 660: Is 60 seconds long enough to not trigger false re-uploads?

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -537,29 +497,87 @@ def report_progress(handler, num_bytes):
...
+            handler.wait_until_parts_uploaded()
+        except DXError:
+            if show_progress:
+                logger.warning("File {} was not uploaded correctly!".format(filename))
+            if retries > max_retries:
+                raise
+            if show_progress:
+                logger.warning("Retrying...({}/{})".format(retries, max_retries))
+            continue
+        if filename is not None:
```
emiloslavsky, line 578: how does this work when uploading from a stding fife (e.g. `dx upload --path x -`)?

src/python/dxpy/bindings/__init__.py
mhrvol: I agree. However I was never able to reproduce the issue where certain part is not closed or it is still "in progress of download". Everytime code goes to this function, these parts were already closed.

I was wondering whether we should just re-upload it again without even waiting, but it seems to me that it might be the case that the part would be finished in a while.

src/python/dxpy/bindings/__init__.py
```diff
@@ -657,6 +657,33 @@ def _wait_on_close(self, timeout=3600*24*1, **kwargs):
...
+        elapsed = 0
+        i = 0
+        while True:
+            parts = self.describe(fields={'parts'}, **kwargs)["parts"]
+            state = self._get_state(**kwargs)
+            if state == "closed":
+                # parts of closed files must have been uploaded successfully
+                break
+            if not parts:
```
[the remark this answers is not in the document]
mhrvol: there are no parts for empty file. I thought of this more like a safety measure, in case there is some weird state where there are no parts.

src/python/dxpy/bindings/__init__.py
```diff
@@ -657,6 +657,33 @@ def _wait_on_close(self, timeout=3600*24*1, **kwargs):
...
+                break
+            else:
+                is_uploaded = True
+                for key in parts:
+                    if parts[key].get(state, 'complete') != 'complete':
```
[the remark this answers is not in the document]
mhrvol: I thought the same, I did not want to leave the possibility of the code failing due to key not existing. If this is not possible, I can remove this.

src/python/dxpy/bindings/dxfile_functions.py
mhrvol: I tried `dx upload /dev/stdin` which worked. I guess this is the same thing, right?

src/python/dxpy/bindings/__init__.py
geetduggal: I share @emiloslavsky's concern on this one.  Part sizes can get pretty big so I would suggest a significantly longer timeout coupled with perhaps a ceilinged exponential backoff.  My guess is it will be hard to get historical metrics at the part-granularity so we will have to do judgement calls here.  We want to guard against releasing this and having uploads fail _en masse_ for many users.

geetduggal: (Actually, I can query to see if logs on our file closers can give an idea here)

src/python/dxpy/bindings/__init__.py
```diff
@@ -657,6 +657,33 @@ def _wait_on_close(self, timeout=3600*24*1, **kwargs):
...
+                        is_uploaded = False
+                        break
+                if is_uploaded:
+                    break
+            if elapsed >= timeout or elapsed < 0:
```
geetduggal, line 679: Consider the case where is_uploaded is true but we happened to exceed the timeout

src/python/dxpy/bindings/__init__.py
```diff
@@ -657,6 +657,33 @@ def _wait_on_close(self, timeout=3600*24*1, **kwargs):
...
+                raise DXError("Reached timeout while waiting for parts of the file ({}) to be uploaded".format(self.get_id()))
+
+            wait = min(2**7, 2**i)
```
geetduggal, line 682: Consider having n=11 as opposed n=7 for this.  This an important timeout because it's a global limit on the upload file (1+2+4+8+... geometric series). Setting n=11 will result in this closed form sum for the geometric series: `((1-2^(11+1))/(1-2))/60` which is an hour and change.  That seems reasonable, but I feel like we need real data to base this on before landing it.  We may want to even expose this parameter for cases of really large files.  I will chat with the data team on this more.

src/python/dxpy/bindings/__init__.py
```diff
@@ -657,6 +657,33 @@ def _wait_on_close(self, timeout=3600*24*1, **kwargs):
...
+            parts = self.describe(fields={'parts'}, **kwargs)["parts"]
+            state = self._get_state(**kwargs)
+            if state == "closed":
+                # parts of closed files must have been uploaded successfully
+                break
+            if not parts:
+                break
+            else:
+                is_uploaded = True
+                for key in parts:
```
geetduggal, line 673: (Minor: consider alternatively using the `any` procedure in Python`

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -510,15 +472,13 @@ def read(num_bytes):
         # If file cannot be mmap'd (e.g. is stdin, or a fifo), fall back
         # to doing an actual read from the file.
         if not can_be_mmapd(fd):
-            return fd.read(handler._write_bufsize)
+            return fd.read(num_bytes)
```
geetduggal, line 475: What motivated this change?

geetduggal (commented):
Nice job @mhrvol!  Some comments related mostly to timeouts.  I'll do some research on my end as well.  We should also define a clear test plan for this one.

src/python/dxpy/bindings/__init__.py
geetduggal: Got it. Thanks!

src/python/dxpy/bindings/__init__.py
geetduggal: Sorry for my misunderstanding.  Yes, a smaller timeout makes sense perhaps setting `timeout` to  something like `((1-2^(7+1))/(1-2))` so it will allow all 7 retries to occur.

geetduggal (approved):
LGTM modulo minor comments. Thanks!

emiloslavsky:
@mhrvol - can this be merged?

mhrvol:
> @mhrvol - can this be merged?

I am meeting with Kurt today, there's a weird cache issue.  @emiloslavsky

PR #679, opened 2021-04-20
APPS-495 Add low priority for dx run

mhrvol:

PR #697, opened 2021-06-29
APPS-632 Adding currency formatting based on provided metadata

jkotrs:
https://jira.internal.dnanexus.com/browse/APPS-632

### DIFF
--- a/src/python/dxpy/utils/describe.py
+++ b/src/python/dxpy/utils/describe.py
@@ -908,7 +908,7 @@ def print_execution_desc(desc):
                 else:
                     print_nofill_field(" sys reqs", YELLOW() + json.dumps(cloned_sys_reqs) + ENDC())
     if not desc.get('isFree') and desc.get('totalPrice') is not None:
-        print_field('Total Price', "$%.2f" % desc['totalPrice'])
+        print_field('Total Price', print_currency(desc['totalPrice'], desc['currency']))
     if desc.get('invoiceMetadata'):
         print_json_field("Invoice Metadata", desc['invoiceMetadata'])
     if desc.get('sshHostKey'):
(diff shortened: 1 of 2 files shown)


mhrvol (commented):
LGTM, but update the changelog - add change description to unreleased - an example of this can be seen here:
https://github.com/dnanexus/dx-toolkit/pull/679/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

jkotrs:
Thanks Marek, will do

mhrvol (approved):

src/python/dxpy/utils/describe.py
```diff
@@ -918,6 +918,14 @@ def print_execution_desc(desc):
         if field not in recognized_fields:
             print_json_field(field, desc[field])
 
+def print_currency(value, meta):
+    prefix = '-' if value < 0 else ''
+    prefix += meta['symbol'] if meta['symbolPosition'] == 'left' else ''
+    suffix = f' {meta["symbol"]}' if meta['symbolPosition'] == 'right' else ''
```
kpjensen, line 924: Are f strings supported in Python 2.7? If not, we probably need to change to `str.format()` here for compatibility.  

We have not yet officially deprecated Python 2.7 support, still building for the 16.04 job execution environment and tarball published to the Downloads page in our docs.

kpjensen (changes requested):

src/python/dxpy/utils/describe.py
jkotrs: I see, I assumed we've moved over to python 3+ based on the commend in README.md (I think it states April 2021 as end of support). I can rewrite to 2.7 np

r-i-v-a (changes requested):
Pending changes from Kurt's comment

r-i-v-a (approved):

kpjensen (approved):

PR #699, opened 2021-07-07
Apps-554 set globalworkflow project

YuxinShi0423:
This PR fixed the issue that when `dx run <globalworkflow> --project project-XXXXXX` does not submit the global workflow to the target project. Similar problems are observed when using flags '--destination', '--folder'.

It is caused by missing the 'project' argument when updating run inputs for the underlying workflow, and it is now fixed by bringing the 'project' argument back.

[APPS-554: Submitting globalworkflow does not respect project flag](https://jira.internal.dnanexus.com/browse/APPS-554)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 * `Total price` for `dx describe` prints formatted currency based on `currency` metadata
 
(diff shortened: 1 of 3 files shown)


mhrvol (changes requested):
LGTM, just update the changelog

mhrvol (approved):

r-i-v-a (approved):

PR #703, opened 2021-07-21
APPS-624 build globalworkflow from workflow

YuxinShi0423:
This PR is to support building global workflow from workflow (both wdl and non-wdl ones).

Requirements included:

- Detect that local workflow is built with dxCompiler or dxWDL of an older version. If so, refuse to create a global workflow
- Notify the global workflow builder to check workflow description for the dependencies and share them with authorized users when detecting a dxCompiler-build local workflow with dependencies. It should work for non-WDL workflows as well
- Global workflow description should be cloned from the local workflow description, including the dependencies listing generated by the dxCompiler


src/python/dxpy/workflow_builder.py
```diff
@@ -68,6 +68,33 @@ def _parse_executable_spec(src_dir, json_file_name, parser):
...
+def _check_dxcompile_version(json_spec):
+    if "tags" in json_spec and "dxCompiler" in json_spec["tags"]:
+        if "details" in json_spec and "version" in json_spec["details"]:
+            import requests
+            from distutils.version import StrictVersion
+            url = 'https://github.com/dnanexus/dxCompiler/releases/latest'
+            r = requests.get(url)
+            latest_version = r.url.split('/')[-1]
+            if StrictVersion(json_spec["details"]["version"]) < StrictVersion(latest_version):
+                raise WorkflowBuilderException("Workflow spec is not compiled using latest dxCompiler")
```
r-i-v-a, line 97: Eventually edit message to dxCompiler >= version when we know what version

src/python/dxpy/workflow_builder.py
```diff
@@ -346,11 +396,29 @@ def _assert_executable_regions_match(workflow_enabled_regions, workflow_spec):
...
+        
+        if billable_regions:
+            workflow_enabled_regions.intersection_update(billable_regions)
+    
     for exect in executables:
-
-        if exect.startswith("applet-") and len(workflow_enabled_regions) > 1:
-            raise WorkflowBuilderException("Building a global workflow with applets in more than one region is not yet supported.")
+        if exect.startswith("applet-"):
+            if len(workflow_enabled_regions) > 1:
```
r-i-v-a, line 416: This same thing is getting checked multiple times, for each executable?

src/python/dxpy/workflow_builder.py
```diff
@@ -68,6 +68,33 @@ def _parse_executable_spec(src_dir, json_file_name, parser):
...
+    clean_json = re.sub('"\w*": null,','',json.dumps(json_spec))
+    
+    return json.loads(clean_json)
+
+def _check_dxcompile_version(json_spec):
+    if "tags" in json_spec and "dxCompiler" in json_spec["tags"]:
+        if "details" in json_spec and "version" in json_spec["details"]:
+            import requests
+            from distutils.version import StrictVersion
+            url = 'https://github.com/dnanexus/dxCompiler/releases/latest'
```
r-i-v-a, line 93: You don't want to always enforce it's the latest version of dxCompiler -- that is going to change over time. The version to compare against will be 1 specific version (the release of dxCompiler where all features on that side are implemented.) Maybe set a placeholder number here with a comment to update it later, like 100.100.100

src/python/dxpy/workflow_builder.py
r-i-v-a: I guess it will stop on the first applet if the exception will be thrown. Probably could be more nicely written as if len(workflow_enabled_regions) > 1 and filter(check it startsWith applet-) returns a non-empty result, then throw the exception. If only 1 region, the left side of the "and" will be false, and no need to proceed to the more computationally expensive right side of the "and."

src/python/dxpy/workflow_builder.py
```diff
@@ -502,7 +576,7 @@ def _build_global_workflow(json_spec, args):
 
         # Create a new global workflow version on the platform
         global_workflow_id = dxpy.api.global_workflow_new(gwf_final_json)["id"]
```
r-i-v-a, line 578: Have to look in more detail at what happens inside that global_workflow_new() in order to review this.

src/python/dxpy/workflow_builder.py
YuxinShi0423: I changed it because it should be allowed to assign a single region where the applet is. But the old code is throwing an error.

YuxinShi0423: But if we request one region that does not contain the applet, checking the `len(workflow_enabled_regions)` could not detect the conflict.

src/python/dxpy/workflow_builder.py
```diff
@@ -383,28 +421,37 @@ def _build_regular_workflow(json_spec, keep_open=False):
...
+    # Get billable regions
+    enabled_regions = set(enabled_regions)
+    billable_regions = dxpy.executable_builder.get_permitted_regions(args.bill_to, WorkflowBuilderException)
+    enabled_regions.intersection_update(billable_regions)
+    
+    # Verify all the stages are also enabled in these regions
+    enabled_regions = _assert_executable_regions_match(enabled_regions, json_spec)
+    
     if not enabled_regions:
         raise AssertionError("This workflow should be enabled in at least one region")
```
YuxinShi0423, line 456: The enabled region(s) will be the union of the following:
- regionalOptions of source workflow spec(if exists)
- `--region` from CLI (if given)
- the permitted regions of billTo
- the dependent executables listed in the source workflow's description.

src/python/dxpy/workflow_builder.py
```diff
@@ -383,32 +418,47 @@ def _build_regular_workflow(json_spec, keep_open=False):
...
+    # Determine in which regions the global workflow are requested to be available
     enabled_regions = dxpy.executable_builder.get_enabled_regions('globalworkflow',
                                                                   json_spec,
-                                                                  from_command_line,
+                                                                  args.region,
                                                                   WorkflowBuilderException)
     if not enabled_regions:
         enabled_regions = []
-        if not dxpy.WORKSPACE_ID:
+        if not dxpy.PROJECT_CONTEXT_ID:
```
YuxinShi0423, line 437: I changed it to project context id based on http://autodoc.dnanexus.com/bindings/python/current/dxpy.html since we are not running a job when building gwf. @commandlinegirl

src/python/dxpy/workflow_builder.py
YuxinShi0423: for example if the `workflow_enabled_regions` is `azure:us-west` but the `applet-xxxx` is in `aws:us-west`, the previous code will not detect it.

src/python/dxpy/workflow_builder.py
```diff
@@ -242,7 +271,7 @@ def _validate_json_for_global_workflow(json_spec, args):
     # TODO: get a confirmation before building a workflow that may be costly
     if 'name' not in json_spec:
         raise WorkflowBuilderException(
-            "dxworkflow.json contains no 'name' field, but it is required to build a global workflow")
+            "Workflow spec contains no 'name' field, but it is required to build a global workflow")
```
mhrvol, line 274: For apps we use `dxapp.json` ( in src/python/dxpy/scripts/dx_build_app.py). Maybe we should change that as well, just to stay consistent?

src/python/dxpy/workflow_builder.py
```diff
@@ -342,15 +369,22 @@ def _get_validated_json_for_build_or_update(json_spec, args):
 def _assert_executable_regions_match(workflow_enabled_regions, workflow_spec):
     """
     Check if the global workflow regions and the regions of stages (apps) match.
-    If the workflow contains any applets, the workflow can be currently enabled
+    If the workflow contains any applets, then workflow can be currently enabled
```
mhrvol, line 372:
```suggestion
    If the workflow contains any applets, then the workflow can be currently enabled
```

src/python/dxpy/executable_builder.py
```diff
@@ -98,6 +98,53 @@ def delete_temporary_projects(projects):
...
+            exception_msg='Cannot retrieve billing information for {}. Please check your access level and the org\'s billing policy.'.format(bill_to)
+    else:
+        exception_msg='The field "billTo" must be a valid ID of a user/org'
+    
+    if exception_msg:
+        raise executable_builder_exception(exception_msg)
+    
+    return bill_to
+
+def add_developer(prefixed_name, requested_developers):
```
mhrvol, line 127: Where is this function used?

src/python/dxpy/executable_builder.py
```diff
@@ -98,6 +98,53 @@ def delete_temporary_projects(projects):
...
+    assert(prefixed_name.startswith('app-') or prefixed_name.startswith('globalworkflow-'))
+
+    if prefixed_name.partition('-')[0] == 'app':
+        exception_type = dxpy.app_builder.AppBuilderException
+        add_developer_method = dxpy.api.app_add_developers
```
mhrvol, line 132: it looks like a function to add to DXExecutable etc.

src/python/dxpy/scripts/dx.py
```diff
@@ -2541,7 +2541,8 @@ def get_mode(args):
             if args._from.startswith("applet"):
                 return "app"
             elif args._from.startswith("workflow"):
-                build_parser.error('--from option with a workflow is not supported')
+                ## build_parser.error('--from option with a workflow is not supported')
```
mhrvol, line 2544: Can we remove this comment?

src/python/dxpy/executable_builder.py
```diff
@@ -118,7 +165,7 @@ def verify_developer_rights(prefixed_name):
         describe_method = dxpy.api.global_workflow_describe
         exception_msg = \
             'A global workflow with the given name already exists and you are not a developer of that workflow'
-
+     # TODO: add developer?
```
mhrvol, line 168: Do we want to add developer or remove this todo??

src/python/dxpy/workflow_builder.py
```diff
@@ -69,6 +69,32 @@ def _parse_executable_spec(src_dir, json_file_name, parser):
...
+
+def _cleanup_empty_keys(json_spec):
+    import re
+    clean_json = re.sub('\"\w*\": (null|\{\}|\"\"|\[\])(\,|)\s*','',json.dumps(json_spec)).replace(", }","}")
+    return json.loads(clean_json)
+
+def _check_dxcompiler_version(json_spec):
+    if  json_spec.get("details") and json_spec["details"].get("version"):
+        from distutils.version import StrictVersion
+        supported_version = "2.5.0"
```
r-i-v-a, line 91: Before merging, this should be updated to the actual dxCompiler version where the feature is supported (which we don't know yet until dxCompiler part is done).

src/python/dxpy/workflow_builder.py
```diff
@@ -69,6 +69,32 @@ def _parse_executable_spec(src_dir, json_file_name, parser):
...
+        current_compiler_version = str.split(json_spec["details"].get("version"),"-")[0]
+        if StrictVersion(current_compiler_version) < StrictVersion(supported_version):
+            raise WorkflowBuilderException("Source workflow {} is not compiled using dxCompiler (version>={}) that supports dependency annotation.".format(json_spec["name"], supported_version))
```
r-i-v-a, line 94:
```suggestion
            raise WorkflowBuilderException("Source workflow {} is not compiled using dxCompiler (version>={}) that supports creating global workflows.".format(json_spec["name"], supported_version))
```

r-i-v-a (changes requested):
Minor comments about doc strings. Otherwise, looks good, pending final dxCompiler supported_version to use. @YuxinShi0423

src/python/dxpy/executable_builder.py
YuxinShi0423: Deleted. Thank you :D

src/python/dxpy/workflow_builder.py
```diff
@@ -69,6 +69,46 @@ def _parse_executable_spec(src_dir, json_file_name, parser):
...
+        print("Note: {workflow} was compiled with -instanceTypeSelection=static "
+              "and will produce a global workflow that relies on instance types that may not be available to all users."
+              .format(workflow=json_spec["id"]))
+
+def _notify_dependencies(json_spec):
+     dependency_list = json_spec.get(["description"])
+     if dependency_list:
+         print("Note: {workflow} was compiled with unbundled dependencies. "
+              "Please check the workflow description and make sure access to these dependencies is provided to all authorized users."
+              .format(workflow=json_spec["id"]))
```
YuxinShi0423, line 110: @r-i-v-a @emiloslavsky as said here:https://confluence.internal.dnanexus.com/display/TTN/Global+WDL+workflows

> If we detect a dxCompiler-built local workflow with dependencies and --brief is not specified, notify the global workflow builder to check workflow description for these dependencies and share them with authorized users

So shall I just put a short message like above, or parse the description and print the whole list out?

YuxinShi0423: > Updated "details" (if static instance type) looks like
>     "details": {
>         "staticInstanceTypeSelection": true,
>     },
> If dynamic instance type selection, that field will not be in "details."

src/python/dxpy/workflow_builder.py
```diff
@@ -69,6 +69,46 @@ def _parse_executable_spec(src_dir, json_file_name, parser):
...
+            raise WorkflowBuilderException("Source workflow {} is not compiled using dxCompiler (version>={}) that supports creating global workflows.".format(json_spec["id"], supported_version))
+    else:
+        raise WorkflowBuilderException("Cannot find the dxCompiler version from the spec of the source workflow {}. Please specify the dxCompiler version in the 'details' field of the workflow spec.".format(json_spec["id"]))
+
+def _notify_instance_type_selection(json_spec):
+    is_static = json_spec["details"].get("staticInstanceTypeSelection", False)
+    if is_static:
+        print("Note: {workflow} was compiled with -instanceTypeSelection=static "
+              "and will produce a global workflow that relies on instance types that may not be available to all users."
+              .format(workflow=json_spec["id"]))
```
YuxinShi0423, line 103: > Updated "details" (if static instance type) looks like
> ```
>      "details": {
>          "staticInstanceTypeSelection": true,
>      },
> ```
> If dynamic instance type selection, that field will not be in "details."
@r-i-v-a @commandlinegirl

src/python/dxpy/workflow_builder.py
```diff
@@ -569,16 +622,32 @@ def skip_update():
...
+                json_spec = _fetch_spec_from_workflow(args, parser)
+            else:
+                json_spec = _parse_executable_spec(args.src_dir, "dxworkflow.json", parser)
+            
+            # Check if the local or source workflow is compiled by dxCompiler that supported dependency annotation
+            if json_spec.get("tag") and "dxCompiler" in json_spec["tag"]:
+                _check_dxcompiler_version(json_spec)
+                if not args.brief:
+                    _notify_instance_type_selection(json_spec)
+                    _notify_dependencies(json_spec)
```
YuxinShi0423, line 646: Here checking instance type selection and unbundled dependency only run when --brief is false, because it is just sending notifications to the user without changing the workflow.

src/python/dxpy/workflow_builder.py
r-i-v-a: Just put a short message that tells them to check the metadata details field. @YuxinShi0423

src/python/dxpy/workflow_builder.py
r-i-v-a: Yes, I think @mhrvol means the error message is worded that way when building an app from an applet. @YuxinShi0423

src/python/dxpy/workflow_builder.py
```diff
@@ -619,9 +696,8 @@ def build(args, parser):
         raise Exception("Arguments not provided")
 
     try:
-        json_spec = _parse_executable_spec(args.src_dir, "dxworkflow.json", parser)
-        workflow_id = _build_or_update_workflow(json_spec, args)
+        workflow_id = _build_or_update_workflow(args, parser)
         _print_output(workflow_id, args)
     except WorkflowBuilderException as e:
-        print("Error: %s" % (e.args,), file=sys.stderr)
+        print("Error: {}" .format(e.args,), file=sys.stderr)
```
r-i-v-a, line 702: Shouldn't it be `".format()` without a space?

YuxinShi0423:
Built globalworkflow-G6XZ1bj018Zyx444749fY0bz dependency_report_wf1-0.0.3.json from workflow-G6XY6500yzZyx444749fY0ZK
https://staging.dnanexus.com/projects/Fy9QqgQ0yzZbg9KXKP4Jz6Yq/monitor/analysis/G6XZ20j0yzZxB2597z1f0zP3

@r-i-v-a

r-i-v-a (approved):

src/python/dxpy/executable_builder.py
```diff
@@ -152,18 +179,18 @@ def verify_developer_rights(prefixed_name):
...
 
-def assert_consistent_regions(from_spec, from_command_line, builder_exception):
+def assert_consistent_regions(from_spec, from_command_line, executable_builder_exception):
     """
     Verifies the regions passed with --region CLI argument and the ones
     specified in regionalOptions are the same (if both CLI and spec were used)
     """
     if from_spec is None or from_command_line is None:
         return
     if set(from_spec) != set(from_command_line):
```
YuxinShi0423, line 189: I reverted my previous chanage, which only check if --region is a subset of regionalOptions from the json spec. It sounds more consistent with the helpstring `help="Enable the app/globalworkflow in this region. This flag can be specified multiple times to enable the app/globalworkflow in multiple regions. If --region is not specified, then the enabled region(s) will be determined by 'regionalOptions' in dxapp.json, or the project context."` , but I don't want to change the behavior here since I am not sure about the side effect when building app/gwf.

src/python/dxpy/workflow_builder.py
YuxinShi0423: I have checked `src/python/dxpy/scripts/dx_build_app.py` and all messages containing "dxapp.json" are inside the building process of apps from dxapp.json.

src/python/dxpy/app_builder.py
```diff
@@ -904,15 +904,15 @@ def _create_app(applet_or_regional_options, app_name, src_dir, publish=False, se
         if authorized_users_to_add:
             dxpy.api.app_add_authorized_users(app_id, input_params={'authorizedUsers': list(authorized_users_to_add)})
         if skip_adding_public:
-            logger.warn('the app was NOT made public as requested in the dxapp.json. To make it so, run "dx add users app-%s PUBLIC".' % (app_spec["name"],))
+            logger.warn('the app was NOT made public as requested in the app spec. To make it so, run "dx add users app-%s PUBLIC".' % (app_spec["name"],))
```
YuxinShi0423, line 907: @mhrvol @r-i-v-a the messages in app_builder are updated. Thanks for the suggestion!

src/python/dxpy/workflow_builder.py
YuxinShi0423: But in `app_builder.py` there are some messages that need to be updated. Please check the other comment.

src/python/dxpy/workflow_builder.py
```diff
@@ -325,53 +364,63 @@ def _get_validated_json_for_build_or_update(json_spec, args):
...
+        return workflow_enabled_regions
+    
     executables = [i.get("executable") for i in workflow_spec.get("stages")]
 
     for exect in executables:
-
-        if exect.startswith("applet-") and len(workflow_enabled_regions) > 1:
-            raise WorkflowBuilderException("Building a global workflow with applets in more than one region is not yet supported.")
+        if exect.startswith("applet-"):
+            applet_region = dxpy.api.applet_describe(dxpy.api.applet_describe(exect)["project"])["region"]
```
commandlinegirl, line 395: How about using dx bindings here:
```suggestion
            applet_project = dxpy.DXApplet(exect).project
            applet_region = dxpy.DXProject(applet_project).region
```

src/python/dxpy/workflow_builder.py
```diff
@@ -325,53 +364,63 @@ def _get_validated_json_for_build_or_update(json_spec, args):
...
+            if applet_region in workflow_enabled_regions:                
+                workflow_enabled_regions = {applet_region} # only one region is allowed when using applet
```
commandlinegirl, line 397: After this we can add a `break` since it's all the applets will be in the same region so no need to iterate over the remaining ones. This pertains to the workflow used in the --from option as well as a nested workflow.

src/python/dxpy/workflow_builder.py
```diff
@@ -325,53 +364,63 @@ def _get_validated_json_for_build_or_update(json_spec, args):
...
     return validated
 
 
-def _assert_executable_regions_match(workflow_enabled_regions, workflow_spec):
+def _assert_executable_regions_match(workflow_enabled_regions: set, workflow_spec):
     """
     Check if the global workflow regions and the regions of stages (apps) match.
-    If the workflow contains any applets, the workflow can be currently enabled
+    If the workflow contains any applets, then the workflow can be currently enabled
     in only one region - the region in which the applets are stored.
```
commandlinegirl, line 386: Please update this docstring to say what it returns now.

commandlinegirl (changes requested):

src/python/test/test_dxclient.py
```diff
@@ -6917,7 +6917,66 @@ def test_build_workflow_with_bill_to(self):
...
+                                                     json.dumps(dxworkflow_json))
+        with self.assertSubprocessFailure(stderr_regexp="Cannot find the dxCompiler version from the spec of the source workflow", exit_code=3):
+            run("dx build --globalworkflow --version 0.0.1 {}".format(workflow_dir))
+        
+        dxworkflow_json.update({"details": {"version":"0.0.1"}})
+        workflow_dir = self.write_workflow_directory(gwf_name,
+                                                     json.dumps(dxworkflow_json))
+        with self.assertSubprocessFailure(stderr_regexp="Source workflow {} is not compiled using dxCompiler \(version>={}\) that supports creating global workflows.".format(dxworkflow_json["name"], SUPPORTED_DXCOMPILER_VERSION), exit_code=3):
+            run("dx build --globalworkflow {}".format(workflow_dir))
+
```
commandlinegirl, line 6979: All these tests are negative tests. The positive tests will be in the dxCompiler (so as not to add the dxCompiler dependency to tests in dx-toolkit)?

r-i-v-a: Yes, I will put the positive tests in dxCompiler. APPS-627

src/python/dxpy/app_builder.py
r-i-v-a: I thought @mhrvol 's suggestion was to go the other way, and use dxapp.json / dxworkflow.json throughout? dxapp.json is used often enough in our user-facing documentation that it should be clear to the user what it refers to https://documentation.dnanexus.com/developer/apps/app-metadata , maybe clearer than calling it "app spec."

src/python/dxpy/executable_builder.py
r-i-v-a: I don't see a problem with leaving the original behavior here. The way I read it, you are meant to either use `--region` without restrictions if the JSON doesn't contain `regionalOptions`, or not use the flag if `regionalOptions` are present. In case both are used at the same time, it enforces that they match exactly to avoid ambiguity.

src/python/dxpy/app_builder.py
r-i-v-a: Ok, I defer to @commandlinegirl 's comment on the other thread, so it can stay as "app / workflow spec."

src/python/test/test_dxclient.py
```diff
@@ -6917,7 +6917,66 @@ def test_build_workflow_with_bill_to(self):
...
+        # build global workflow without specified version
+        source_wf = "workflow-0000000000000000000000NA"
+        with self.assertSubprocessFailure(stderr_regexp="--version must be specified when using the --from option", exit_code=2):
+            run("dx build --globalworkflow --from {}".format(source_wf))
+    
+    def test_build_globalworkflow_from_old_WDL_workflow(self):
+        SUPPORTED_DXCOMPILER_VERSION = "2.8.0"
+        # build global workflow from wdl workflows
+        gwf_name = "globalworkflow_build_from_wdl_workflow"
+        dxworkflow_json = dict(self.dxworkflow_spec, name=gwf_name)
```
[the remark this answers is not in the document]
YuxinShi0423: Added some comments about testing wdl wf

src/python/dxpy/workflow_builder.py
```diff
@@ -69,6 +69,48 @@ def _parse_executable_spec(src_dir, json_file_name, parser):
...
+
+    return json_spec
+
+def _cleanup_empty_keys(json_spec):
+    import re
+    clean_json = re.sub('\"\w*\": (null|\{\}|\"\"|\[\])(\,|)\s*','',json.dumps(json_spec)).replace(", }","}")
+    return json.loads(clean_json)
+
+def _check_dxcompiler_version(json_spec):
+    SUPPORTED_DXCOMPILER_VERSION = "2.8.0"
```
YuxinShi0423, line 89: @commandlinegirl the version number is moved to the top of the function.

src/python/dxpy/workflow_builder.py
```diff
@@ -341,37 +381,48 @@ def _get_validated_json_for_build_or_update(json_spec, args):
 
 def _assert_executable_regions_match(workflow_enabled_regions, workflow_spec):
     """
-    Check if the global workflow regions and the regions of stages (apps) match.
-    If the workflow contains any applets, the workflow can be currently enabled
+    Check if the dependent apps/applets/subworkflows in the workflow are enabled in requested regions
+    Returns the subset of requested regions where all the dependent executables are enabled
+    If the workflow contains any applets, then the workflow can be currently enabled
```
YuxinShi0423, line 386: The docstring is updated @commandlinegirl

src/python/dxpy/workflow_builder.py
commandlinegirl: Thanks

src/python/dxpy/executable_builder.py
commandlinegirl: After some consideration and discussions with Yuxin (and her investigation), we've decided it's best to tackle `--region` separately and its behavior won't be changed here.

commandlinegirl (approved):
LGTM, thank you!

PR #708, opened 2021-07-29
PTFM-28916 Client token test fix

ocbrewing:
authserver newAuthToken no longer allows getting a token while only
supplying a token. This commit updates the file upload test to just
use the full scope token rather than creating a file upload only one.


ocbrewing:
Reverted the authserver commit that would have necessitated this change..

PR #709, opened 2021-07-30
fix(symlink-download): fixes bugs and improves reliability around symlink downloading

claymcleod:
On the St. Jude Cloud project, we've experienced multiple users having trouble downloading symlinked files using the `dx` command line utility. After extensive debugging, I believe I've isolated the source of the problem. This PR attempts to make a solid improvement by eliminating the use of `wget` for downloading symlinked files.

## Description

In the way that the DNAnexus API currently works, GET-ing a download link for a symlinked file returns a 302 response with a `Location` header (see example below). The `Location` header contains a signed link that points to the actual storage location (for us, in Azure) that lasts for 5 minutes. After 5 minutes, the intent is that you should rerequest the original URI and receive an updated signed link. This is consistent with the [spec](https://datatracker.ietf.org/doc/html/rfc2068#section-10.3.3), and I personally feel it's the safest too (signed links that expire quickly).

```HTTP/1.1 302 Moved Temporarily
...
Location: [SIGNED URL THAT LASTS 5 MINUTES]
Server: nginx/1.18.0

...
```

Unfortunately, it appears that some versions of `wget` do not follow this process effectively. I'm still not sure whether this is because `wget` doesn’t support going back to the original request URI after 5 minutes or not, but [this StackOverflow question](https://stackoverflow.com/questions/38273828/make-wget-retry-original-url-after-3xx-redirect) suggests that there are conditions under which it does not work correctly. This has been my experience as well for various users of St. Jude Cloud.

## Solution

Thus, I'm proposing we just limit the downloading of symlinks to `aria2c` and improve some of the options around this. `aria2c` should be available on all DNAnexus nodes (as far as I am aware), and I think it's reasonable to ask users to install this if the tradeoff is that they have the best possible experience downloading. Again, I don't think it's a clear-cut case that we couldn't fix using `wget` or even introduce `curl`, but I didn't consider that within scope as it could be a rabbit hole of time invested.

If someone from DNAnexus wants to (a) try to get it to work with `wget` or `curl` more effectively or (b) just support the in-tool downloading of symlinks as any other "normal" file in DNAnexus, then feel free to reject this PR.

## Details

- Removes `wget` altogether, and clarifies how to install `aria2c`.
- Removes `--check-certificate=false`, I'm not sure why we would have introduced this in the first place and considered it unneccesarily insecure.
- I've moved the check to ensure `aria2c` is in the `$PATH` to the front of the download. This ensures we don't generate any signed links if `aria2c` isn't installed. Just an extra safety measure.
- Introduced `--symlink-unlimited-retries` option. This will set `-m 0` on `aria2c`, translating to unlimited retries on the download.
- Added missing `-c` argument on `aria2c` command that is generated to continue existing downloads.
- Given the removal of `wget`, the new `aria2c` arguments, and the `--symlink-unlimited-retries` option, I believe the entire structure around the `call_cmd` function is unnecessary now.
  - First, I don't believe this was operating as intended fully regardless. Exit code 22 only applies to `aria2c` ([exit codes for wget](https://www.gnu.org/software/wget/manual/html_node/Exit-Status.html)), and it was being thrown when `aria2c` encountered a 403 response. The 403 was being thrown when the signature on the signed link expired, so it wasn't a scenario where a back-off timer would help the situation.
  - Second, with the new options, this can be handled much more sanely by exposing arguments that use the facilities provided by `aria2c` to accomplish this goal.
- Remove the print about following redirect (most users are confused as to what this means or if things are working correctly in my experience).

## Assumptions

- `aria2c` should be available on every DNAnexus node (need someone from DNAnexus to confirm in the comments).
  - Confirmed that `aria2c` is supported in the default image and there are no plans to remove it by Geet through email on July 30, 2021.

## TODO

- I cannot get the tests to work on my local machine, even on `master`. So I'm hoping there is CI set up where the environment is configured correctly to get them to run. After that, I possibly need to update them.

FWIW, this is the error I'm getting.

```
❯ python build/run_python_integration_tests.py --tests test_dx_symlink
test_symlinks (test_dx_symlink.TestSymlink) ... ERROR

======================================================================
ERROR: test_symlinks (test_dx_symlink.TestSymlink)
----------------------------------------------------------------------
...
PermissionDenied: Cannot create a new project as an anonymous user, code 401. Request Time=1627666473.16, Request ID=1627666473307-400755

----------------------------------------------------------------------
Ran 1 test in 0.169s

FAILED (errors=1)
*** unittest invocation failed with code 1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 ### Fixed
 
 * `Total price` for `dx describe` prints formatted currency based on `currency` metadata
+* Improvements to symlink downloading reliability by solely using `aria2c` and enhancing various options around its use
 
 ### Fixed
 * `dx run <globalworkflow> --project/--destination/--folder` now submits analysis to given project or path
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -97,7 +97,7 @@ def new_dxfile(mode=None, write_buffer_size=dxfile.DEFAULT_BUFFER_SIZE, expected
 
 
 def download_dxfile(dxid, filename, chunksize=dxfile.DEFAULT_BUFFER_SIZE, append=False, show_progress=False,
-                    project=None, describe_output=None, **kwargs):
+                    project=None, describe_output=None, symlink_unlimited_retries=False, **kwargs):
     '''
     :param dxid: DNAnexus file ID or DXFile (file handler) object
     :type dxid: string or DXFile
@@ -114,6 +114,9 @@ def download_dxfile(dxid, filename, chunksize=dxfile.DEFAULT_BUFFER_SIZE, append
             It should contain the default fields of the describe API call output and
             the "parts" field, not included in the output by default.
     :type describe_output: dict or None
+    :param symlink_unlimited_retries: when downloading a symlink, call aria2c with the 
+            unlimited retries parameter specified.
+    :type symlink_unlimited_retries: bool
 
     Downloads the remote file referenced by *dxid* and saves it to *filename*.
 
@@ -134,6 +137,7 @@ def download_dxfile(dxid, filename, chunksize=dxfile.DEFAULT_BUFFER_SIZE, append
                                    show_progress=show_progress,
                                    project=project,
                                    describe_output=describe_output,
+                                   symlink_unlimited_retries=symlink_unlimited_retries,
                                    **kwargs)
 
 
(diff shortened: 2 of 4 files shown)


claymcleod:
Would someone be able to advise on whether this is a reasonable approach and also on how to get my tests working?

cc @kpjensen @geetduggal

geetduggal:
Thanks @claymcleod! @kpjensen could you review this code from an engineering perspective and @emiloslavsky the description from a product?

emiloslavsky:
- @claymcleod 
I'd like to test wget before eliminating it from dx-toolkit as I feel it's more more commonly available than aria2c.
Could you share a project containing 0.1, 1G and 10G files backed by symlink with org-support, user-emiai and user-kjensen so we can test this out.
- I'm fine with removing `--check-certificate=false` (unless it breaks existing use cases).
- re: --symlink-unlimited-retries: generally unlimited retries may lead to hanging jobs, so I'd prefer `--symlink-max-tries`

claymcleod:
I understand your thoughts about wget, and I’m happy to turn it over to someone at DNAnexus if it will get prioritized to get fixed.

I originally made this PR because we run into these issues often in production—I’m happy if there is a different fix, but it needs to cover the older versions of wget to fix these issues in practice. Thus, I think we need _some_ code change to handle these cases.

Unfortunately, I can’t provide these since they are protected access data. You can grab HTSeq files from the Genomics Platform data browser, and that should give you some smaller files. For the larger files, you should be able to link them up using Azure store with random data. Note that the issues above will only be apparent for files that take longer than the 5 minute timeout.

I’m okay with someone changing to “max tries”.

claymcleod:
> I understand your thoughts about wget, and I’m happy to turn it over to someone at DNAnexus if it will get prioritized to get fixed.
> 
> I originally made this PR because we run into these issues often in production—I’m happy if there is a different fix, but it needs to cover the older versions of wget to fix these issues in practice. Thus, I think we need _some_ code change to handle these cases.
> 
> Unfortunately, I can’t provide these since they are protected access data. You can grab HTSeq files from the Genomics Platform data browser, and that should give you some smaller files. For the larger files, you should be able to link them up using Azure store with random data. Note that the issues above will only be apparent for files that take longer than the 5 minute timeout.
> 
> I’m okay with someone changing to “max tries”.

Even better, I would love the native downloading within the DX toolkit to work for symlinks rather than shelling out to other command line tools.

claymcleod:
Hi all,

Just following up on this to see if you've had a chance to look at the PR, we're still experiencing these issues in production. @kpjensen

emiloslavsky:
Planning on integrating this change in the next few weeks.

mhrvol:
This was merged in PR https://github.com/dnanexus/dx-toolkit/pull/773 (with some changes.)

emiloslavsky:
@claymcleod  - [dxpy 324.1](https://pypi.org/project/dxpy/0.324.1/) has the fix.

claymcleod:
That is really great news. Thank you @emiloslavsky!

PR #717, opened 2021-10-07
APPS-837 dx get app resources from permitted region

YuxinShi0423:


src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -140,46 +151,60 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
+                    break
+        # Reversing the order of the asset_depends[] so that original order is maintained
+        asset_depends[region].reverse()
     # resources/ directory
     created_resources_directory = False
     if not omit_resources:
-        for dep in info["runSpec"]["bundledDepends"]:
-            if dep in deps_to_remove:
+        if not enabled_regions:
+            raise DXError("Cannot download resources of the requested executable {} since it is not available in any of the billable regions.".format(info["name"]))
```
commandlinegirl, line 184: How about adding something like: "You can run use the omit-resources flag to skip downloading the resources"

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -140,46 +151,60 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
-                                              "version": asset_record.describe(fields={"properties": True}
-                                                                               )["properties"]["version"]
-                                              }
+                        # asset_desc = dxpy.api.record_describe(asset_record_id, {"properties": True})
+                        # asset_json = {"name": asset_desc.get("name"),
+                        #             "project": asset_desc.get("project"),
+                        #             "folder": asset_desc.get("folder"),
+                        #             "version": asset_desc.get("properties", {}).get("version")
+                        #             }
+                        asset_json={"id": asset_record_id}
```
commandlinegirl, line 167: Nice, thanks. Does it work when you "dx get app-swiss-army-knife"? Currently it's pointing to the project ID which I have selected (instead of the record's project) so using record ID should fix it.

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -140,46 +151,60 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
-            asset_record_id = file_handle.get_properties().get("AssetBundle")
-            asset_record = None
-            if asset_record_id:
-                asset_record = dxpy.DXRecord(asset_record_id)
-                if asset_record:
+    for region in info["runSpec"]["bundledDependsByRegion"]:
+        for dep in reversed(info["runSpec"]["bundledDependsByRegion"][region]):
+            file_handle = get_handler(dep["id"])
+            if isinstance(file_handle, dxpy.DXFile):
+                asset_record_id = file_handle.get_properties().get("AssetBundle")
```
commandlinegirl, line 158: If the user doesn't have access to the region where the particular file resides, will this not fall over here? It may not be possible in such case to recreate the whole regionalOptions as it is done at L303.. 

Also, since we only need assetDepends from one region, and we always strive to use as few API calls as it is possible, how about checking the accessible region here (instead of lines 188-195) and getting the file handle and the properties of the file just once?

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -275,15 +313,22 @@ def untar_strip_leading_slash(tarfname, path):
...
+            # Add regional bundledDepends to regionalOptions
+            region_depends = dxapp_json["runSpec"]["bundledDependsByRegion"][region]
+            region_bundle_depends = [d for d in region_depends if d["name"] not in deps_downloaded]
 
+            dxapp_json["regionalOptions"][region] = \
+                dict(systemRequirements=region_sys_reqs,
+                bundledDepends=region_bundle_depends
+                )
+    # Remove "bundledDependsByRegion" and "bundledDepends" field from "runSpec".
+    # assetDepends and bundledDepends data are stroed in regionalOptions instead.
```
mhrvol, line 325:
```suggestion
    # assetDepends and bundledDepends data are stored in regionalOptions instead.
```

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -115,14 +116,42 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
+                "You can use the --omit-resources flag to skip downloading the resources".format(info["name"]))
+        else:
+            # Pick a source region. The current selected region is preferred
+            try:
+                current_region = dxpy.api.project_describe(dxpy.WORKSPACE_ID, input_params={"fields": {"region": True}})["region"]
+            except:
+                current_region = None
+
+            if current_region and current_region in enabled_regions:
+                enabled_regions = [current_region] + [r for r in enabled_regions if r != current_region]
```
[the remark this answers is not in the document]
YuxinShi0423: `set.remove()` returns None :P https://www.programiz.com/python-programming/methods/set/remove

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -115,14 +116,47 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
         Writes the string `code` into a file at the relative path
         "src/<region>_<entry_point>_clusterBootstrap.<suffix>"
         """
-        script_name = "src/%s_%s_clusterBootstrap.%s" % (region, entry_point, suffix)
+        script_name = "src/{}_{}_clusterBootstrap.{}".format(region, entry_point, suffix)
         with open(script_name, "w") as f:
             f.write(code)
         return script_name
+    
+    # Get regions where the user's billTo are permitted
```
YuxinShi0423, line 124: Moved bill_to check before looping the dependency list. @commandlinegirl

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -140,48 +174,47 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
+                # if dep is not a file (record etc.), check the next dep
+                if not isinstance(file_handle, dxpy.DXFile):
+                    continue
+                
+                # check if the file is an asset dependency
+                # if so, skip downloading
+                # if failed to describe it, will try the next region
+                if file_handle.get_properties().get("AssetBundle"):
+                    deps_assets.add(dep.get("name"))
+                    continue
```
YuxinShi0423, line 207: downloading is skipped if:

- it is not a file
- it is an asset
- it has been identified as an asset in the past but failed loops

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -193,7 +226,30 @@ def untar_strip_leading_slash(tarfname, path):
...
-                deps_to_remove.append(dep)
+                # add dep name to deps_downloaded set
+                deps_downloaded.add(dep.get("name"))
+                
+            except DXError:
+                print("Download failed on dependent file with ID {} from region {}. Will try the next permitted region.\n".format(handler_id, region),
+                        file=sys.stderr)
+                # clean up deps already downloaded
+                deps_downloaded.clear()
+                created_resources_directory = False
```
YuxinShi0423, line 237: Files we already downloaded will be removed. I think it is better if all resources are retrieved from the same region just in case of version conflicts etc.

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -193,7 +226,30 @@ def untar_strip_leading_slash(tarfname, path):
...
+                shutil.rmtree("resources")
+                break
+            
+        # if all deps have been checked without an error, mark downloading as completed
+        else:
```
commandlinegirl, line 242: There is a code block between `if` and `else` statements that is not indented, which is not very pythonic (not a good software development practice in general). What is the intended corresponding `if` to the `else` here? There is a code block between `if` and `else` statements that is not indented, which is not very pythonic (not a good software development practice in general).

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -193,7 +226,30 @@ def untar_strip_leading_slash(tarfname, path):
...
+            download_completed = True                    
+    
+    # Check if downloading is completed in one of the enabled regions
+    # if so, files in deps_downloaded will not shown in dxapp.json
+    # if not, deps_downloaded is an empty set. So ID of all deps will be in dxapp.json
+    if not download_completed:
+        print("Downloading resources failed in all enabled regions. Please try downloading with their IDs in dxapp.json.")
```
commandlinegirl, line 249: How about adding: "You can also skip downloading resources entirely by using the --omit-resources flag."

src/python/dxpy/utils/executable_unbuilder.py
YuxinShi0423: Oh it is part of the `for-else`. See Line 190.

commandlinegirl: Oh, I see. This is an interesting construct. How about (generally) adding the `# no break` comment like in this explanation https://stackoverflow.com/a/23748240 to making it more clear what the `else` corresponds to?

src/python/dxpy/utils/executable_unbuilder.py
YuxinShi0423: Added.

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -140,48 +172,32 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
                     created_resources_directory = True
-                handler_id = handler.get_id()
-                fname = "resources/%s.tar.gz" % (handler_id)
+                
+                fname = "resources/{}.tar.gz" .format(handler_id)
                 download_dxfile(handler_id, fname)
-                print("Unpacking resources", file=sys.stderr)
+                print("Unpacking resource {}".format(dep.get("name")))
 
                 def untar_strip_leading_slash(tarfname, path):
```
commandlinegirl, line 202: It looks like this helper function can be defined outside of the loop.

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -115,14 +116,45 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
+            current_region = None
+
+        if current_region in enabled_regions:
+            source_region  = current_region
+            print("Trying to download resources from the current region {}...".format(source_region))
+        else:
+            source_region = enabled_regions.pop()
+            print("Trying to download resources from one of the enabled region {}...".format(source_region))
 
     # When an applet is built bundledDepends are added in the following order:
```
commandlinegirl, line 159: Please align it with the code (and place it above the code that that the comment is referring to)

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -193,7 +209,30 @@ def untar_strip_leading_slash(tarfname, path):
...
+    
+    # Check if downloading is completed in one of the enabled regions
+    # if so, files in deps_downloaded will not shown in dxapp.json
+    # if not, deps_downloaded is an empty set. So ID of all deps will be in dxapp.json
+    if not download_completed:
+        print("Downloading resources from region {} failed. "
+              "Please try downloading with their IDs in dxapp.json, "
+              "or skip downloading resources entirely by using the --omit-resources flag.".format(source_region))
+    # if anything has been downloaded
+    if deps_downloaded:
```
commandlinegirl, line 234: Have you tried `dx getting` a couple of our public apps with resources/ or bundledDepends? Do they have a lot of items in `deps_downloaded`? Are they familiar or some names of libraries that don't add to the understanding of what was downloaded? If they don't, I'd remove this new message entirely.

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -140,48 +172,32 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
-            if dep in deps_to_remove:
-                continue
-            handler = get_handler(dep["id"])
-            if isinstance(handler, dxpy.DXFile):
+        download_completed = omit_resources
+        deps_downloaded = set()
+        created_resources_directory = False
+
+        # Download resources from the source region      
+        for dep in info["runSpec"]["bundledDependsByRegion"][source_region]:
```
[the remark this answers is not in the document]
YuxinShi0423: I changed it to hit resources first, so if downloading fails we can quit it earlier. But if there is no error, it would check all deps regardless of the order we use, so the total number of calls should be the same.

src/python/dxpy/utils/executable_unbuilder.py
YuxinShi0423: Yes I have tried the ones that has bundled and asset depends (SAK for example) and most of them have one or two (except for canu, which has asset files without "AssetBundle" that link it to its record)... I will remove it since we already print out that info. Thanks :D

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -115,85 +116,119 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
+        # 1. bundledDepends explicitly specified in the dxapp.json
+        # 2. resources (contents of resources directory added as bundledDepends)
+        # 3. assetDepends (translated into bundledDepends)
+        #
+        # Therefore while translating bundledDepends to assetDepends, we are traversing the
+        # list in reverse order and exiting when we can't find the "AssetBundle" property
+        # with the tarball file.
+        #
+        # NOTE: If last item (and contiguous earlier items) of bundledDepends (#1 above) refers to an
+        # AssetBundle tarball, those items will be converted to assetDepends.
```
commandlinegirl, line 169: I think think comments needs to be updated since we are not trying to extract and reconstruct assets in dxapp.json any more.

Would you add to this comment what you say here: https://github.com/dnanexus/dx-toolkit/pull/717/files#r731166369 That assetDepends are unpacked to bundledDepends in dxapp.json?

One more thing: Did you try to build a sequence: dx get, dx build, dx get? It may be tricky and out of scope for this task to get it completely right so just let me know if this causes issues under this new design/code.

commandlinegirl (changes requested):

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -115,85 +116,119 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
-                print("Unpacking resources", file=sys.stderr)
-
-                def untar_strip_leading_slash(tarfname, path):
-                    t = tarfile.open(tarfname)
-                    for m in t.getmembers():
-                        if m.name.startswith("/"):
-                            m.name = m.name[1:]
-                        t.extract(m, path)
-                    t.close()
+                print("Unpacking resource {}".format(dep.get("name")))
```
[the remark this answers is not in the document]
YuxinShi0423: I checked the other prints and seems we send it to stderr only when an error raises? But sure I will change that. Thanks!

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -115,85 +116,113 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
+        #
+        # To get the resources, we will tranverse the bundleDepends list in the source region and do the following:
+        # - If an file ID refers to an AssetBundle tarball (a file with the "AssetBundle" property), 
+        #   skip downloading the file, and keep this file ID as a bundledDepends in the final dxapp.json
+        # - Otherwise, download the file and remove this ID from the bundledDepends list in the final dxapp.json
+
+        def untar_strip_leading_slash(tarfname, path):
+            t = tarfile.open(tarfname)
+            for m in t.getmembers():
+                if m.name.startswith("/"):
```
mhrvol, line 172: There is `lstrip` method, which removes characters from the beginning of the string if there are such characters, otherwise returns the original string. 

(`m.name.lstrip("/")` would remove all leading slashes from the name - `//string` would become `string`)

there is also `removeprefix` method, which removes an exact prefix in case it is there, but it was added only in python 3.9.

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -115,85 +116,113 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
+        # 2. resources (contents of resources directory added as bundledDepends)
+        # 3. assetDepends in the dxapp.json (with their record IDs translated into file IDs)
+        #
+        # To get the resources, we will tranverse the bundleDepends list in the source region and do the following:
+        # - If an file ID refers to an AssetBundle tarball (a file with the "AssetBundle" property), 
+        #   skip downloading the file, and keep this file ID as a bundledDepends in the final dxapp.json
+        # - Otherwise, download the file and remove this ID from the bundledDepends list in the final dxapp.json
+
+        def untar_strip_leading_slash(tarfname, path):
+            t = tarfile.open(tarfname)
```
mhrvol, line 170: I'd prefer using with  `with tarfile.open(tarfname) as t` or putting `t.close()` to `finally` clause, otherwise the  opened file may not be closed.

commandlinegirl (approved):
LGTM, thanks!

r-i-v-a (approved):

YuxinShi0423:
Manually tested with SAK: get app/applet then build with/without --omit-resources flag.

src/python/dxpy/utils/executable_unbuilder.py
```diff
@@ -115,85 +116,118 @@ def make_cluster_bootstrap_script_file(region, entry_point, code, suffix):
...
-            if isinstance(handler, dxpy.DXFile):
+        download_completed = False
+        
+        # Check if at least one region is enabled 
+        if not enabled_regions:
+            raise DXError(
+                "Cannot download resources of the requested executable {} since it is not available in any of the billable regions. "
+                "You can use the --omit-resources flag to skip downloading the resources. ".format(info["name"]))
+        
+        print("Dependencies could be retrieved from region: {}. ".format(", ".join(enabled_regions)), file=sys.stderr)
```
commandlinegirl, line 152: One additional comment. IMHO this message is not that useful since it prints what we know from the app spec. I'd remove it so that there is not too much stuff printed to the terminal

commandlinegirl (approved):

PR #720, opened 2021-10-27
TIP-2612 Update jackson libraries to work well with Spark 3.2

sushanthdn:


sushanthdn:
@arunbhat This is the change needed in dxtoolkit for Spark 3.2

arunbhat:
When will we merge this? I am guessing right at the time we put Spark3.2 in the master branch? Would dx-toolkit then be non compatible with Spark 2.4.4?

sushanthdn:
> When will we merge this? I am guessing right at the time we put Spark3.2 in the master branch? Would dx-toolkit then be non compatible with Spark 2.4.4?

Yes we should merge this only when Spark 3.2 is merged to master branch in nucleus. It does not really matter in general since we are using the built jar and adding it to nucleus.

PR #726, opened 2021-11-09
APPS-894 dx run priority with other flags

YuxinShi0423:


src/python/dxpy/scripts/dx.py
```diff
@@ -2989,6 +2989,15 @@ def run_body(args, executable, dest_proj, dest_path, preset_inputs=None, input_n
...
+    if any([args.watch or args.ssh or args.allow_ssh]):
+        if run_kwargs["priority"] in ["low", "normal"]:
+            print(fill(BOLD("WARNING") + ": You have requested that jobs be run under " +
+                    BOLD(run_kwargs["priority"]) +
+                    " priority, which may cause them to be restarted at any point, interrupting interactive work."))
+            print()
+        else:
+            run_kwargs["priority"] = "high"
+    
     if run_kwargs["priority"] in ["low", "normal"] and not args.brief:
```
YuxinShi0423, line 3001: Shall we suppress the warning when `--brief` just like below? @commandlinegirl @emiloslavsky

src/python/test/test_dxclient.py
```diff
@@ -2811,6 +2812,60 @@ def _get_analysis_id(dx_run_output):
...
+        expected_warning = dx_run_output.split('\n\n')[0]
+        ssh_job_id = dx_run_output.split('\n\n')[1].strip()
+        ssh_job_desc = dxpy.describe(ssh_job_id)
+        self.assertEqual(ssh_job_desc['applet'], applet_id)
+        self.assertEqual(ssh_job_desc['priority'], 'low')
+        for string in ["WARNING", "low", "interrupting interactive work"]:
+            self.assertIn(string, expected_warning)
+
+        # don't actually need it to run
+        run("dx terminate " + ssh_job_id)
```
YuxinShi0423, line 2849: Both ssh tests would not work due to the error below and failure to connect to aws server.

> Pseudo-terminal will not be allocated because stdin is not a terminal.

I guess the reason why the test in L2785 worked is that the error is passed and the job id is still the `watched_job_id`. @commandlinegirl

src/python/dxpy/scripts/dx.py
```diff
@@ -2989,6 +2989,15 @@ def run_body(args, executable, dest_proj, dest_path, preset_inputs=None, input_n
         "extra_args": args.extra_args
     }
 
+    if any([args.watch or args.ssh or args.allow_ssh]):
+        if run_kwargs["priority"] in ["low", "normal"]:
```
commandlinegirl, line 2993: The condition should be if `run_kwargs["priority"] != high` (so that includes the case when it is not set at all).

src/python/dxpy/scripts/dx.py
commandlinegirl: Yes, imo it should not be printed with --brief, which will be consistent with out other commands. With this flag we usually only print ID or a short string in json or nothing.

commandlinegirl (changes requested):

src/python/dxpy/scripts/dx.py
emiloslavsky: agree

YuxinShi0423: Great I will merge the check with the one below. Thanks!

src/python/dxpy/scripts/dx.py
YuxinShi0423: If it is unset it would be `None` and set to high in the else below. But it would be easier to read by saying `run_kwargs["priority"] != high`. I will change it in this way and change the check below as well.

commandlinegirl: Also, please check the spec in the JIRA ticket (`(if priority is None, use "default" as the priority value in the warning above)`) and update it accordingly. Thanks!

src/python/test/test_dxclient.py
```diff
@@ -2781,36 +2782,90 @@ def _get_analysis_id(dx_run_output):
         # don't actually need it to run
         run("dx terminate " + watched_job_id)
 
-        # --ssh implies --priority high
+        # --ssh implies high priority when --priority is not specified
```
YuxinShi0423, line 2785: https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/60/console
@commandlinegirl If starting from --ssh test the `dx_run_output` is not defined. I guess even Jenkins could not launch the ssh test as I mentioned :( 
Previously `dx_run_out` here was not from this job but the `watched_job`.
======================================================================
ERROR: test_dx_run_priority (test_dxclient.TestDXClientRun)
----------------------------------------------------------------------
  File "/dx-toolkit/src/python/test/dxpy_testutil.py", line 694, in wrapper
    retval = func(*args, **kwargs)
  File "/dx-toolkit/src/python/test/test_dxclient.py", line 2792, in test_dx_run_priority
    ssh_job_id = dx_run_output.split('\n')[0]
UnboundLocalError: local variable 'dx_run_output' referenced before assignment

----------------------------------------------------------------------
Ran 1 test in 2.054s

FAILED (errors=1)
$ dx new project other --brief
project-G666x3j0Kk70zjVZ30vvZ0Xz

$ dx cd project-G666x3j0gYgF43gf24jp1k27:/

$ dx run myapplet -y --ssh --brief

commandlinegirl (approved):
LGTM, thank you!

src/python/test/test_dxclient.py
```diff
@@ -2781,36 +2782,87 @@ def _get_analysis_id(dx_run_output):
...
+        except subprocess.CalledProcessError:
+            # ignore any ssh errors; just want to test requested
+            # priority
+            pass
+        allow_ssh_job_id = re.search('job-[A-Za-z0-9]{24}', allow_ssh_run_output).group(0)
+        allow_ssh_job_desc = dxpy.describe(allow_ssh_job_id)
+        self.assertEqual(allow_ssh_job_desc['applet'], applet_id)
+        self.assertEqual(allow_ssh_job_desc['priority'], 'normal')
+        for string in ["WARNING", "normal", "interrupting interactive work"]:
+            self.assertNotIn(string, allow_ssh_run_output)
```
commandlinegirl, line 2844: Can this be moved under try? This should fix the failing test. Then run it in a branch on jenkins if possible.

YuxinShi0423: Sure that is a nice solution. Will do it and let you check :)

PR #731, opened 2021-11-16
APPS-894 fix tests for dx run priority with other flags

YuxinShi0423:


YuxinShi0423:
Currently tests with ssh could not pass:
https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/lastBuild/console
`Warning: Unable to verify configuration of your account for SSH connectivity:
[Errno 2] No such file or directory:
'/home/jenkins/.dnanexus_config/ssh_id.pub'. SSH connection will likely fail.`

However the changes in dx run with priority are tested here:

> $ dx run myapplet -y --watch --priority normal
WARNING: You have requested that jobs be run under normal priority, which may
cause them to be restarted at any point, interrupting interactive work.

src/python/test/test_dxclient.py
```diff
@@ -2770,98 +2770,104 @@ def _get_analysis_id(dx_run_output):
         # --watch implies high priority when --priority is not specified
         try:
```
kpjensen, line 2771: Can you wrap most of this test with `with self.configure_ssh() as wd:` to set up the keypair and update the public key of the test user?

kpjensen (changes requested):

src/python/test/test_dxclient.py
YuxinShi0423: https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/77/console
I have wrapped those tests. Works great! Thanks! @kpjensen

kpjensen (approved):

commandlinegirl (approved):

PR #741, opened 2022-01-05
DEVEX-2013 Restrict default SSH access, detect client IP

kpjensen:
If no IP is specified with `--allow-ssh`, instead of adding `*`, detect the client IP via the `/system/whoami` API route. 

Automatically update the allowed IP ranges list for SSH access when attempting to access a job via `dx ssh`. 

Add `--allow-ssh` parameter to `dx ssh` to allow specifying an IP address if it is different than what is detected by the `/syste/whoami`. 

Add `--no-firewall-update` parameter to `dx ssh` to skip all firewall updates if the job's allowSSH has already been configured at runtime.


emiloslavsky:
@commandlinegirl - could you assign someone on the APPS team to review this change please

src/python/dxpy/scripts/dx.py
```diff
@@ -3674,8 +3708,8 @@ def ssh(args, ssh_config_verified=False):
     if connected:
         sys.stdout.write(GREEN("OK") + "\n")
     else:
-        msg = "Failed to connect to {h}. Please check your connectivity and try {cmd} again."
-        err_exit(msg.format(h=host, cmd=BOLD("dx ssh {}".format(args.job_id))),
+        msg = "Failed to connect to {h}. Please check the allowed IP ranges for ssh in the job's describe output and if needed update with 'dx update job {job_id} --allow-ssh ADDRESS'. Then try {cmd} again."
```
emiloslavsky, line 3711: `dx ssh ... --allow-ssh ADDRESS'

commandlinegirl:
> @commandlinegirl - could you assign someone on the APPS team to review this change please

@mhrvol Would you review this?

src/python/dxpy/scripts/dx.py
```diff
@@ -3249,15 +3249,26 @@ def print_run_input_help():
...
-
+    client_ip = None
     if args.allow_ssh is not None:
-        args.allow_ssh = [i for i in args.allow_ssh if i is not None]
-    if args.allow_ssh == [] or ((args.ssh or args.debug_on) and not args.allow_ssh):
-        args.allow_ssh = ['*']
+        for i, ip in enumerate(args.allow_ssh):
+            if ip is None:
+                if client_ip is not None:
+                    del args.allow_ssh[i]
```
mhrvol, line 3257: Is it safe to delete from list while iterating through it?

could we iterate over something like this instead?
```
[ip for ip in args.allow_ssh if not (ip is None and client_ip is not None)]
```

Also, why do we delete `args.allow_ssh[i]` if client_ip is known, but if it isnt, we go get it and then set `args.allow_ssh[i]` to the value?

mhrvol:
> > @commandlinegirl - could you assign someone on the APPS team to review this change please
> 
> @mhrvol Would you review this?

I just reviewed this and then realized that it's been already merged. Did we you mean to comment this PR @emiloslavsky 

Also, there are some ruby and perl code changes, which look OK, but I've never really written in these languages.

PR #742, opened 2022-01-06
APPS-975 support full id when building from applet/workflow

YuxinShi0423:


src/python/dxpy/utils/resolver.py
```diff
@@ -679,13 +679,11 @@ def _check_resolution_needed(path, project, folderpath, entity_name, expected_cl
...
         return False, project, folderpath, None
     elif is_hashid(entity_name):
-
+        entity_class = str.split(entity_name,"-")[0]
         found_valid_class = True
-        if expected_classes is not None:
+        if expected_classes is not None and entity_class not in expected_classes:
             found_valid_class = False
-            for klass in expected_classes:
-                if entity_name.startswith(klass):
```
YuxinShi0423, line 687: a minor bug: if we found an entity `applet-xxxx` but we are expecting an `app`, since `applet` startswith `app` here `found_valid_class` becomes True.

src/python/dxpy/scripts/dx.py
```diff
@@ -2522,26 +2522,49 @@ def wait(args):
 def build(args):
     sys.argv = ['dx build'] + sys.argv[2:]
 
+    def get_from_exec_desc(args):
+        """
+        Return source executable description when --from option is used
+        
+        """
+        exec_describe_fields={'fields':{"properties":True, "details":True},'defaultFields':True}
+        _, _, exec_result = try_call(resolve_existing_path,
```
YuxinShi0423, line 2531: The resolved entity should be a single applet/workflow and the `exec_result` is a dict with field `id` and `description`.

src/python/dxpy/scripts/dx.py
```diff
@@ -2522,26 +2522,49 @@ def wait(args):
...
+
+        if exec_result is None:
+            err_exit('Could not resolve {} to an existing applet or workflow.'.format(args._from), 3)
+        elif len(exec_result)>1:
+            err_exit('More than one match found for {}. Please use an applet/workflow ID instead.'.format(args._from), 3)
+        else:
+            if exec_result[0]["id"].startswith("applet"):
+                return exec_result[0]["describe"]
+            elif exec_result[0]["id"].startswith("workflow"):
+                return exec_result[0]["describe"]
```
YuxinShi0423, line 2548: Return the description field only which contains the id already.

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -681,10 +681,10 @@ def _build_app_remote(mode, src_dir, publish=False, destination_override=None,
         shutil.rmtree(temp_dir)
 
 
-def build_app_from(applet_id, version, publish=False, do_try_update=True, bill_to_override=None,
+def build_app_from(applet_desc, version, publish=False, do_try_update=True, bill_to_override=None,
```
YuxinShi0423, line 684: We are injecting `args._from` still but now it is the description of the applet.

src/python/dxpy/scripts/dx.py
commandlinegirl: Could you add to the docstring what is expected in the `--from` argument (hash ID, qualified path, unqualified path, etc.)

src/python/dxpy/scripts/dx.py
commandlinegirl, line 2548: Since `expected_classes` argument allows only an applet or a workflow, we don't need the ifs here.

src/python/test/test_dxclient.py
```diff
@@ -7083,16 +7083,20 @@ def test_build_workflow_with_invalid_bill_to(self):
...
         # build global workflow without specified version
-        source_wf = "workflow-0000000000000000000000NA"
+        source_wf = self.create_workflow(project_id=self.project).get_id()
         with self.assertSubprocessFailure(stderr_regexp="--version must be specified when using the --from option", exit_code=2):
             run("dx build --globalworkflow --from {}".format(source_wf))
-    
+        with self.assertSubprocessFailure(stderr_regexp="--version must be specified when using the --from option", exit_code=2):
+            run("dx build --globalworkflow --from :{}".format(source_wf))
+        with self.assertSubprocessFailure(stderr_regexp="--version must be specified when using the --from option", exit_code=2):
+            run("dx build --globalworkflow --from {}:{}".format(self.project, source_wf))
```
commandlinegirl, line 7098: Please also add a test with a workflow name instead of ID.

commandlinegirl (changes requested):
Thanks @YuxinShi0423 for this PR! I found only a few minor issues.

src/python/dxpy/scripts/dx.py
YuxinShi0423: Actually we still need this. The resolving function would not check the class of entities if the input is applet/workflow name instead of ID.  Then a corner case is we found something by name but it is not applet/workflow. But I can merge these two cases since they are returning the same thing.

commandlinegirl: Got it. Thanks for the clarification.

src/python/dxpy/scripts/dx.py
YuxinShi0423: Added.

src/python/test/test_dxclient.py
```diff
@@ -7083,16 +7083,31 @@ def test_build_workflow_with_invalid_bill_to(self):
...
     def test_build_globalworkflow_without_version_override(self):
         # build global workflow without specified version
-        source_wf = "workflow-0000000000000000000000NA"
+        source_wf_id = self.create_workflow(project_id=self.project).get_id()
         with self.assertSubprocessFailure(stderr_regexp="--version must be specified when using the --from option", exit_code=2):
-            run("dx build --globalworkflow --from {}".format(source_wf))
-    
+            run("dx build --globalworkflow --from {}".format(source_wf_id))
+
+    def test_build_globalworkflow_with_workflow_path(self):
```
YuxinShi0423, line 7096: Added another test for different path formats.

commandlinegirl (approved):
LGTM, thank you!

PR #754, opened 2022-01-20
DEVEX-2034 Fix help for timestamp find args

kpjensen:


emiloslavsky (changes requested):
please add similar verbiage fixes in other places

emiloslavsky (approved):

PR #766, opened 2022-02-24
DEVEX-2055 Only query for folder when checking folder exists in `dx cd`

kpjensen:
Optimize the `project-xxxx/listFolder` call to only return folders, not objects in case the folder contains thousands of objects. Safe to do as this call is only checking that the folder argument to `dx cd` exists.


commandlinegirl (approved):
Nice! LGTM!

PR #773, opened 2022-04-20
APPS-1157 symlink download improvements

mhrvol:
Based on https://github.com/dnanexus/dx-toolkit/pull/709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Unreleased
 
+## Fixed
+
+* Improvements to symlink downloading reliability by solely using `aria2c` and enhancing options around its use
+
 ## [323.0] - beta
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Fixed
 
-* Improvements to symlink downloading reliability by solely using `aria2c` and enhancing options around its use
+* Improvements to symlink downloading reliability by solely using `aria2c` and enhancing options around its use (removes `wget` option for downloading symlinked files, adds the ability to set max tries for aria2c, adds `-c` flag for continuing downloads, removes the `--check-certificate=false` option).
 
 ## [323.0] - beta
 
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -97,7 +97,7 @@ def new_dxfile(mode=None, write_buffer_size=dxfile.DEFAULT_BUFFER_SIZE, expected
 
 
 def download_dxfile(dxid, filename, chunksize=dxfile.DEFAULT_BUFFER_SIZE, append=False, show_progress=False,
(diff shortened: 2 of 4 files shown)


src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -175,65 +178,53 @@ def _verify(filename, md5digest):
 
 # [dxid] is a symbolic link. Create a preauthenticated URL,
 # and download it
-def _download_symbolic_link(dxid, md5digest, project, dest_filename):
+def _download_symbolic_link(dxid, md5digest, project, dest_filename, symlink_max_tries=15):
+    print(symlink_max_tries)
+    if symlink_max_tries < 1:
```
mhrvol, line 183: symlink_max_tries is currently checked in the download parser, so there is no way this code will ever be triggered. I added this code just to make sure that if somebody ever uses this, we will never get infinite loop (which would be case for symlink_max_tries == 0 if the file is for some reason 'undownloadable')

src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -97,7 +97,7 @@ def new_dxfile(mode=None, write_buffer_size=dxfile.DEFAULT_BUFFER_SIZE, expected
 
 
 def download_dxfile(dxid, filename, chunksize=dxfile.DEFAULT_BUFFER_SIZE, append=False, show_progress=False,
-                    project=None, describe_output=None, **kwargs):
+                    project=None, describe_output=None, symlink_max_tries=False, **kwargs):
```
commandlinegirl, line 100: The default should not be boolean.

src/python/dxpy/scripts/dx.py
```diff
@@ -4566,6 +4571,9 @@ def register_parser(parser, subparsers_action=None, categories=('other', ), add_
                              action='store_false', default=sys.stderr.isatty())
 parser_download.add_argument('--lightweight', help='Skip some validation steps to make fewer API calls',
                              action='store_true')
+parser_download.add_argument('--symlink-max-tries', help='Set maximum tries for downloading symlinked files using aria2c',
```
commandlinegirl, line 4574:
```suggestion
parser_download.add_argument('--symlink-max-tries', help='Set maximum number of tries for downloading symlinked files using aria2c',
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Unreleased
 
+## Fixed
+
+* Improvements to symlink downloading reliability by solely using `aria2c` and enhancing options around its use
```
commandlinegirl, line 11:
```suggestion
* Improvements to symlink downloading reliability by solely using `aria2c` and enhancing options around its use (removes `wget` option for downloading symlinked files, adds the ability to set max tries for aria2c, adds `-c` flag for continuing downloads, removes the `--check-certificate=false` option).
```

commandlinegirl (changes requested):
Minor requests. Thanks!

src/python/dxpy/bindings/dxfile_functions.py
mhrvol: Thanks! fixed

commandlinegirl (approved):
LGTM, thanks!

PR #776, opened 2022-04-25
DEVEX-2079 Fix comparison of workflow directory to workflow name in spec

kpjensen:


commandlinegirl (approved):
LGTM, thanks for the fix to this!

PR #779, opened 2022-05-12
DEVEX-2094 Supply default 15 symlink_max_tries if not present in args

kpjensen:
Not recommended, but `dx get file-xxxx` does work to download files. 
```
$ dx get file-xxxx
  File "/Users/kjensen/Library/Python/3.8/lib/python/site-packages/dxpy/cli/download.py", line 98, in download_one_file
    symlink_max_tries=args.symlink_max_tries)
AttributeError: 'Namespace' object has no attribute 'symlink_max_tries'
```

### DIFF
--- a/src/python/dxpy/cli/download.py
+++ b/src/python/dxpy/cli/download.py
@@ -89,13 +89,14 @@ def download_one_file(project, file_desc, dest_filename, args):
 
 
     try:
+        symlink_max_tries = args.symlink_max_tries if vars(args).get('symlink_max_tries') is not None else 15
         dxpy.download_dxfile(
                             file_desc['id'],
                             dest_filename,
                             show_progress=show_progress,
                             project=project,
                             describe_output=file_desc,
-                            symlink_max_tries=args.symlink_max_tries)
+                            symlink_max_tries=symlink_max_tries)
         return
     except:
         err_exit()


src/python/dxpy/cli/download.py
```diff
@@ -89,13 +89,14 @@ def download_one_file(project, file_desc, dest_filename, args):
...
     try:
+        symlink_max_tries = args.symlink_max_tries if vars(args).get('symlink_max_tries') is not None else 15
         dxpy.download_dxfile(
                             file_desc['id'],
                             dest_filename,
                             show_progress=show_progress,
                             project=project,
                             describe_output=file_desc,
-                            symlink_max_tries=args.symlink_max_tries)
+                            symlink_max_tries=symlink_max_tries)
```
commandlinegirl, line 99: Shouldn't it be supplied by the `default` value when args is not provided by the user? https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/scripts/dx.py#L4576

kpjensen: Yes, but this specific `download_one_file()` function is called by `dx get file-xxxx` https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/scripts/dx.py#L1897 which does not have the `--symlink-max-tries` parameter.

I  added the check here in case `download_one_file()` is called by anything else, although that's doubtful.

commandlinegirl: Thanks!

PR #781, opened 2022-05-18
DEVEX-2041 Add '--rank' arg for `dx run`

kpjensen:


src/python/dxpy/scripts/dx.py
```diff
@@ -5181,6 +5182,7 @@ def positive_integer(value):
 parser_run.add_argument('--cost-limit', help=fill("Maximum cost of the job before termination. In case of workflows it is cost of the "
                                                   "entire analysis job. For batch run, this limit is applied per job.",
                                               width_adjustment=-24), metavar='cost_limit', type=float)
+parser_run.add_argument('-r', '--rank', type=int, metavar="[-1024-1023]", help='Set the rank of the root execution. Default is 0.', default=None)
```
emiloslavsky, line 5185:
```suggestion
parser_run.add_argument('-r', '--rank', type=int, metavar="[-1024,1023]", help='Set the rank of the root execution. Default is 0. Larger values lead to earlier execution.', default=None)
```

@XingGemini
1.  please review as well.
2.  I don't see [ workflow-xxxx/run ](https://documentation.dnanexus.com/developer/api/running-analyses/workflows-and-analyses#api-method-workflow-xxxx-run) accepting rank; if that's the case we need to mention that only apps and applets accept this flag.
3. is executionRankEnabled feature required to specify rank at runtime?  If so, we need to mention this fact

kpjensen: 2. The doc is incorrect, it's supported. 
3. Update the helpstring with the rquired license.

emiloslavsky (approved):

src/python/dxpy/scripts/dx.py
XingGemini: workflow-xxxx/run is supported

PR #785, opened 2022-06-02
TIP-2946 Update DXHTTPRequest to pass ssl_context

durae:
Update DXHTTPRequest to pass ssl_context to urllib3.PoolManager. This change enables to use our own ssl_context for request to private API in internal apiserver.

### DIFF
--- a/src/python/dxpy/__init__.py
+++ b/src/python/dxpy/__init__.py
@@ -247,7 +247,7 @@ def _get_env_var_proxy(print_proxy=False):
           file=sys.stderr)
   return proxy
 
-def _get_pool_manager(verify, cert_file, key_file):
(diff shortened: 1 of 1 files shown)


durae:
@kpjensen , @emiloslavsky what do you think about this change? Is it something that we can add to DXHTTPRequest for internal use?

emiloslavsky:
@durae does this introduce any user-visible changes?

durae:
@emiloslavsky no user visible changes. Vizserver calls apiserver private route only internally. This new param just allows to pass one more setting to the request.

emiloslavsky:
I'll defer to Kurt's review in that case.

kpjensen (approved):

PR #786, opened 2022-06-08
SCIPROD-432 SCIPROD-499 SCIPROD-540 SCIPROD-560: CAWFR dataset utilities

Niranjanpandeshwar:
Adding new command `dx extract_dataset` to dx-toolkit.


src/python/dxpy/scripts/dx.py
```diff
@@ -6002,6 +6003,21 @@ def positive_integer(value):
...
+                                   description='(Preview feature) Retrieves the data, or generates SQL to retrieve the data, from a dataset or cohort for a set of entity.fields.',
+                                   prog='dx extract_dataset')
+parser_extract_dataset.add_argument('path', help='v3.0 Dataset or Cohort object ID (project-id:record-id where ":record-id" indicates the current selected project) or name')
+parser_extract_dataset.add_argument('-ddd', '--dump-dataset-dictionary', action="store_true", default=False, help='If provided, the three dictionary files: <record_name>.data_dictionary.csv, <record_name>.entity_dictionary.csv, <record_name>.codings.csv are generated using the -delimiter specified, if applicable, in the output directory, if not provided, in the current location')
+parser_extract_dataset.add_argument('--fields', nargs='+', help='A comma-separated STRING  where each value is the phenotypic entity name and field name separated by a dot. If multiple entities are provided, they are inner-joined into one return. For example: <entity_name>.<field_name>,<entity_name>.<field_name>')
```
arunbhat, line 6014: Is this help string from the spec? Not sure why STRING is capitalized. @amesma0 @andydnanexus could one of you please review all the help strings from line 6009-6017?

src/python/dxpy/scripts/dx.py
```diff
@@ -6002,6 +6003,21 @@ def positive_integer(value):
...
+parser_extract_dataset.add_argument('--sql', action="store_true", default=False, help='If provided, a SQL statement (STRING) will be returned instead of data')
+parser_extract_dataset.add_argument('--delim', '--delimiter', nargs='?', const=',', default=',', help='Always use exactly one of DELIMITER to separate fields to be printed; if no delimiter is provided with this flag, COMMA will be used')
```
arunbhat, line 6016: Not sure what "one of DELIMITER" means

src/python/dxpy/scripts/dx.py
amesma0: Confused on "Retrieves the data, or generates SQL".

Niranjanpandeshwar: @arunbhat - yes, help strings come from Spec.
@amesma0 - if only --fields in passed in input, data is retrieved and returned. If --fields and --sql both are passed, a SQL to retrieve the data is generated and returned.

src/python/dxpy/scripts/dx.py
Niranjanpandeshwar: I believe when the spec was started, we had a limited set of delimiters. This line has to change. Any single character delimiter is now allowed.

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -0,0 +1,559 @@
...
+from __future__ import print_function, unicode_literals, division, absolute_import
+
+import sys
+import collections
+import json
+import os
+import re
+import csv
+import dxpy
+import pandas as pd
```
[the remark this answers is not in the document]
jkotrs: Niranjan, this is the preferred way according to PEP8 (https://peps.python.org/pep-0008/#imports). If you'd opt for importing in smaller scope (like method) then you'd need to spell it in every such method as that would become the only scope the imported module is visible. However subsequent scoped-down imports would make use of the module loaded so they are de-facto just a reference and do not add any extra overhead (see https://stackoverflow.com/questions/4789711/scope-of-imported-modules-functions-in-python)

So I'd keep it this way as `pandas` is made part of `requirements.txt` so it will get installed anyways.

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -0,0 +1,559 @@
...
+                                        {"project": project, "cohortBrowser": False} )
+    except Exception as details:
+        if details.name == 'PermissionDenied':
+            print("Insufficient permissions")
+            sys.exit(1)
+        elif details.name in ["InvalidInput", "InvalidState"]:
+            print('%r : Invalid cohort or dataset' % entity_result['id'])
+            sys.exit(1)
+        else:
+            raise ResolutionError(str(details))
```
jkotrs, line 72: Minor: I would suggest to use multiple `except` statements for clarity and also to avoid `name` checking as string (like https://stackoverflow.com/questions/6095717/python-one-try-multiple-except)

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -0,0 +1,559 @@
...
+    
+    if file_already_exist:
+        err_exit(fill("Error: path already exists {path}".format(path=file_already_exist)))
+
+    rec_descriptor = DXDataset(dataset_id, project=dataset_project).get_descriptor()
+    if args.fields is not None:
+        fields_list = ''.join(args.fields).split(',')
+        error_list = []
+        for entry in fields_list:
+            entity_field = entry.split('.')
```
jkotrs, line 152: Minor: since you're only ever referring to the first and second index, you might save a bit of typing by unpacking, like:
```
a, b = entry.split(".")[:2]
```

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -0,0 +1,559 @@
...
+                print(sql_results)
+            else:
+                with open(out_file_field, 'w') as f:
+                    print(sql_results, file=f)
+        else:
+            resource_val = resp['url'] + '/data/3.0/' + resp['dataset'] + '/raw'
+            try:
+                resp_raw = dxpy.DXHTTPRequest(resource=resource_val, data=payload, prepend_srv=False)
+                if 'error' in resp_raw.keys():
+                    if resp_raw['error']['details'] and 'PermissionDenied: Access to the specified database is not allowed' in resp_raw['error']['details']:
```
jkotrs, line 185: Would be great if we had specific error codes to target here. We currently only operate in english but should we ever introduce some localization to our error messages (or even just change the english wording) this check would fail.

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -0,0 +1,559 @@
...
+    csv_writer = csv.DictWriter(fields_output, delimiter=sep, doublequote=True, escapechar = None, lineterminator = "\n", quotechar = str('"'), 
+                                quoting = csv.QUOTE_MINIMAL, skipinitialspace = False, strict = False, fieldnames=column_names)
+    csv_writer.writeheader()
+    for entry in raw_results:
+        csv_writer.writerow(entry)
+
+    if not print_to_stdout:
+        fields_output.close()
+    
+class DXDataset(DXRecord):
```
jkotrs, line 218: Would it make sense to move this class definition into its separate file? To follow the python/dxpy/bindings example. But not necessary!:)

src/python/test/test_extract_dataset.py
```diff
@@ -0,0 +1,125 @@
...
+from __future__ import print_function, unicode_literals, division, absolute_import
+
+import unittest
+import tempfile
+import shutil
+import os
+import subprocess
+import pandas as pd
+
+class TestDXExtractDataset(unittest.TestCase):
```
jkotrs, line 28: Great test set! Since there were a couple of exceptions raised in the code (like non-existent output directory) I would suggest to add a (could be single) negative test to verify we indeed fail on that (instead of silently passing).:)

jkotrs (approved):
Great work!

src/python/dxpy/scripts/dx.py
amesma0: Can you just add that part in?

src/python/dxpy/cli/dataset_utilities.py
Niranjanpandeshwar: We are not limiting to [:2] to handle situations like `entry=entity_name.field_name.random_chars`. The code should fail and show this as an invalid input.

src/python/dxpy/cli/dataset_utilities.py
Niranjanpandeshwar: Seeking opinion from other reviewers. Will let @kpjensen make the final call.

src/python/dxpy/cli/dataset_utilities.py
Niranjanpandeshwar: Agreed! We don't have a code for this error as of now. In this case, even if the error message gets changed somewhere, the else part will catch the error. But I totally agree that having error codes is much more cleaner.

src/python/dxpy/cli/dataset_utilities.py
Niranjanpandeshwar: Updated! Thank you Jan.

src/python/test/test_extract_dataset.py
Niranjanpandeshwar: Added 1 test case for `file already exists` error. Other exceptions are being tested through QE.

src/python/test/test_extract_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+    def test_e2e_dataset_ddd(self):
+        dataset_record = "project-G9j1pX00vGPzF2XQ7843k2Jq:record-GBGpKZ80G8j8k4G84P7GP0B5"
+        out_directory = tempfile.mkdtemp()
+        cmd = ["dx", "extract_dataset", dataset_record, "-ddd", "-o", out_directory]
+        subprocess.check_call(cmd)
+        self.end_to_end_ddd(out_directory=out_directory, rec_name = "extract_dataset_test")
+
+    def test_e2e_cohortbrowser_ddd(self):
+        cohort_record = "project-G9j1pX00vGPzF2XQ7843k2Jq:record-GBGq1pQ0vGPf0qYg30Jg7kkG"
```
commandlinegirl, line 37: Please try not to hardcode the record but create it programmatically when the tests are run (something very simple, with a couple of fields only required to only test the currently introduced commands).

The created cohort record could be a minimized version of the one you're currently using. This would also make the `truth_output` (below) simpler, more readable.

src/python/test/test_extract_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+    def test_file_already_exists(self):
+        cohort_record = "project-G9j1pX00vGPzF2XQ7843k2Jq:record-GBGq1pQ0vGPf0qYg30Jg7kkG"
+        out_directory = tempfile.mkdtemp()
+        open(os.path.join(out_directory, "Combined_Cohort_Test.csv"), 'w').close()
+        cmd = ["dx", "extract_dataset", cohort_record, "--fields", "patient.patient_id" , ",", "patient.name", "-o", out_directory]
+        output = subprocess.run(cmd, capture_output=True, text=True)
+        self.assertTrue("Error: path already exists" in output.stderr)
+        shutil.rmtree(out_directory)
+
+    def end_to_end_ddd(self, out_directory, rec_name):
```
commandlinegirl, line 92: what does `ddd` stand for? Please rename the method to be more readable, thanks!

src/python/test/test_extract_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+        truth_files_directory = tempfile.mkdtemp()
+        os.chdir(truth_files_directory)
+        cmd = ["dx", "download", "project-G9j1pX00vGPzF2XQ7843k2Jq:file-GBGp61j0vGPpz58XK9f7Vkj6",
```
commandlinegirl, line 95: Is the project public (on staging)?

We try to keep dx-toolkit test environment-agnostic (prod/stg-agnostic) so is it possible to create the content of the files on the fly as the tests are run (and use it for the assertion/truth)?

src/python/dxpy/scripts/dx.py
```diff
@@ -6002,6 +6003,21 @@ def positive_integer(value):
 parser_unarchive.set_defaults(func=archive, request_mode="unarchival")
 register_parser(parser_unarchive, categories='fs')
 
+#####################################
+# extract_dataset
+#####################################
+parser_extract_dataset = subparsers.add_parser('extract_dataset', help='(Preview feature) Retrieves the data, or generates SQL to retrieve the data, from a dataset or cohort for a set of entity.fields.',
+                                   description='(Preview feature) Retrieves the data, or generates SQL to retrieve the data, from a dataset or cohort for a set of entity.fields.',
+                                   prog='dx extract_dataset')
+parser_extract_dataset.add_argument('path', help='v3.0 Dataset or Cohort object ID (project-id:record-id where ":record-id" indicates the current selected project) or name')
```
commandlinegirl, line 6012: Shouldn't it say "where "record-id" indicates the currently selected record"?

src/python/dxpy/scripts/dx.py
```diff
@@ -6002,6 +6003,21 @@ def positive_integer(value):
...
+parser_extract_dataset.add_argument('-ddd', '--dump-dataset-dictionary', action="store_true", default=False, help='If provided, the three dictionary files: <record_name>.data_dictionary.csv, <record_name>.entity_dictionary.csv, <record_name>.codings.csv are generated using the -delimiter specified, if applicable, in the output directory, if not provided, in the current location')
```
commandlinegirl, line 6013: in the current location on the Platform

src/python/dxpy/scripts/dx.py
commandlinegirl: Hm, it's not clear to me either. Should it say "Always use exactly one delimiter"?

Also, I suggest using "comma" (without capitalization), here and in other places in the helpstrings (e.g. STRING -> string)

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -0,0 +1,561 @@
...
+        raise err_exit(fill('Invalid delimiter specified'))
+    
+    project, path, entity_result = resolve_existing_path(args.path)
+
+    try:
+        resp = dxpy.DXHTTPRequest('/' + entity_result['id'] + '/visualize',
+                                        {"project": project, "cohortBrowser": False} )
+    except PermissionDenied:
+        print("Insufficient permissions")
+        sys.exit(1)
```
commandlinegirl, line 66: How about calling `err_exit`, e.g. like [here](https://github.com/dnanexus/dx-toolkit/pull/786/files#diff-9fa6c7c89cfc29f5fe6314b6c36a1dfab692ac8a3378d7e1d90e244ed233d40bR77)

src/python/dxpy/cli/dataset_utilities.py
commandlinegirl: Agree with @jkotrs, we shouldn't rely on error outputs that are not a part of exception/error API. How about simply printing/throwing the error we got from the server?

commandlinegirl (changes requested):
Partial review (`dataset_utilities.py` left to review).

Thanks @Niranjanpandeshwar!

src/python/test/test_extract_dataset.py
Niranjanpandeshwar: Dump Dataset Dictionary.
`-ddd/--dump-dataset-dictionary` is one of the arguments that can be passed with `dx extract_dataset` command.

src/python/test/test_extract_dataset.py
Niranjanpandeshwar: I don't think the project is public. May be @kpjensen can tell us if this will be made public. `user-jenkins` has view access in the project for the purpose of jenkins build and test.

The input to `extract_dataset` is a dataset or cohort on staging/platform. If this has to be created on the fly, we will need to check-in `data_dictionary.csv`, `entity_dictionary.csv`, `codings.csv` and the`test_data_to_load.csv` to the repo. Then run `data model loader`(DML) to create the dataset and use it in extract_dataset. Running DML can take around 5 minutes to create a small dataset.

After running DML and extract dataset, for validating the results we need truth files. Those will also have to be checked in the repo.

src/python/dxpy/scripts/dx.py
Niranjanpandeshwar: Not on the platform. The files are created on local where the command is run from. Agreed, this should be made more clear. We can change it to -> In the current location where `dx extract_dataset` is being run.

src/python/dxpy/cli/dataset_utilities.py
Niranjanpandeshwar: We can print the error message from server, but that will stop us from printing the error message defined in the spec to handle this error. @durae what do you suggest here?

durae: That error message is actually not correct in the condition, so let's do something similar to: 
```
if resp_raw['error']['type'] == 'InvalidInput':
    print("Insufficient permissions due to the project policy.")
    print(resp_raw['error']['message'])
```

Niranjanpandeshwar: Updated!

src/python/dxpy/cli/dataset_utilities.py
Niranjanpandeshwar: `err_exit` raises an exception which will again be caught by the next except statement. We are trying to keep error messages simple and not print stacktrace wherever possible. Updated the except statement below this from ResolutionError to err_exit.

src/python/dxpy/scripts/dx.py
Niranjanpandeshwar: Updated to `v3.0 Dataset or Cohort object ID (project-id:record-id where ":record-id" indicates the record-id in current selected project) or name`

Thank you!

src/python/dxpy/scripts/dx.py
Niranjanpandeshwar: Updated!

src/python/dxpy/scripts/dx.py
Niranjanpandeshwar: Has been decided to leave it as-is to match the help string in other dx commands.

src/python/dxpy/scripts/dx.py
Niranjanpandeshwar: Updated!

src/python/dxpy/cli/dataset_utilities.py
kpjensen: In this PR I think it's fine to leave in this file as the classes are not used elsewhere within dxpy like the other DNAnexus object bindings.

src/python/test/test_extract_dataset.py
kpjensen: Not planning to make the staging project public. @commandlinegirl given that this functionality is isolated from the rest of dx-toolkit I'm okay with the current approach, although perhaps we should set up the same data in prod and then change these tests to refer to `project_name:filename` instead of object IDs.

src/python/dxpy/cli/dataset_utilities.py
kpjensen: `err_exit()` can be provided expected exceptions and then will only print the stacktrace if `dxpy._DEBUG > 1`
e.g. https://github.com/dnanexus/dx-toolkit/blob/d321b315d4c03c5aab9125b083544ed3e35cd051/src/python/dxpy/scripts/dx.py#L1491-L1492

kpjensen (changes requested):
LGTM overall.

src/python/dxpy/cli/dataset_utilities.py
Niranjanpandeshwar: Updated.

src/python/test/test_extract_dataset.py
Niranjanpandeshwar: Replaced hard coded record IDs with record names as suggested in [comment](https://github.com/dnanexus/dx-toolkit/pull/786#discussion_r912257101).

src/python/test/test_extract_dataset.py
Niranjanpandeshwar: Updated.

ksoni:
@Niranjanpandeshwar, Looks good to me, very nice!

kpjensen (approved):

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -184,17 +188,18 @@ def extract_dataset(args):
             try:
                 resp_raw = dxpy.DXHTTPRequest(resource=resource_val, data=payload, prepend_srv=False)
                 if 'error' in resp_raw.keys():
-                    if resp_raw['error']['details'] and 'PermissionDenied: Access to the specified database is not allowed' in resp_raw['error']['details']:
-                        print("Insufficient permissions due to the project policy")
+                    if resp_raw['error']['type'] == 'InvalidInput':
+                        print("Insufficient permissions due to the project policy.")
```
commandlinegirl, line 192: "project policy" is not too informative. Should it say "project access policy"? Or which policy is it about? The wording should be such that the user ideally can quickly understand by a quick search in our documentation.

src/python/dxpy/scripts/dx.py
```diff
@@ -6006,13 +6006,13 @@ def positive_integer(value):
 #####################################
 # extract_dataset
 #####################################
-parser_extract_dataset = subparsers.add_parser('extract_dataset', help='(Preview feature) Retrieves the data, or generates SQL to retrieve the data, from a dataset or cohort for a set of entity.fields.',
-                                   description='(Preview feature) Retrieves the data, or generates SQL to retrieve the data, from a dataset or cohort for a set of entity.fields.',
+parser_extract_dataset = subparsers.add_parser('extract_dataset', help="(Preview feature) Retrieves the data or generates SQL to retrieve the data from a dataset or cohort for a set of entity.fields. Additionally, the dataset’s dictionary can be extracted independently or in conjunction with data.",
```
commandlinegirl, line 6009:
```suggestion
parser_extract_dataset = subparsers.add_parser('extract_dataset', help="(Preview feature) Retrieves the data or generates SQL to retrieve the data from a dataset or cohort for a set of entity.fields. Additionally, the dataset's dictionary can be extracted independently or in conjunction with data.",
```

src/python/dxpy/scripts/dx.py
```diff
@@ -6006,13 +6006,13 @@ def positive_integer(value):
...
+                                   description="(Preview feature) Retrieves the data or generates SQL to retrieve the data from a dataset or cohort for a set of entity.fields. Additionally, the dataset’s dictionary can be extracted independently or in conjunction with data.",
```
commandlinegirl, line 6010:
```suggestion
                                   description="(Preview feature) Retrieves the data or generates SQL to retrieve the data from a dataset or cohort for a set of entity.fields. Additionally, the dataset's dictionary can be extracted independently or in conjunction with data.",
```

src/python/dxpy/scripts/dx.py
```diff
@@ -6006,13 +6006,13 @@ def positive_integer(value):
...
-parser_extract_dataset.add_argument('-ddd', '--dump-dataset-dictionary', action="store_true", default=False, help='If provided, the three dictionary files: <record_name>.data_dictionary.csv, <record_name>.entity_dictionary.csv, <record_name>.codings.csv are generated using the -delimiter specified, if applicable, in the output directory, if not provided, in the current location')
-parser_extract_dataset.add_argument('--fields', nargs='+', help='A comma-separated STRING  where each value is the phenotypic entity name and field name separated by a dot. If multiple entities are provided, they are inner-joined into one return. For example: <entity_name>.<field_name>,<entity_name>.<field_name>')
-parser_extract_dataset.add_argument('--sql', action="store_true", default=False, help='If provided, a SQL statement (STRING) will be returned instead of data')
+parser_extract_dataset.add_argument('path', help='v3.0 Dataset or Cohort object ID (project-id:record-id where ":record-id" indicates the record-id in current selected project) or name')
```
commandlinegirl, line 6012:
```suggestion
parser_extract_dataset.add_argument('path', help='v3.0 Dataset or Cohort object ID (project-id:record-id where "record-id" indicates the record ID in the currently selected project) or name')
```

src/python/dxpy/scripts/dx.py
```diff
@@ -6006,13 +6006,13 @@ def positive_integer(value):
...
+parser_extract_dataset.add_argument('-ddd', '--dump-dataset-dictionary', action="store_true", default=False, help='If provided, the three dictionary files, <record_name>.data_dictionary.csv, <record_name>.entity_dictionary.csv, and <record_name>.codings.csv will be generated. Files will be comma delimited and written to the local working directory, unless otherwise specified using --delimiter and --output arguments. If any of the three dictionary files does not contain data (i.e, the dictionary is empty), then that particular file will not be created.')
```
commandlinegirl, line 6013:
```suggestion
parser_extract_dataset.add_argument('-ddd', '--dump-dataset-dictionary', action="store_true", default=False, help='If provided, the three dictionary files, <record_name>.data_dictionary.csv, <record_name>.entity_dictionary.csv, and <record_name>.codings.csv will be generated. Files will be comma delimited and written to the local working directory, unless otherwise specified using --delimiter and --output arguments. If any of the three dictionary files does not contain data (i.e. the dictionary is empty), then that particular file will not be created.')
```

src/python/dxpy/scripts/dx.py
```diff
@@ -6006,13 +6006,13 @@ def positive_integer(value):
...
+parser_extract_dataset.add_argument('--fields', nargs='+', help='A comma-separated string where each value is the phenotypic entity name and field name, separated by a dot.  For example: “<entity_name>.<field_name>,<entity_name>.<field_name>”. If multiple entities are provided, field values will be automatically inner joined. If only the --fields argument is provided, data will be retrieved and returned. If both --fields and --sql arguments are provided, a SQL statement to retrieve the specified field data will be automatically generated and returned.')
```
commandlinegirl, line 6014:
```suggestion
parser_extract_dataset.add_argument('--fields', nargs='+', help='A comma-separated string where each value is the phenotypic entity name and field name, separated by a dot.  For example: "<entity_name>.<field_name>,<entity_name>.<field_name>". If multiple entities are provided, field values will be automatically inner joined. If only the --fields argument is provided, data will be retrieved and returned. If both --fields and --sql arguments are provided, a SQL statement to retrieve the specified field data will be automatically generated and returned.')
```

commandlinegirl (changes requested):
Only minor requests.

src/python/dxpy/cli/dataset_utilities.py
Niranjanpandeshwar: @keegankelsey, @durae

durae: The error messages came originally from PM with security approval on not specifying project policy. We are currently checking 3 different project policies and not specifying which one caused the error in any other cases. 

I'm just trying to explain the origin of those error messages, not that they are correct or we should do it like that. Maybe @keegankelsey can chime in?

commandlinegirl: If it's general on purpose then it can stay as is. Thanks for providing more context @durae!

keegankelsey: I don't have information as to the rationale for keeping this general. However, if we think of a malicious actor (probably rare use case) then it would be good to minimize any targeted information for follow up on the part of the actor. So, I think keeping it general is fine.

commandlinegirl (approved):
LGTM, thanks @Niranjanpandeshwar!

durae (approved):
LGTM

Thanks very much @Niranjanpandeshwar for all the great work on this!

keegankelsey (approved):
Thanks, @Niranjanpandeshwar! Great work :)

PR #789, opened 2022-07-05
DEVEX-2109 Do not use job's workspace cotnainer-id in /run for detached jobs

kpjensen:
Check that when passing `project` for `/run` for a detached job that it is an actual project-id (to allow for running in another project as of https://github.com/dnanexus/dx-toolkit/commit/4320067aeaa07efeeeeaeece3503776b2c44df1c) and not the default of a job's temporary workspace container-id.


emiloslavsky (commented):
LGTM

PR #792, opened 2022-07-15
APPS-1289 reduce API calls for dx run applet-xxxx

JanProvaznik:


src/python/dxpy/toolkit_version.py
```diff
@@ -1 +1 @@
-version = '0.327.0'
+version = '0.327.1'
```
commandlinegirl, line 1: This gets updated during the release so we don't need to do it here.

src/python/test/test_dxclient.py
```diff
@@ -10656,6 +10656,21 @@ def test_dx_run_specific_project(self):
                             "code": "dx-jobutil-add-output number 32"}
             })
             run("dx run myapplet -inumber=5 --project %s" % temp_project.name)
```
commandlinegirl, line 10658: Minor: please add a newline.

commandlinegirl (changes requested):
Minor requests only.

commandlinegirl (approved):
LGTM, nice work, thank you!

PR #793, opened 2022-07-19
APPS-1291 multiple file upload --visibility hidden bug fix

JanProvaznik:


src/python/dxpy/cli/parsers.py
```diff
@@ -144,7 +144,7 @@ def process_dataobject_args(args):
     process_properties_args(args)
 
     # Visibility
-    args.hidden = (args.hidden == 'hidden')
+    args.hidden = (args.hidden == 'hidden' or args.hidden is True)
```
commandlinegirl, line 147: Why are we treating this arg as of a boolean type here?

commandlinegirl (changes requested):
Could you add a simple test for this, with two files even would be sufficient. Thanks!

src/python/dxpy/cli/parsers.py
JanProvaznik: The same args object is used for all files in a single `dx upload` call. But the same args.hidden is also used in `dxpy.upload_local_file(...)` which takes a bool. The conversion is handled here. Before, if you ran this code on args.hidden='hidden' twice, it would return False which is incorrect.

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ This change log spiritually adheres to [these guidelines](http://keepachangelog.
 Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Security
 
 ## Unreleased
+
+## [327.2]
+
+## Fixed
+
+* `dx upload f1 f2 --visibility hidden` now correctly marks both files as hidden
```
commandlinegirl, line 13: This should be added as the next item, below `Reduce the number..` (since this has not been released yet), and the rest of new additions can be removed. 

Once dx-toolkit is released, this `Changelog` will be updated (the unreleased items will be moved to beta and those from beta, to staging).

src/python/dxpy/cli/parsers.py
commandlinegirl: OK, got it, thanks for the clarification!

commandlinegirl (changes requested):
Comments only related to Changelog updates

commandlinegirl (approved):

PR #796, opened 2022-07-27
APPS-1285 add all SSL error types as retryable exceptions

JanProvaznik:
* Different types of SSL errors are raised depending on Python version,
* add all of them to prevent `dx upload` failing with bad internet connection


commandlinegirl (approved):
LGTM, thank you @JanProvaznik!

kpjensen (approved):

PR #798, opened 2022-07-29
APPS-1316 add --on-behalf-of parameter for `dx new user`

JanProvaznik:
* add `--on-behalf-of <org>` optional parameter for `dx new user`; the API is passed the org in the "provisioningOrg" field, which causes the welcome email to contain the correct org id.
* add a test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## [327.1]
 
+### Added
+
+* `--on-behalf-of <org>` argument for `dx new user`
+
 ## Fixed
 
 * Reduce the number of API calls for `dx run applet-xxxx` and `dx run workflow-xxxx`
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## [327.1]
 
+### Added
```
mhrvol, line 11: This should go into the  ##Unreleased section

src/python/test/test_dxclient.py
```diff
@@ -6268,6 +6268,29 @@ def test_create_user_account_and_set_bill_to_negative(self):
             with self.assertRaisesRegex(subprocess.CalledProcessError,
                                          "DXCLIError"):
                 run(" ".join([cmd, invalid_opts]))
+    
+    def test_create_user_on_behalf_of(self):
```
mhrvol, line 6272: Can we separate this test into  test_create_user_on_behalf_of_negative and  test_create_user_on_behalf_of?

mhrvol (changes requested):

mhrvol (approved):

PR #800, opened 2022-08-01
TIP-2995 Update help string for downloadRestricted project with databases

durae:


src/python/dxpy/bindings/dxproject.py
```diff
@@ -296,7 +296,7 @@ def new(self, name, summary=None, description=None, protected=None,
         :type protected: boolean
         :param restricted: If provided, whether the project should be restricted
         :type restricted: boolean
-        :param download_restricted: If provided, whether external downloads should be restricted
+        :param download_restricted: If provided, whether external downloads should be restricted. For database objects, whether the data in the project can be accessed by users from outside DNAnexus
```
emiloslavsky, line 299:
```suggestion
        :param download_restricted: If provided, whether external file downloads and external reads of database object should be restricted.
```

durae: I like that. @ptn24 what do you think?

ptn24: I would distinguish database objects because it would be reads and writes. I would go with what you had before, or something like "For database objects, whether external access should be prohibited"

ptn24 (approved):

emiloslavsky (approved):

PR #801, opened 2022-08-08
APPS-1232 nextflow progress

mhrvol:


src/python/dxpy/nextflow/nextflow_builder.py
```diff
@@ -0,0 +1,122 @@
...
+    dxpy.DXJob(job_id).wait_on_done(interval=1)
+    applet_id, _ = dxpy.get_dxlink_ids(dxpy.api.job_describe(job_id)['output']['output_applet'])
+    if not args.brief:
+        print("Created Nextflow pipeline %s" % (applet_id))
+    else:
+        print(applet_id)
+    return applet_id
+
+
+def prepare_nextflow(resources_dir) -> str:
```
commandlinegirl, line 56: Why do some function declare a returned type while others (e.g. just the one below) don't? Also, we want to support Python2.

src/python/dxpy/nextflow/nextflow_builder.py
```diff
@@ -0,0 +1,122 @@
...
+            "integer": "int",
+            "number": "float",
+            "boolean": "boolean",
+            "object":"hash"  # TODO: check default values
+        # TODO: add directory + file + path
+        }
+        if nf_type in types:
+            return types[nf_type]
+        return "string"
+        # raise Exception(f"type {nf_type} is not supported by DNAnexus")
```
commandlinegirl, line 93: Please start replacing all these with comments with proper code or delete the comments.

commandlinegirl (changes requested):
Partial review #1. Please add integration tests.

src/python/dxpy/nextflow/nextflow_builder.py
mhrvol: After we support files and directories, we will throw an Exception. Before that I'd like to return string for myself, not to run into unnecessary errors.

commandlinegirl: Got it, ok, thanks!

src/python/dxpy/nextflow/nextflow_builder.py
```diff
@@ -32,15 +32,18 @@ def build_pipeline_from_repository(args):
             "Can't create an applet without specifying a destination project; please use the -d/--destination flag to explicitly specify a project")
     input_hash = {
         "repository_url": args.repository,
+        "repository_tag": args.tag,
```
commandlinegirl, line 35: Which method the docstring from line 23 applies to? Should it perhaps be more informative, eg (?):

Creates files needed to build a nextflow applet and returns a folder with these files (dxapp.json, src/code.sh, etc). The folder also includes resources/ which contains the nextflow pipeline. Note that folder is created as a tempfile.

Please add a short description to this module itself as well. Thanks!

commandlinegirl: Also, per our discussion, please add docstrings. One example you can follow is [here](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/bindings/dxfile.py#L651).

src/python/dxpy/scripts/dx.py
```diff
@@ -2652,8 +2652,8 @@ def handle_arg_conflicts(args):
         if args.mode in ("globalworkflow", "applet", "app") and args.keep_open:
             build_parser.error("Global workflows, applets and apps cannot be kept open")
 
-        if args.nextflow and args.repository and not args.remote:
-            build_parser.error("Cannot build Nextflow pipeline from repository locally. Please use --remote.")
+        if args.nextflow and args.repository and args.remote:
+            build_parser.error("Nextflow pipeline built from repository is always build using Nextflow Pipeline Importer. This is not compatible with --remote.")
```
commandlinegirl, line 2656: Would this suggestion be more unambiguous?

```suggestion
            build_parser.error("Nextflow pipeline built from a remote Git repository is always build using the Nextflow Pipeline Importer app. This is not compatible with --remote.")
```

src/python/test/test_nextflow.py
```diff
@@ -0,0 +1,77 @@
...
+    return json.loads(build_output)['id']
+
+
+class TestNextflow(DXTestCase):
+    def test_temp(self):
+        print("test-message")
+        assert False
+
+    def test_basic_hello(self):
+        applet = build_nextflow_applet("./nextflow/", "project-GFYvg4Q0469VKVVVP359Yfpp")
```
commandlinegirl, line 70: We try not to hardcode projects in dx-toolkit tests (either from prod or staging). Please, check how we’re using and reusing the temporary test project built at setup time in one of the [parent](https://github.com/dnanexus/dx-toolkit/blob/f277ab349c51ff41c7a3737d012e797f9cdb8cef/src/python/test/dxpy_testutil.py#L341
) classes.

commandlinegirl (changes requested):
A friendly reminder about addressing my previous comments, thanks!

src/python/dxpy/scripts/dx.py
mhrvol: Fixed

mhrvol: Fixed

src/python/test/test_nextflow.py
mhrvol: Yes, I was playing around with the test code , it will be changed.

PR #814, opened 2022-08-29
APPS-1384 dynamic applet info

mhrvol:


src/python/dxpy/nextflow/nextflow_builder.py
```diff
@@ -87,11 +85,11 @@ def prepare_nextflow(resources_dir, profile):
     """
     assert os.path.exists(resources_dir)
     inputs = []
-    os.makedirs(".dx.nextflow", exist_ok=True)
     dxapp_dir = os.path.join(resources_dir, '.dx.nextflow')
+    os.makedirs(dxapp_dir, exist_ok=True)
     if os.path.exists(f"{resources_dir}/nextflow_schema.json"):
         inputs = prepare_inputs(f"{resources_dir}/nextflow_schema.json")
-    dxapp_content = get_nextflow_dxapp(inputs)
+    dxapp_content = get_nextflow_dxapp(inputs, os.path.basename(resources_dir))
```
YuxinShi0423, line 92: This method only allows the user to include a custom inputSpec, but all the other fields (name, titile, version, description etc.) are specified in the hidden template `dxapp.json`, right?

Would that be a better option if the user can also provide their own `dxapp.json` (just like when building regular dx applets) if they are building a local nf workflow? So we just need to add the `execDepends` (nextflow + nextaur) to their `dxapp.json`.

The template would still be used when they are building from git repository.
@mhrvol @commandlinegirl

mhrvol: I see. This is a thing to consider, I would take that as a separate task if this should be done.

commandlinegirl: Thanks for the suggestions, @YuxinShi0423. Agree that starting with the user's own dxapp.json is beyond the scope.

PR #815, opened 2022-08-30
APPS-1357 use job dest as nextflow outdir

YuxinShi0423:
When running a nextflow applet with output files, there could be cases like:

1. no `publishDir` defined in the pipeline code: nothing should be "published" explicitly, and all the output files remain in the working directory, which is a folder inside the job container, and will be destroyed after job completion
2. `publishDir` is defined as an dx path (`dx://project-xxx:/some/folder`): the output files will be uploaded to that exact dx path (if the project id is not given, then the job container id will be used)
3. `publishDir 'some/path'`: the files will be uploaded to the actual path relative to '/home/dnanexus/' in job container in the job container, since the main nf command is running in `/home/dnanexus`
    - `/home/dnanexus/some/path`
    - `some/path` if it is an absolute one
5. `publishDir params.user_given_output_path + '/some/subfolder'`, where the `user_given_output_path` is given at runtime as an extra arg via `dx run applet-xxxx -inf_run_args_and_pipeline_params=' --user_given_output_path some/user/output/folder'`
    - the files will be published to `some/user/output/folder/some/subfolder`
    - the `some/user/output/folder` could be a path string or a dx path
    - the parameter name "some_user_given_output_path" could be anything as long as it is referred in the definition of `publishDir` in the pipeline code


src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -64,7 +64,15 @@ main() {
     filtered_inputs=""
     
     @@RUN_INPUTS@@
-    nextflow -trace nextflow.plugin $nf_advanced_opts -log ${LOG_NAME} run /home/dnanexus/nfp @@PROFILE_ARG@@ -name run-${NXF_UUID} $nf_run_args_and_pipeline_params ${filtered_inputs}
+
+    # parse dnanexus-job.json to get job output destination
+    # and use that dxpath as the nextflow params.outdir
+    # TODO: override publishDir in user's main.nf to `publishDir=params.outdir`
```
YuxinShi0423, line 70: To make this work it requires the publishDir in user's workflow to be `params.outdir`. If not the file would be stored in the local path inside the container but never get uploaded to the job destination given by `dx run`

We might need to change the plugin code to recognize this param. Alternatively, we can embed this param in a config_file and force use this config with `-C config_file`. But if we do that we need to merge it with user's config, otherwise the user's config will be ignored.

mhrvol: we can use --publishDir=params.outdir, we dont need config file to override a value.

Though, if user had their own publishDir in their config, this would override the config value

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -64,7 +64,15 @@ main() {
...
+    OUT_PROJECT=$(cat /home/dnanexus/dnanexus-job.json | jq -r .project)"
```
mhrvol, line 71:
```suggestion
    OUT_PROJECT=$(jq -r .project /home/dnanexus/dnanexus-job.json)"
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
YuxinShi0423: https://www.nextflow.io/docs/latest/cli.html?highlight=directive#pipeline-parameters
I am not sure that `--publishDir` would work cuz double-dashed will be referred as params.publishDir. Could you share an example that we could override the directives like publishDir from command line? @mhrvol

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
YuxinShi0423: Fixed.

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
commandlinegirl: @YuxinShi0423 If we merged your code, the example that I posted in the JIRA ticket APPS-1357 works in the same way as when running locally?

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -64,7 +64,15 @@ main() {
     filtered_inputs=""
     
     @@RUN_INPUTS@@
-    nextflow -trace nextflow.plugin $nf_advanced_opts -log ${LOG_NAME} run /home/dnanexus/nfp @@PROFILE_ARG@@ -name run-${NXF_UUID} $nf_run_args_and_pipeline_params ${filtered_inputs}
+
+    # parse dnanexus-job.json to get job output destination
+    # and use that dxpath as the nextflow params.outdir
+    # TODO: override publishDir in user's main.nf to `publishDir=params.outdir`
+    OUT_PROJECT=$(jq -r .project /home/dnanexus/dnanexus-job.json)"
+    OUT_FOLDER=$(jq -r .folder /home/dnanexus/dnanexus-job.json)"
```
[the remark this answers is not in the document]
YuxinShi0423: Fixed. Thanks Ola!

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
mhrvol: I misunderstand what publishDir referred to back then.

I wonder, is it possible to see whether publishDir was used from the plugin?
Would creating job outputs solve our problem? (are files deleted if set as output of a job?)

YuxinShi0423: - is it possible to see whether publishDir was used from the plugin?
yes but it need to be evaluated if defined with some runtime params (see the case 4 above). I don't have enough time to fully investigate how to fetch and handle them now.
- Would creating job outputs solve our problem?
when subjobs are created the output specs are not specified in the api call `job/new` (and cannot be, see https://documentation.dnanexus.com/developer/api/running-analyses/job-input-and-output#output and
https://documentation.dnanexus.com/developer/api/running-analyses/applets-and-entry-points#api-method-job-new
could you elaborate on "creating job outputs"?
@mhrvol

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -2,12 +2,21 @@
...
-  # upload log file
-  dx upload $LOG_NAME --path $DX_LOG --wait --brief --no-progress --parents || true
   # backup cache
   echo "=== Execution complete — uploading Nextflow cache metadata files"
   dx rm -r "$DX_PROJECT_CONTEXT_ID:/.nextflow/cache/$NXF_UUID/*" 2>&1 >/dev/null || true
   dx upload ".nextflow/cache/$NXF_UUID" --path "$DX_PROJECT_CONTEXT_ID:/.nextflow/cache/$NXF_UUID" --no-progress --brief --wait -p -r || true
+  # parse dnanexus-job.json to get job output destination
+  OUT_PROJECT=$(jq -r .project /home/dnanexus/dnanexus-job.json)
+  OUT_FOLDER=$(jq -r .folder /home/dnanexus/dnanexus-job.json)
+  OUTDIR="$OUT_PROJECT:${OUT_FOLDER#/}"
```
mhrvol, line 12: If I understand it correctly, we are currently uploading all the files to the folder where the applet is run. correct?

YuxinShi0423: we will upload the log file as output `nextflow log` and all the published file as array of file `output_files` to the destination of the current job.

src/python/dxpy/templating/templates/nextflow/dxapp.json
```diff
@@ -35,6 +35,18 @@
     }
   ],
   "outputSpec": [
+    {
+      "name": "output_files",
```
commandlinegirl, line 39: Is there a way to read the output block of the nextflow pipeline and name the outputs after the names of the outputs in the pipeline (this would have to be done during the import).

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -60,7 +60,33 @@ on_exit() {
   # backup cache
   echo "=== Execution complete — uploading Nextflow cache metadata files"
   dx rm -r "$DX_PROJECT_CONTEXT_ID:/.nextflow/cache/$NXF_UUID/*" 2>&1 >/dev/null || true
-  dx upload ".nextflow/cache/$NXF_UUID" --path "$DX_PROJECT_CONTEXT_ID:/.nextflow/cache/$NXF_UUID" --no-progress --brief --wait -p -r || true
+
+  if [[ -d .nextflow/cache/$NXF_UUID ]]; then
+    dx upload ".nextflow/cache/$NXF_UUID" --path "$DX_PROJECT_CONTEXT_ID:/.nextflow/cache/$NXF_UUID" --no-progress --brief --wait -p -r || true
+  else
+    echo "No nextflow cache has been generated."
```
commandlinegirl, line 67: As per our discussion earlier today and the user doc (section "Scratch Folder and Intermediate Files") we won't be uploading cache (only the output files mentioned in the output block and the outputs mentioned with the publishDir directive will be uploaded).

commandlinegirl (changes requested):

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
commandlinegirl: Please remove this part (line 64-67)

src/python/dxpy/templating/templates/nextflow/dxapp.json
commandlinegirl: It looks like it depends on APPS-1348, which is to be implemented. OK, please ignore this comment for now then.

commandlinegirl (changes requested):
This LGTM modulo the integration tests.

YuxinShi0423:
Sample test: https://staging.dnanexus.com/projects/GG1P1p80728ypkgp9vyYg2kX/monitor/job/GGXP6QQ0728v9G163JVVv9Zg

commandlinegirl (approved):

src/python/test/test_nextflow.py
```diff
@@ -202,6 +226,40 @@ def test_dx_build_nextflow_from_repository_with_extra_args(self):
...
+            "dx build --nextflow '{}' --json".format(applet_dir)))["id"]
+        desc = dxpy.describe(applet_id)
+
+        # Run with "dx run".
+        dxfile = dxpy.upload_string("foo", name="foo.txt", folder="/a/b/c",
+                                    project=self.project, parents=True, wait_on_close=True)
+        inFile_path = "dx://{}:/a/b/c/foo.txt".format(self.project)
+        inFolder_path = "dx://{}:/a/".format(self.project)
+        outdir = "nxf_outdir"
+        pipeline_args = "'--outdir {} --inFile {} --inFolder {}'".format(
```
YuxinShi0423, line 247:
```suggestion
        pipeline_args = "\'--outdir {} --inFile {} --inFolder {}\'".format(
```
@commandlinegirl

PR #822, opened 2022-09-14
APPS-1384, APPS-1253 Fix pipeline naming; add tests

commandlinegirl:


src/python/dxpy/nextflow/nextflow_builder.py
```diff
@@ -56,6 +56,15 @@ def build_pipeline_from_repository(repository, tag, profile="", github_creds=Non
...
     return applet_id
 
+def get_resources_dir_name(resources_dir):
+    """
+    :param resources_dir: Directory with all source files needed to build an applet. Can be an absolute or a relative path.
+    :type resources_dir: str or Path
+    :returns: The name of the folder
+    :rtype: str
+    """
+    return os.path.basename(os.path.abspath(resources_dir))
```
YuxinShi0423, line 66: Actually I changed it as well because if the resources_dir happens to end with a `/`, i.e. `dx build --nextflow test/nextflow/` it would work still but the name of the applet would be empty.

src/python/test/test_nextflow.py
```diff
@@ -117,7 +116,53 @@ def test_prepare_inputs_large_file(self):
...
+    def test_dx_build_nextflow_set_default_metadata(self):
+        pipeline_name = "hello"
+        applet_dir = self.write_nextflow_applet_directory(pipeline_name, existing_nf_file_path=self.base_nextflow_nf)
+        applet_id = json.loads(run("dx build --nextflow --json " + applet_dir))["id"]
+        app = dxpy.describe(applet_id)
+        self.assertEqual(app['name'], pipeline_name)
+        self.assertEqual(app['title'], pipeline_name)
+        self.assertEqual(app['summary'], pipeline_name)
+
+    def test_dx_build_nextflow_with_abs_and_relative_path(self):
```
YuxinShi0423, line 130: I would suggest adding a test that applet_dir ends with a slash. Plz see my comment above :)

src/python/dxpy/nextflow/nextflow_builder.py
commandlinegirl: Could you provide an example where the name ends with / and the name end up empty? I don't see such a problem (expect for when it's /)

```
>>> resources_dir = "foo/bar/"
>>> os.path.basename(os.path.abspath(resources_dir))
'bar'
>>> resources_dir = "/"
>>> os.path.basename(os.path.abspath(resources_dir))
''
```

YuxinShi0423:
```
>>> os.path.basename('test/nextflow/')
''
>>> os.path.basename(os.path.abspath('test/nextflow/'))
'nextflow'
```
if my `main.nf` is stored under test/nextflow

I think your code has already made it fixed. Just want to comment on it and suggest another test. No further change needed. Sorry for the confusion!

YuxinShi0423 (approved):

PR #823, opened 2022-09-15
APPS-1323 show internetUsageIPs if job describe has --verbose

JanProvaznik:
For the test to pass, the user has to have jobInternetUsageMonitoring license feature.


src/python/dxpy/bindings/dxjob.py
```diff
@@ -213,6 +215,8 @@ def describe(self, fields=None, io=None, **kwargs):
         describe_input = {}
         if fields is not None:
             describe_input['fields'] = fields
+        if defaultFields is not None:
```
mhrvol, line 218: Would it make sense to consider default value for defaultFields (false I assume), so we could avoid None?

JanProvaznik: I'm slightly against it, we don't want to send it to the API at all, so in the case of backed changing the default value it would be propagated correctly. Now it'd just make the request longer without any semantic difference.

mhrvol (approved):

PR #827, opened 2022-09-19
DEVEX-2138 Install pyreadline3 on Windows Python 3.10+

kpjensen:
Add `pyreadline3` as a Windows only import for Python >= 3.5. Move the readline import inter a helper function to import the proper package depending on the OS.


commandlinegirl (approved):
LGTM

PR #845, opened 2022-09-23
Fix nextflow relative path test

kpjensen:
Switch back working directory after running relative path test.  

Skip windows-only test on other platforms.


commandlinegirl (approved):
LGTM, thank you!

PR #848, opened 2022-09-26
APPS-1279 fix `dx describe user-xxxx` with platformadminsupport role

JanProvaznik:


src/python/dxpy/utils/describe.py
```diff
@@ -983,7 +983,8 @@ def format_currency(value, meta, currency_locale=None):
 
 def print_user_desc(desc):
     print_field("ID", desc["id"])
-    print_field("Name", desc["first"] + " " + ((desc["middle"] + " ") if desc["middle"] != '' else '') + desc["last"])
+    if "first" in desc and "last" in desc: 
+        print_field("Name", desc["first"] + " " + ((desc["middle"] + " ") if desc["middle"] != '' else '') + desc["last"])
```
mhrvol, line 987: What was the problem beforehand?  Have you tested this?

From changelog I can see that empty name was printed - will this solve the issue?

if we have desc={"first":"", "last":""}, we would still pass the if and print empty field, or am I incorrect?

(If the issue was that there was an error in finding "first" and "last"  (desc = {}), shouldnt we check for desc["middle"] as well?)

commandlinegirl (approved):
LGTM, thank you!

PR #853, opened 2022-09-29
APPS-1438, APPS-1439  Remove default args from the Nextflow script and make inputs optional

mhrvol:


commandlinegirl (approved):
LGTM, modulo this passes both py2 and py3 test suites on Jenkins and the pipelines runs that ran into this pass. Thank you!

PR #856, opened 2022-10-03
APPS-1445 dx build --nextflow --brief returns json

mhrvol:


src/python/dxpy/nextflow/nextflow_builder.py
```diff
@@ -66,7 +66,7 @@ def parse_extra_args(extra_args):
     if not brief:
         print("Created Nextflow pipeline %s" % (applet_id))
     else:
-        print(applet_id)
+        print(dict(id=applet_id))
```
commandlinegirl, line 69: Is there a difference between this and [this](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/scripts/dx_build_app.py#L1042) (which prints [this](https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/scripts/dx_build_app.py#L758))? I.e. using dict vs dumping json? 

I made a note about tests in the ticket, please check it. I think the app needs to be updated and its update synced with the platform release.

commandlinegirl (changes requested):

src/python/dxpy/nextflow/nextflow_builder.py
mhrvol: I think it should be pretty much the same, because it outputs a string. I will change it to json dumps, just to make it same as we already had it in codebase.

Yes, I will create PR for that

mhrvol: There is actually no need to change anything in the NPI, as it uses --json anyway (that prints way more information about the applet in json format.)

commandlinegirl (approved):

src/python/dxpy/nextflow/nextflow_builder.py
commandlinegirl: What about the test?

commandlinegirl (changes requested):
Please, update the test(s)

https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/test_nextflow.py#L249
https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/test_nextflow.py#L222
plus at least one more

Please, run nextflow tests before submitting a PR, this will speed us up.

src/python/test/test_nextflow.py
```diff
@@ -203,7 +203,7 @@ def test_dx_build_nextflow_from_repository_default_metadata(self):
         pipeline_name = "hello"
         hello_repo_url = "https://github.com/nextflow-io/hello"
         applet_id = run(
-            "dx build --nextflow --repository '{}' --brief".format(hello_repo_url)).strip()
+            "dx build --nextflow --repository '{}' --brief | jq .id -r".format(hello_repo_url)).strip()
```
commandlinegirl, line 206: Thanks for the fix. This LGTM.

A sidenote for the future: you can also use [json.loads](https://github.com/dnanexus/dx-toolkit/blob/ea014fdf22bf970b800286cfba16f693fcf28f06/src/python/test/test_nextflow.py#L175-L176) (in case `jq` is not available in the test environment).

commandlinegirl (approved):
LGTM, thanks!

src/python/test/test_nextflow.py
mhrvol: I've changed it and rerun the tests.

PR #858, opened 2022-10-10
APPS-1412 Fixes to setting name in dx build

mhrvol:
--extra-args overrides --destination (name only)
warning is printed. Also fixed wrong error message when -f or -a was not specified (as dx-toolkit preferred destination over extra-args, but our backend handled it the other way around.)

### DIFF
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -726,7 +726,7 @@ def build_app_from(applet_desc, version, publish=False, do_try_update=True, bill
         "details", "access", "ignoreReuse"
     )
     inherited_metadata = {}
-    for field in fields_to_inherit: 
+    for field in fields_to_inherit:
         if field in applet_desc:
             inherited_metadata[field] = applet_desc[field]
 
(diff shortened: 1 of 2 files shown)


src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -850,12 +856,10 @@ def build_and_upload_locally(src_dir, mode, overwrite=False, archive=False, publ
             projects_by_region = {region: dest_project}
 
             if not overwrite and not archive:
-                # If we cannot overwite or archive an existing applet and an
+                # If we cannot overwrite or archive an existing applet and an
                 # applet in the destination exists with the same name as this
                 # one, then we should err out *before* uploading resources.
                 try:
-                    if not override_applet_name and kwargs.get("name"):
```
mhrvol, line 857: This check is no longer needed as we are dealing with everything when assigning value to `override_applet_name `

commandlinegirl (approved):
LGTM but integration tests must pass before merging this. Thank you!

mhrvol:
@commandlinegirl please merge if changelog is ok (feel free to update it.)

PR #860, opened 2022-10-14
APPS-1458 Fix to issue with overriding every required parameter

mhrvol:


src/python/dxpy/nextflow/nextflow_templates.py
```diff
@@ -78,7 +78,11 @@ def get_nextflow_src(custom_inputs=None, profile=None, resources_dir=None):
             '''% (i['name'], i['name'], value)
         # required inputs need to be added as runtime pipeline params
         else:
-            required_runtime_params += " --{}={}".format(i["name"], value)
+            required_runtime_params = required_runtime_params + f'''
+            if [ -n "${i['name']}" ]; then
+                required_inputs="${{required_inputs}} --{i['name']}=${i['name']}"
+            fi
+            '''
```
YuxinShi0423, line 85: f string might not be compatible with python 2

YuxinShi0423 (commented):
Partially reviewed. will test it out and review again. Thanks!

src/python/dxpy/nextflow/nextflow_templates.py
commandlinegirl: The pipeline applets are always Ubuntu 20.04+python3 though so IMHO we can leave the f string.

commandlinegirl:
> Partially reviewed. will test it out and review again. Thanks!

Hi @YuxinShi0423 did you have a chance to review it?

commandlinegirl (changes requested):
As we discussed, please add integration tests for this. If you think it's better to keep them [here](https://github.com/dnanexus/nextaur-app/tree/master/integration_tests), that's also OK, as long as we can test this (semi)-automatically before merging. Thanks!

mhrvol:
> As we discussed, please add integration tests for this. If you think it's better to keep them [here](https://github.com/dnanexus/nextaur-app/tree/master/integration_tests), that's also OK, as long as we can test this (semi)-automatically before merging. Thanks!

Few tests are added here: https://github.com/dnanexus/nextaur-app/pull/11

Currently run manually, but this is the main difference that this PR makes (tested on hello world program, where one required input is printed and it is set in nextflow profile config.):

Without this PR:
Ciao world!
 world!
Hola world!
Hello world!

With this PR:
test_input_something world!
Hello world!
Hola world!
Ciao world!

Basically, without this, the config input was never taken into consideration.
@commandlinegirl  feel free to merge in case there is no other issue.

commandlinegirl:
> > As we discussed, please add integration tests for this. If you think it's better to keep them [here](https://github.com/dnanexus/nextaur-app/tree/master/integration_tests), that's also OK, as long as we can test this (semi)-automatically before merging. Thanks!
> 
> Few tests are added here: [dnanexus/nextaur-app#11](https://github.com/dnanexus/nextaur-app/pull/11)
> 
> Currently run manually, but this is the main difference that this PR makes (tested on hello world program, where one required input is printed and it is set in nextflow profile config.):
> 
> Without this PR: Ciao world! world! Hola world! Hello world!
> 
> With this PR: test_input_something world! Hello world! Hola world! Ciao world!
> 
> Basically, without this, the config input was never taken into consideration. @commandlinegirl feel free to merge in case there is no other issue.

This is great! Thank you @mhrvol!

PR #863, opened 2022-10-24
APPS-1460 Set restartableEntryPoints to all for nextflow pipeline applets

commandlinegirl:


src/python/dxpy/templating/templates/nextflow/dxapp.json
```diff
@@ -60,6 +60,7 @@
   ],
   "runSpec": {
     "headJobOnDemand": true,
+    "restartableEntryPoints": "all",
```
mhrvol, line 63: Why do we need this? 

Nextflow creates a new job for retry (or am I misunderstanding something with retry errorStrategy?), should we care about DNAnexus restarts? Nextflow will restart the job on its own if needed.

mhrvol (approved):
LGTM

PR #867, opened 2022-10-24
APPS-1469 Ignore reuse in the template

commandlinegirl:
An integration test on this branch running a simple applet passed.


mhrvol (approved):
LGTM

PR #866, opened 2022-10-24
APPS-1466 print-nextflow-run-cmd

YuxinShi0423:
Screen Shot 2022-10-24 at 1 59 37 PM
https://staging.dnanexus.com/projects/GJ839q80QpbFg9BZ1Jk5J6Yg/monitor/job/GJFPX480QpbKZ511G7B8gzgK

PR #869, opened 2022-11-01
APPS-1466 Print command in the log (#866)

commandlinegirl:


commandlinegirl:
Already reviewed in https://github.com/dnanexus/dx-toolkit/pull/866

PR #877, opened 2022-11-15
DEVEX-2153 DEVEX-2154 Add '--external-upload-restricted' flag for `dx update project` and `dx find projects`

kpjensen:
Add `--external-upload-restricted` parameter to `dx update project` and `dx find projects`. Accepted values are `true` or `false`. 

Adjust the maximum width of describe output from 22 --> 28 characters to fit the new field name `externalUploadRestricted`.

### DIFF
--- a/src/python/dxpy/cli/parsers.py
+++ b/src/python/dxpy/cli/parsers.py
@@ -351,8 +351,9 @@ def get_update_project_args(args):
         input_params['allowedExecutables'] = args.allowed_executables
     if args.unset_allowed_executables:
         input_params['allowedExecutables'] = None
+    if args.external_upload_restricted is not None:
+        input_params['externalUploadRestricted'] = True if args.external_upload_restricted == 'true' else False
     return input_params
-
 
 def process_phi_param(args):
     if args.containsPHI is not None:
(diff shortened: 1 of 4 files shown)


src/python/dxpy/cli/parsers.py
```diff
@@ -351,12 +351,20 @@ def get_update_project_args(args):
...
             args.containsPHI = True
         elif args.containsPHI == "false":
             args.containsPHI = False
+
+def process_external_upload_restricted_param(args):
+    if args.external_upload_restricted is not None:
+        if args.external_upload_restricted == "true":
+            args.external_upload_restricted = True
+        elif args.external_upload_restricted == "false":
+            args.external_upload_restricted = False
```
commandlinegirl, line 370: Can this also be this (we'll also set it to False if the original value was any random string .. and perhaps we don't want that?) EDIT: though I can see the argparse only allows these two values so that suggestion should be OK.

```suggestion
        args.external_upload_restricted = (args.external_upload_restricted == "true")
```

commandlinegirl (approved):
LGTM, with a minor suggestion

PR #883, opened 2022-11-23
APPS-1512 Update records in the Nextflow template

commandlinegirl:
Before merging production assets must also be built and propagated.


commandlinegirl (commented):
LGTM (cannot officialy approve since I originally created this PR)

PR #888, opened 2022-11-30
APPS-1510 Add tests for handling custom inputs in nextflow applet script

YuxinShi0423:


src/python/test/test_nextflow.py
```diff
@@ -103,11 +115,27 @@ def test_src_profile(self):
     @unittest.skipIf(USING_PYTHON2,
         'Skipping as the Nextflow template from which applets are built is for Py3 interpreter only')
     def test_src_inputs(self):
-        src = get_nextflow_src(custom_inputs=[input1, input2])
-        self.assertTrue("if [ -n \"${}\" ];".format(input2.get("name")) in src)
-        self.assertTrue("--{}=${}".format(input2.get("name"), input2.get("name")))
```
YuxinShi0423, line 108: Seems like we are not asserting anything here

YuxinShi0423: @mhrvol I saw we have some tests https://github.com/dnanexus/nextaur-app/pull/11. Is any of them you think shall be included?

YuxinShi0423:
https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/261/console

src/python/test/test_nextflow.py
```diff
@@ -103,11 +115,27 @@ def test_src_profile(self):
     @unittest.skipIf(USING_PYTHON2,
         'Skipping as the Nextflow template from which applets are built is for Py3 interpreter only')
     def test_src_inputs(self):
```
commandlinegirl, line 117: Would it be possible to add a docstring to this test that described what it tests. So what in the future we know what's going on here, for example """ Tests that the --input= ... was passed to the nextflow run command in the applet source code. This happens because .. """ Whatever will make it easier to read the code and understand our motivations in the future. Thanks a bunch.

commandlinegirl (commented):
This PR is only about improving the test?

YuxinShi0423:
> This PR is only about improving the test?

I think the issue itself is already resolved by pull/871, so I had asked @yihchii to try the pipeline again. So yes if as expected the bug is fixed, then yes I will close this PR with just the test improvements.

commandlinegirl:
> > This PR is only about improving the test?
> 
> I think the issue itself is already resolved by [pull/871](https://github.com/dnanexus/dx-toolkit/pull/871), so I had asked @yihchii to try the pipeline again. So yes if as expected the bug is fixed, then yes I will close this PR with just the test improvements.

OK :) 

You mean merge this PR, not close?

src/python/test/test_nextflow.py
mhrvol: I've used multiple of these tests when debugging some changes in nextaur, I dont we need we need most of them. There are some for errorStrategy testing, but Jana from QE is creating test suite with these tests (or similar ones)

src/python/test/test_nextflow.py
```diff
@@ -54,6 +54,18 @@
     "help": "second input",
     "label": "Test2"
 }
+input3 = {
+    "class": "file",
+    "name": "third_input",
+    "help": "(Nextflow pipeline optional)third input",
```
mhrvol, line 60: I see that help message of the newly added inputs contain ()Nextflow pipeline optional), but the help message of input1/2 does not. Shouldnt they be the same?

src/python/test/test_nextflow.py
YuxinShi0423: Done. Please help take a look, thanks! @commandlinegirl

src/python/test/test_nextflow.py
YuxinShi0423: I kept the previous ones in case it is needed in some other tests you are using. But let me know if those are not needed anymore. 

I think we can keep all, although currently the only case that we have custom inputs is when using schema, and the inputs will have help message either with (Nextflow pipeline optional) or Nextflow pipeline required). However if later we allow getting more customizable inputs, input1/2 will cover those cases. cc @commandlinegirl

src/python/test/test_nextflow.py
commandlinegirl: Yes, this is great. We should also have it documented as a part of official documentation

commandlinegirl (approved):

commandlinegirl (approved):
If you could only change the PR title to something like .. APPS-1510 Add test for ...

PR #891, opened 2022-12-02
APPS-1533 updated nextaur version

mhrvol:


mhrvol:
I've updated the records @commandlinegirl  feel free to merge - btw I was not entirely sure what to put to the changelog, feel free to update it. (as we do not have nextflow public, at least not in dxtoolkit help, also I am not sure if "nextaur plugin update" would be readable enough.)

commandlinegirl (approved):

PR #897, opened 2022-12-16
APPS-1517 APPS-1568 terminate errorStrategy - failed subjob should fail

mhrvol:


mhrvol:
Records to be added in all regions.

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -406,6 +406,14 @@ nf_task_exit() {
   # mark the job as successful in any case, real task
   # error code is managed by nextflow via .exitcode file
   if [ -z ${exit_code} ]; then export exit_code=0; fi
+
+ terminate_record=$(dx find data --name $DX_JOB_ID --path $DX_WORKSPACE_ID --brief | head -n 1)
```
commandlinegirl, line 410: Could you fix the indent?

commandlinegirl: Also, if you could add the comment why we need this?

commandlinegirl (changes requested):

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
mhrvol: Fixed and commented. Also added few echo messages for the user.

commandlinegirl (approved):
LGTM, thank you @mhrvol

PR #898, opened 2022-12-16
TIP-3170 Removed downloadRestricted check for extract_dataset

austinlocke:
TIP-3170 Removed downloadRestricted check for extract_dataset command since it is properly checking in /visualize call

### DIFF
--- a/src/python/dxpy/cli/dataset_utilities.py
+++ b/src/python/dxpy/cli/dataset_utilities.py
@@ -155,8 +155,6 @@ def extract_dataset(args):
     if args.fields:
         if args.sql:
             file_name_suffix = '.data.sql'
-        elif resp["downloadRestricted"]:
-            err_exit('Insufficient permissions due to the project policy')
         else:
             file_name_suffix = out_extension
         


commandlinegirl (approved):

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -155,8 +155,6 @@ def _check_pandas_version(python_range, current_pandas_version, pandas_version_r
     if args.fields:
         if args.sql:
             file_name_suffix = '.data.sql'
-        elif resp["downloadRestricted"]:
```
ksoni, line 158: Looks like we can just remove this elif part.
But when I tested with that change, it eventually started connecting to thrift and fails there:
oot@job-GKKFPQk2FVxJ81qJPXP641jy:/# dx extract_dataset project-GK35pk92FVx10Q9fKQg4kqgK:record-GK9Gf0k2FVxP3V2G8Zv0xv8b --fields participant.pid
=========== downloadRestricted is true
{'type': 'ThreadPoolException', 'message': 'Failed to fetch data', 'details': "'NoneType' object has no attribute 'cursor'", 'time': '12/16/2022, 22:20:15'}

But that looks like different issue, related to the signature verification. So this change looks fine

ksoni (approved):
Looks good to me.

ksoni:
Hi @austinlocke , let's merge this change as it is working fine. we have now another issue related to the signature verification, maybe can enter a separate ticket for that as that might be just the configuration change in staging.
Also, try the same in the production to see if the signature verification issue exists in prod or not.

durae:
> Hi @austinlocke , let's merge this change as it is working fine. we have now another issue related to the signature verification, maybe can enter a separate ticket for that as that might be just the configuration change in staging. Also, try the same in the production to see if the signature verification issue exists in prod or not.

The verification works on production and doesn't work on staging in every region I tried so far. I've created TIP-3173 and I'm trying to find out why.

PR #907, opened 2023-01-06
DEVEX-2180 regex global flags must be at the start position since python 3.11

ptdtan:


commandlinegirl (commented):
Thanks for the update, @ptdtan. Did you also test it with Python version lower than 3.11 to make sure it's backwards compatible?

ptdtan:
Yes I did. This feature was deprecated since python 3.6 and is terminated officially in python 3.11. More info on the release notes of Python 3.11 here: https://docs.python.org/3/whatsnew/3.11.html.

> In [re](https://docs.python.org/3/library/re.html#module-re) [Regular Expression Syntax](https://docs.python.org/3/library/re.html#re-syntax), global inline flags (e.g. (?i)) can now only be used at the start of regular expressions. Using them elsewhere has been deprecated since Python 3.6.

commandlinegirl (commented):
LGTM, thanks!

commandlinegirl:
> Yes I did. This feature was deprecated since python 3.6 and is terminated officially in python 3.11. More info on the release notes of Python 3.11 here: https://docs.python.org/3/whatsnew/3.11.html.
> 
> > In [re](https://docs.python.org/3/library/re.html#module-re) [Regular Expression Syntax](https://docs.python.org/3/library/re.html#re-syntax), global inline flags (e.g. (?i)) can now only be used at the start of regular expressions. Using them elsewhere has been deprecated since Python 3.6.

Got it, thanks!

PR #908, opened 2023-01-06
APPS-1560 Add dxNextflow verification

commandlinegirl:
Tests: [Py3](https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/284/), [Py2](https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-master-integration-tests-on-staging-16.04-python2/1930/).

### DIFF
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -1028,7 +1028,7 @@ def verify_nf_license(destination, extra_args):
             return dxpy.PROJECT_CONTEXT_ID
 
     dest_project_to_check = get_project_to_check()
-    features = dxpy.DXHTTPRequest("/" + dest_project_to_check + "/checkFeatureAccess", {"features": ["dxNextflow"]})['features']
+    features = dxpy.DXHTTPRequest("/" + dest_project_to_check + "/checkFeatureAccess", {"features": ["dxNextflow"]}).get('features', {})
     # Expecting output {'features': {'dxNextflow': True}}
     
     dx_nextflow_lic = features.get("dxNextflow", False)
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -1028,7 +1028,7 @@ def verify_nf_license(destination, extra_args):
             return dxpy.PROJECT_CONTEXT_ID
 
     dest_project_to_check = get_project_to_check()
-    features = dxpy.DXHTTPRequest("/" + dest_project_to_check + "/checkFeatureAccess", {"features": ["dxNextflow"]}).get('features', {})
+    features = dxpy.DXHTTPRequest("/" + dest_project_to_check + "/checkFeatureAccess", {"features": ["dxNextflow"]}).get("features", {})
     # Expecting output {'features': {'dxNextflow': True}}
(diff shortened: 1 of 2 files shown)


src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -1014,6 +1014,27 @@ def get_destination_region(destination):
...
+        if "project" in extra_args:
+            return extra_args["project"]
+        if destination:
+            dest_project_id, _, _ = parse_destination(destination)
+            return dest_project_id
+        else:
+            return dxpy.PROJECT_CONTEXT_ID
+
+    dest_project_to_check = get_project_to_check()
+    features = dxpy.DXHTTPRequest("/" + dest_project_to_check + "/checkFeatureAccess", {"features": ["dxNextflow"]})['features']
```
mhrvol, line 1031: Shouldnt we have .get('features', {}) to avoid error in case of it missing?

mhrvol (approved):
minor, otherwise LGTM.

Btw users that try to build the applet do not need to have the license themselves, just the billto to the project? 

I am not exactly sure if any of our customers even somehow 'share' some projects etc.  I am just wondering if this is the correct way of checking the license.

mhrvol:
Also, should we include this change in the changelog?

src/python/dxpy/scripts/dx_build_app.py
YuxinShi0423: @mhrvol I think this feature is already been included with a default value `false` of orgs features. if any name here is invalid we will have `dxpy.exceptions.InvalidInput: Expected key "features" of input to include only valid features.`

but agree that better to use `get` method here

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -1014,6 +1014,27 @@ def get_destination_region(destination):
         dest_project_id = dxpy.WORKSPACE_ID
     return dxpy.api.project_describe(dest_project_id, input_params={"fields": {"region": True}})["region"]
 
+def verify_nf_license(destination, extra_args):
+    
+    def get_project_to_check():
+        # extra args overrides the destination argument
+        # so we're checking it first
+        if "project" in extra_args:
```
YuxinShi0423, line 1022: corner case here: what if in extra_args we have `destination` instead?

commandlinegirl: It will be ignored since it's not a valid input to [/applet/new](https://documentation.dnanexus.com/developer/api/running-analyses/applets-and-entry-points#api-method-applet-new).

src/python/dxpy/scripts/dx_build_app.py
commandlinegirl: Updated, thx!

src/python/dxpy/scripts/dx_build_app.py
YuxinShi0423: Then just curious: what if the billTo is also overrided using `--bill-to`? do we need to check it?

commandlinegirl: This is an option only applicable to apps :)

PR #913, opened 2023-01-13
APPS-1359 Nextflow - private docker registries support

mhrvol:
At first I wanted to support private repositories on the plugin level, but this solution seems to be easier, as all jobs share the same container.


src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -2,34 +2,38 @@
...
-    echo "        organization": "<(optional, default value equals username) organization as defined by DockerHub or Quay.io>",
-    echo "        token": "<API token>"
-    echo "    }"
-    echo "}"
-    exit 1
-fi
-
-jq '.docker_registry.token' "$CREDENTIALS" -r | docker login $REGISTRY --username $REGISTRY_USERNAME --password-stdin 2> >(grep -v -E "WARNING! Your password will be stored unencrypted in |Configure a credential helper to remove this warning. See|https://docs.docker.com/engine/reference/commandline/login/#credentials-store")
+
+dx-registry-login() {
```
commandlinegirl, line 9: Please follow the convention used for function names in this script (it seems it should be `dx_registry_login`). Although, a more informative name would be `docker_registry_login`

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -2,34 +2,38 @@
...
+  if [ ! -f $CREDENTIALS ]; then
+    dx download "${DOCKER_CREDS_FOLDER}${DOCKER_CREDS_FILENAME}" -o $CREDENTIALS
+  fi
+  command -v docker >/dev/null 2>&1 || (echo "ERROR: docker is required when running with the Docker credentials."; exit 1)
```
commandlinegirl, line 13: We can safely assume docker is available in workers, so we can remove this check.

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -2,34 +2,38 @@
...
+      echo "    docker_registry: {"
+      echo "        registry": "<Docker registry name, e.g. quay.io or docker.io>",
+      echo "        username": "<registry login name>",
+      echo "        organization": "<(optional, default value equals username) organization as defined by DockerHub or Quay.io>",
+      echo "        token": "<API token>"
+      echo "    }"
+      echo "}"
+      exit 1
+  fi
+
```
commandlinegirl, line 35: Please add a comment what it does (does not print the warning).

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -428,7 +433,8 @@ nf_task_exit() {
 }
 
 nf_task_entry() {
-  if [ -n "$docker_creds" ]; then
+  docker_credentials=$(dx find data --path "$DOCKER_CREDS_FOLDER" --name "$DOCKER_CREDS_FILENAME")
```
commandlinegirl, line 436: How can we avoid this API call, esp since it doesn't qualify the search with a project/container? Is it not enough to check if the docker_creds input was provided (`if [ -n "$docker_creds" ]; then`)? Was the motivation to check that this input file was indeed downloaded into the workspace? But if it was not, that means that something went wrong anyway (the user provided the creds but they are not in the workspace) so perhaps it's better to throw an error here early?

commandlinegirl (changes requested):

commandlinegirl:
Please also make the PR/commit title more clear (it's about `private Docker registries`), thank you!

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
mhrvol: No really, this is task entry and it does not have docker_creds in the input, this is the reason why I saved the file to the container.

With Yih-Chii, the general premise is not to polute the input space of subjob.

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
mhrvol: I added a comment to this function, is this what you meant by this comment?

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
mhrvol: Btw this search is done only within a container, isnt it? Otherwise --all-projects would have to be selected, right? We also search only through certain folder - which should reduce the weight of this api call

commandlinegirl: It could be either done with the container context or the project context, so this is either something to be checked or, if possible, if you could qualify the search space with WORKSPACE_ID?

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
commandlinegirl: Looks good, thanks for better formatting.

I meant the line 46. It was supposed to not print the standard waring that Docker prints. Is it printed in the logs?

commandlinegirl (changes requested):

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
mhrvol: It is not printed, this code was previously reused from cwic. The thing is, nobody else can access the file, only someone who has access to the job, thus I believe there is no reason to scare the user, do you think we should print this warning?

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
mhrvol: Added DX_WORKSPACE_ID to the search.

commandlinegirl (approved):
LGTM

commandlinegirl:
Please, as I mentioned before, remember to update the title and set the PR commit title to something more clear.

PR #920, opened 2023-01-25
APPS-1611 Fix feature switch check for Nextflow pipeline build

commandlinegirl:
Details are provided in the ticket.

### DIFF
--- a/src/python/dxpy/nextflow/nextflow_builder.py
+++ b/src/python/dxpy/nextflow/nextflow_builder.py
@@ -65,7 +65,7 @@ def build_pipeline_from_repository(repository, tag, profile="", git_creds=None, 
         parser.error(
             "Can't create an applet without specifying a destination project; please use the -d/--destination flag to explicitly specify a project")
 
-    nf_builder_job = dxpy.DXApp("app-GP8gZ8j0FKzB35v0yFzgQxb9").run(app_input=input_hash, project=build_project_id,
+    nf_builder_job = dxpy.DXApp(name=get_importer_name()).run(app_input=input_hash, project=build_project_id,
                                                               folder=build_folder,
                                                               name="Nextflow build of %s" % (repository), detach=True)
 
--- a/src/python/dxpy/nextflow/nextflow_builder.py
+++ b/src/python/dxpy/nextflow/nextflow_builder.py
@@ -65,7 +65,7 @@ def build_pipeline_from_repository(repository, tag, profile="", git_creds=None, 
         parser.error(
             "Can't create an applet without specifying a destination project; please use the -d/--destination flag to explicitly specify a project")
 
-    nf_builder_job = dxpy.DXApp(name=get_importer_name()).run(app_input=input_hash, project=build_project_id,
+    nf_builder_job = dxpy.DXApp("app-GP8gZ8j0FKzB35v0yFzgQxb9").run(app_input=input_hash, project=build_project_id,
                                                               folder=build_folder,
                                                               name="Nextflow build of %s" % (repository), detach=True)
 
--- a/src/python/dxpy/scripts/dx_build_app.py
+++ b/src/python/dxpy/scripts/dx_build_app.py
@@ -1143,7 +1143,7 @@ def _build_app(args, extra_args):
         if not args.check_syntax:
             more_kwargs['do_check_syntax'] = False
         if args.nextflow and args.repository is not None:
-            # Feature switch if verified remotely
+            # Feature switch if verified remotely so we won't be doing it here
             return build_pipeline_from_repository(args.repository, args.tag, args.profile, args.git_credentials,
                                                   args.brief, args.destination, extra_args)
         app_json = _parse_app_spec(source_dir)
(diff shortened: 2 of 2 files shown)


src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -1022,14 +1022,26 @@ def get_project_to_check():
         if "project" in extra_args:
             return extra_args["project"]
         if destination:
```
YuxinShi0423, line 1024: here I think since destination already has a default value, we can skip the if condition

also, I would suggest not changing this function, but instead checking it right before we launch the api call.

commandlinegirl: > here I think since destination already has a default value, we can skip the if condition
> 
> also, I would suggest not changing this function, but instead checking it right before we launch the api call.

You mean, right before the applet-xxxx/build?

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -1022,14 +1022,26 @@ def get_project_to_check():
...
+            else:
+                dest_project_id, _, _ = parse_destination(destination)
             return dest_project_id
         else:
             return dxpy.PROJECT_CONTEXT_ID
 
     dest_project_to_check = get_project_to_check()
     features = dxpy.DXHTTPRequest("/" + dest_project_to_check + "/checkFeatureAccess", {"features": ["dxNextflow"]}).get("features", {})
     # Expecting output {'features': {'dxNextflow': True}}
+
```
YuxinShi0423, line 1041: Duplicate lines?

commandlinegirl: Removed duplication, thank you!

src/python/dxpy/scripts/dx_build_app.py
YuxinShi0423: Oh sorry I need to correct it a little here:

I was wondering if we happen to use destination something like `:` which is resolved to a container id, then the following if condition will not be satisfied. 
`if destination == "." or destination.startswith("container-"):`

so I would suggest, maybe we still parse_destination and let this function return the get_project_to_check, then before the checkFeatureAccess request, we check if we are in a execution env with a DX_JOB_ID already set and this "project_to_check" is a valid one

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -1022,15 +1022,21 @@ def get_project_to_check():
...
+                dest_project_id, _, _ = parse_destination(destination)
             return dest_project_id
         else:
             return dxpy.PROJECT_CONTEXT_ID
 
     dest_project_to_check = get_project_to_check()
-    features = dxpy.DXHTTPRequest("/" + dest_project_to_check + "/checkFeatureAccess", {"features": ["dxNextflow"]}).get("features", {})
-    # Expecting output {'features': {'dxNextflow': True}}
-    
+    features = dxpy.DXHTTPRequest("/" + dest_project_to_check + "/checkFeatureAccess", {"features": ["dxNextflow"]}).get("features", {})
```
YuxinShi0423, line 1039: > .. envvar:: DX_JOB_ID
   Should only be present if run in an Execution Environment; indicates
   the ID of the currently running job.

```suggestion
    if "DX_JOB_ID" in environ:
        dest_project_to_check = dxpy.PROJECT_CONTEXT_ID
        
    features = dxpy.DXHTTPRequest("/" + dest_project_to_check + "/checkFeatureAccess", {"features": ["dxNextflow"]}).get("features", {})    
```

YuxinShi0423: or
    ```
    if dxpy.JOB_ID:
        dest_project_to_check = dxpy.PROJECT_CONTEXT_ID
    ````

src/python/dxpy/nextflow/nextflow_builder.py
```diff
@@ -65,7 +65,7 @@ def parse_extra_args(extra_args):
         parser.error(
             "Can't create an applet without specifying a destination project; please use the -d/--destination flag to explicitly specify a project")
 
-    nf_builder_job = dxpy.DXApp(name=get_importer_name()).run(app_input=input_hash, project=build_project_id,
+    nf_builder_job = dxpy.DXApp("app-GP8gZ8j0FKzB35v0yFzgQxb9").run(app_input=input_hash, project=build_project_id,
```
mhrvol, line 68: What is the reason behind using app id instead of "nextflow_pipeline_importer" ?

src/python/dxpy/scripts/dx_build_app.py
mhrvol: What if the destination is a different project? I.e. if somebody runs this from jupyterlab

I think being a little bit defensive and parsing the destination does not hurt.

src/python/dxpy/nextflow/nextflow_builder.py
commandlinegirl: Good question. I used it for the purpose of testing the file-apps APPS-1611 only and added an explanation to the [description](https://github.com/dnanexus/file-apps/pull/708) and the ticket.. I'll revert the commit so that it is back to `nextflow_pipeline_importer`.

PR #922, opened 2023-01-27
DEVEX-2195 Fix tests after turning off workflowStageReuse

commandlinegirl:


src/python/test/test_dxclient.py
```diff
@@ -3710,7 +3696,7 @@ def test_dx_run_clone_analysis(self):
         finally:
             run("dx rmproject -y " + other_proj_id)
 
-    @unittest.skipUnless(testutil.TEST_RUN_JOBS, 'skipping test that runs jobs')
+    @unittest.skip("Skipping per DEVEX-2195")
```
YuxinShi0423, line 3699: @commandlinegirl shall we create another ticket to correct these tests later? (since these tests also have other functionality tested)

Other than that looks good to me. Thanks!

YuxinShi0423 (approved):

src/python/test/test_dxclient.py
commandlinegirl: Yes, doing it now.

PR #925, opened 2023-02-02
APPS-1614 release automation for Nextaur

mhrvol:


src/python/setup.py
```diff
@@ -80,6 +80,10 @@ def make_valid_pypi_version(raw):
     directory = directory[len("dxpy/templating/"):]
     template_files.extend([os.path.join(directory, _file) for _file in files])
 
+files = os.listdir("dxpy/nextflow")
```
commandlinegirl, line 83: I'd rename files into nextflow_files for clarity.

src/python/dxpy/nextflow/nextflow_utils.py
```diff
@@ -62,50 +64,20 @@ def get_regional_options(region):
     return regional_options
 
 def get_nextflow_assets(region):
+    basepath = path.dirname(__file__)
```
commandlinegirl, line 67: Shouldn't it be dxpy.__file__?

commandlinegirl (dismissed):

commandlinegirl:
Please run integration tests on this branch before merging. Thank you!

src/python/dxpy/nextflow/nextflow_utils.py
mhrvol: Why dxpy.__file__? I was trying to point to the current file, which is __file__ regardless of where its location is, or is there something I am missing?

YuxinShi0423: I guess Ola is suggesting to use the same code for getting the template
https://github.com/dnanexus/dx-toolkit/blob/70a583e20a6eeb5a9d4e74e165877b00314c6804/src/python/dxpy/nextflow/nextflow_utils.py#L29

src/python/dxpy/nextflow/nextflow_utils.py
```diff
@@ -62,50 +64,20 @@ def get_regional_options(region):
...
-            "record-GP5B3p2Kb446439821Qqj02X",
-            "record-GKv1892K6Gkk0K53GFj415XZ")
-    }
-
+    nextaur_assets = path.join(basepath, "nextaur_assets.json")
+    nextflow_assets = path.join(basepath, "nextflow_assets.json")
     try:
-        dxpy.describe(prod_assets[region][0], fields={})
-        return prod_assets[region]
+        with open(nextaur_assets, 'r') as nextaur_f, open(nextflow_assets, 'r') as nextflow_f:
```
YuxinShi0423, line 72: Can we use https://github.com/dnanexus/dx-toolkit/blob/b92005049a92a6260a61043cea0d9c13924c54f4/src/python/dxpy/__init__.py#L979
to figure out whether we are on prod or stag?

mhrvol: I believe this works only if you are not in a running job, that's why I chose to do it this way.

src/python/dxpy/nextflow/nextflow_utils.py
mhrvol: Ah, right, fixed to stay consistent (it had the same functionality, just dxpy.__file__ points to a different file)

src/python/dxpy/nextflow/nextflow_utils.py
YuxinShi0423: then how about using `dxpy.APISERVER_HOST` directly?

mhrvol: That still does not return 'staging' or 'prod', just a proxy IP in case of running from a job.

I've discussed this with Ola and I dont believe we have a convincing way of figuring this out.

YuxinShi0423: Got it. But if we could use env `DX_AUTHSERVER_HOST` in job exec env we could possibly avoid the call. But that could just be a minor improvement.

YuxinShi0423 (approved):
@mhrvol If you got a chance to rerun the test it would be great. But everything looks good :D

PR #926, opened 2023-02-06
APPS-1625 Nextflow - update azure instances 

mhrvol:


mhrvol:
tests: https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/305/

src/python/dxpy/nextflow/nextflow_utils.py
```diff
@@ -50,7 +50,7 @@ def get_regional_options(region):
         region: {
```
YuxinShi0423, line 50: Minor: maybe it is better looking this way:
```
    nextflow_asset, nextaur_asset = get_nextflow_assets(region)
    regional_instance_type = get_instance_type(region)
    
    regional_options = {
        region: {
            "systemRequirements": {
                "*": {
                    "instanceType": regional_instance_type
                }
            },
            "assetDepends": [
                {"id": nextaur_asset},
                {"id": nextflow_asset}
            ]
        }
    }

Comment (YuxinShi0423) [reply src/python/dxpy/nextflow/nextflow_utils.py:50]: Actually since systemRequirements is optional, I do wonder if we need it at build time. Because the user might not realize that it is specified here, and they actually want the default to be specified based on for example their own project/billto settings. cc @yihchii @commandlinegirl

File: CHANGELOG.md [hunk 2]
```diff
@@ -6,6 +6,10 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Unreleased
 
+### Changed
+
+* Nextflow - updated default instance types in Azure region
```
YuxinShi0423, line 11:
```suggestion
* Nextflow - updated default instance types based on destination region
```

src/python/dxpy/nextflow/nextflow_utils.py
mhrvol: We have defaults in all of our apps in file-apps - but I agree that it might be different in the future. 

I recommend we merge this, but discuss this next week with Ola.

YuxinShi0423 (approved):
Looks good to me. Sure we can refine the whole template later. Thanks! @mhrvol

PR #931, opened 2023-02-11
DEVEX-2216 Ubuntu 22.04 + py3.10 'make python' fix

billagee:
Looks like we need to upgrade xattr to build on 22.04. Also, since python 3.10 is the default python3 on 22.04, it was revealed that we expected the python minor release to be a single digit.


wormsik (approved):

billagee:
@wormsik @kpjensen I ran the following test with this branch and verified it passes ... the test runs an applet that uses `dx-upload-all-outputs` and `xattr`, so I believe we're good to merge:

```
test_dx_bash_helpers.TestDXBashHelpers.test_xattr_parameters
```

Interestingly, the xattr_properties test applet has already been testing xattr 0.10.1 for some time, since it runs `python3 -m pip install -U xattr` to install the latest version of the package:

https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/file_load/xattr_properties/run.sh#L5

PR #955, opened 2023-03-06
DEVEX-2184 Explicitly request spotCostSavings field

wormsik:
* `spotCostSavings` should not be returned by default and we need to request it explicitly


wormsik:
will do later in a new issue

PR #984, opened 2023-03-24
DEVEX-2216 Ubuntu 22.04 + py3.10 'make python' fix

billagee:
This change updates the `make python` target to build properly on 22.04 systems that have the `python-is-python3` apt package installed (and thus have python 3.10 as the base system python).

I've also updated the README instructions with the list of apt dependencies to install on 22.04 in order to build the default `make` target on that platform.

Also, `.github/workflows/toolkit-build-ubuntu-2204.yml` was added to let us quickly test toolkit builds in a 22.04 container.


wormsik (approved):
LGTM 👍

kpjensen (approved):

PR #990, opened 2023-03-30
Sciprod 1216 finalize combined code

JoeMulka:
Switches to python-only input validator, adds docstrings to functions, renames turbo filter, removes "main" section of filter converter, cleans up column_conversion and column_condition files


keegankelsey (approved):

Niranjanpandeshwar (approved):
LGTM! Thanks Joe!

PR #997, opened 2023-04-04
DEVEX-2263 Fix ImportError in test_dxpy.py

billagee:
Removing pyopenssl from requirements_test.txt for now, since it was only used by the disabled test `test_ssl_options` - and our pinned pyopenssl version was causing an import failure in test_dxpy.py.

### DIFF
--- a/src/python/requirements_test.txt
+++ b/src/python/requirements_test.txt
@@ -1,5 +1,5 @@
 pexpect==4.6
-pyopenssl==17.5.0
+pyopenssl==23.1.1
 mock==2.0.0
 pytest==4.6.9
 pytest-xdist==1.31.0
--- a/src/python/requirements_test.txt
+++ b/src/python/requirements_test.txt
@@ -1,5 +1,5 @@
 pexpect==4.6
-pyopenssl==23.1.1
+pyopenssl==22.1.0
 mock==2.0.0
 pytest==4.6.9
 pytest-xdist==1.31.0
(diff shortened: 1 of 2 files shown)


src/python/requirements_test.txt
```diff
@@ -1,5 +1,5 @@
 pexpect==4.6
-pyopenssl==17.5.0
+pyopenssl==22.1.0
```
wormsik, line 2: can we please bump it to `23.1.1` so it is compatible with `cryptography 40.0.1` requested by dx-toolkit?

billagee: I'll take a look at this 👍

wormsik: Thanks...and do we know why we actually need PyOpenSSL? it is not required by dxpy and in requests/urllib3 is used just as a fallback backend for Python 2.X without SNI.

billagee: Good point, it would be great if we could simply remove PyOpenSSL from this file. I'll try it out - I don't recall why this package was needed in py2.7 test environments; hopefully a modern py3.x machine no longer requires it when running the test suite.

billagee: @wormsik I believe this is ready for review now, if you don't mind taking a look - thanks! It does indeed look like we can remove pyopenssl from the test requirements. We'll probably only need it again someday if we decide to resurrect `test_ssl_options` in test_dxpy.py (and there isn't an alternate way to implement the test).

wormsik: Thanks for checking and solving this! I hope those tests may be eventually rewritten to expect only `urllib3` internal exceptions.

wormsik (approved):
LGTM 👍 

Requesting also review from @kpjensen just to be sure.

kpjensen (approved):

PR #1043, opened 2023-05-17
using argparse mutually exclusive group to ensure filter mutex

JoeMulka:
Hi Folks,
This PR replaces the logic for ensuring only one filter is given at a time with an argparse mutually exclusive argument group.  I was able to find examples of mutex groups being added to subparsers elsewhere in dx.py, and set it up similarly for extract assay germline.  I also created a test for it in the test_extract_assay script, which demonstrates the condition being caught.


wormsik (approved):
LGTM 👍 

Not sure if relevant to this PR, but I would be very happy to get rid of `args_list` completely and do all checks using standardised `args.<argument>` approach without necessity to hardcode parameter names as strings.

PR #1052, opened 2023-05-30
Sciprod 1401 combine args error msgs

JoeMulka:
Hi folks,
Here's the change that Niranjan and I discussed, it removes a piece of the conditional block that could never be reached, and combines two error messages into one


Niranjanpandeshwar (approved):
LGTM!

PR #1053, opened 2023-05-31
DEVEX-2277 Access to restarted jobs

wormsik:


src/python/dxpy/utils/job_log_client.py
```diff
@@ -220,6 +226,9 @@ def received_message(self, message):
         else:
             print(self.msg_output_format.format(**message_dict))
 
+    def _describe_job(self, job_id):
+        return dxpy.api.job_describe(job_id, {'try': self.job_try} if self.job_has_try else {})
```
kpjensen, line 230: What fields are needed from the describe other than `state`? We can just request them with a `"fields":{}` param here so all of the default ones are not returned.

kpjensen: Ignore this for now as there is an issue on the backend with combining those two inputs, can be revisited after.

kpjensen (approved):

PR #1061, opened 2023-06-06
forcing quotechar to be interpeted as a string by python 2

JoeMulka:
Hi Folks,
This change forces the python2 interpreter to interpret the quotechar in the csv_from_json function to be interpreted as a str type and not as unicode.  I tested both python2 and python3 and all unit tests pass.
More details in the ticket description [here](https://jira.internal.dnanexus.com/browse/SCIPROD-1413)


keegankelsey (approved):
thanks @JoeMulka

PR #1095, opened 2023-06-30
SCIPROD-1505 extract assay somatic info field quotechar

csuloway-dnanexus:
Updated quotechar in extract_assay somatic variant output from "|" to "" with no quoting.


Niranjanpandeshwar (approved):
LGTM

PR #1111, opened 2023-07-11
SCIPROD-1442 missing rsid handling

JoeMulka:
Hi folks,
This creates a friendlier error messaging for the case where an rsid is provided in the filter that isn't present in the dataset.


keegankelsey (approved):

src/python/test/test_extract_assay.py
```diff
@@ -87,9 +88,7 @@ def test_basic_filter_allele(self):
         }
 
         self.assertEqual(
-            basic_filter(
-                table, friendly_name, values, self.proj_id, genome_reference
-            ),
+            basic_filter(table, friendly_name, values, self.proj_id, genome_reference),
```
Niranjanpandeshwar, line 91: Please remove the linter changes here and other places if not necessary! Otherwise LGTM!

PR #1124, opened 2023-07-17
APPS-5819 Print TAT only if returned in describe output for apps and workflows

kpjensen:


Gvaihir (approved):
Ok, thanks for the fix!

PR #1127, opened 2023-07-18
APPS-1859 Tree turnaround time threshold HOTFIX

Gvaihir:
(fix) dxapplet.py - missing attr and failing test

### DIFF
--- a/src/python/dxpy/bindings/dxapplet.py
+++ b/src/python/dxpy/bindings/dxapplet.py
@@ -285,7 +285,7 @@ class DXExecutable:
 _applet_required_keys = ['name', 'title', 'summary', 'types', 'tags',
                          'httpsApp', 'properties', 'dxapi', 'inputSpec', 'outputSpec',
                          'runSpec', 'access', 'details']
-_applet_optional_keys = ['ignoreReuse']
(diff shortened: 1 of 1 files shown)


src/python/dxpy/bindings/dxapplet.py
```diff
@@ -285,7 +285,7 @@ def run(self, executable_input, project=None, folder=None, name=None, tags=None,
 _applet_required_keys = ['name', 'title', 'summary', 'types', 'tags',
                          'httpsApp', 'properties', 'dxapi', 'inputSpec', 'outputSpec',
                          'runSpec', 'access', 'details']
-_applet_optional_keys = ['ignoreReuse']
+_applet_optional_keys = ['ignoreReuse', 'treeTurnaroundTimeThreshold']
```
kpjensen, line 288: Should this also be appended to the optional_keys list in dxapp.py?

Gvaihir: it is [there](https://github.com/dnanexus/dx-toolkit/blob/016d6a51947d167c2f171b3089e1146fa339f2fd/src/python/dxpy/bindings/dxapp.py#L64), I just missed it in the applet

kpjensen (approved):

PR #1151, opened 2023-08-23
SCIPROD-1640, SCIPROD-1642: [CLIEXPRESS] [UR-9] sub-parser and [UR-1, UR-2, UR-3] inputs validation

gabriela-reznik:


gabriela-reznik:
Team, some of the output messages were set automatically and are not exactly as the spec but I believe they're clear enough. We can connect about it later.

The following questions came up during coding:
1.  In the spec, "--json-help" was used in combination with "--retrieve-expression". Should we allow it to also be passed without?
2. In contrast, "--additional-fields-help" is set to be passed without "--retrieve-expression" and should not accept this (or any other) argument. I wonder why points 1 and 2 have such distinguished handlings.

Thank you in advance for your comments :)

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,179 @@
...
+    InvalidState,
+    ResourceNotFound,
+    default_expected_exceptions,
+)
+from ..utils.resolver import resolve_existing_path, is_hashid, ResolutionError
+
+import json
+import os
+
+
```
ashkan0x, line 16: I believe lines 1-16 should be deleted?

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,179 @@
...
+                    end_arg.replace("_", "-")
+                )
+            )
+
+    def forbiden_combination_validation(self, invalid_combination, error_message):
+        if invalid_combination:
+            self.error_handler = error_message
+
+
+# class PathValidator:
```
ashkan0x, line 88: Are these still needed? (lines 88-179)

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,179 @@
...
+        self.input_json_file = args.input_json_file
+        self.json_help = args.json_help
+        self.sql = args.sql
+        self.additional_fields = args.additional_fields
+        self.additional_fields_help = args.additional_fields_help
+        self.expression_matrix = args.expression_matrix
+        self.delim = args.delim
+        self.output = args.output
+
+        self.validations_to_run()
```
ashkan0x, line 36: `__init__ should be used to only initialize the object's attributes. So I suggest removing this line entirely, and then explicitly calling the `validations_to_run` method on an instantiated object of the class.

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1043,6 +1048,20 @@ def print_fields(fields):
                 quoting=csv.QUOTE_NONE,
             )
 
+def extract_assay_expression(args):
+    """
+    Retrieve the selected data or generate SQL to retrieve the data from an expression assay in a dataset or cohort based on provided rules.
+    """
+
+    # Validating input combinations
+    input_validator = ExpressionInputsValidator(args)
```
ashkan0x, line 1057: Class object was instantiated but none of the methods were called. Please see my other comment below. Ideally, here the code should call a method, such as `input_validator.validations_to_run`

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1043,6 +1048,20 @@ def print_fields(fields):
...
+    if input_validator.error_handler:
```
ashkan0x, line 1058: It would be a lot safer if this is entirely handled inside the class, and `err_exit` can be passed as an argument to `ExpressionInputsValidator`.

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,179 @@
...
+
+import json
+import os
+
+
+class ExpressionInputsValidator:
+    """InputsValidator class for extract_assay expresion. Checks for invalid input combinations"""
+
+    def __init__(self, args):
+        self.error_handler = None
```
ashkan0x, line 21: This could possibly have unintended consequences. Best to add `error_handler` explicitly as a parameter to `__init__` so that it can be safely overriden.

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,179 @@
...
+        invalid_combination = getattr(self, end_arg) and any(args_specific_list)
+
+        if invalid_combination:
+            self.error_handler = (
+                '"--{}" cannot be presented with other options.'.format(
+                    end_arg.replace("_", "-")
+                )
+            )
+
+    def forbiden_combination_validation(self, invalid_combination, error_message):
```
ashkan0x, line 83: I would personally get rid of this method entirely and simply let all messages be handled by the `error_handler`.

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
gabriela-reznik: Sorry about this, I was coding the path validation and then I realised this is not for this ticket. Either ways I figured I could leave it there commented out for now. Let me know if you prefer that I remove and add it again in the next PR

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,164 @@
+from ..exceptions import err_exit
```
ashkan0x, line 1: This is not needed here, it can be passed directly to the `input_validator = ExpressionInputsValidator` call in `src/python/dxpy/cli/dataset_utilities.py`

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class ExpressionInputsValidator:
+    """InputsValidator class for extract_assay expresion. Checks for invalid input combinations"""
+
+    def __init__(self, args, error_handler):
```
ashkan0x, line 7: Let's have a default value for `error_handler` so that it can print something if none is provided. I recommend Python's own `print` function which will work in both Python 2 and 3

```suggestion
    def __init__(self, args, error_handler=print):
```

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,164 @@
...
+        if self.json_help and any(
+            [
+                self.assay_name,
+                self.sql,
+                self.additional_fields,
+                self.expression_matrix,
+                self.output,
+            ]
+        ):
+            self.error_handler = (
```
ashkan0x, line 36: You can just let error_hadler print the message without overriding its value. 
```suggestion
            self.error_handler(
```

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,164 @@
...
+                '"--json-help" cannot be passed with any of "--assay-name", "--sql", "--additional-fields", "--expression-matrix", or "--output"',
+            )
+
+        if self.retrieve_expression_flag and not any([self.input_json, self.json_help]):
+            self.error_handler = (
```
ashkan0x, line 41: Same thing here, please see my comment above. Since there is already an `if` condition, then let's just use `error_handler` as a method/function and immediately write the message.

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,164 @@
...
+                'The flag "--retrieve_expression" must be followed by a json input or json help.',
+            )
+
+        if self.input_json == "{}":
+            self.error_handler = (
```
ashkan0x, line 46:
```suggestion
            self.error_handler(
```

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,164 @@
...
+        invalid_combination = getattr(self, end_arg) and any(args_specific_list)
+
+        if invalid_combination:
+            self.error_handler = (
+                '"--{}" cannot be presented with other options.'.format(
+                    end_arg.replace("_", "-")
+                )
+            )
+
+    def error_exit(self):
```
ashkan0x, line 68: This method is not needed. `error_handler` is now a method and will be called directly within the class whenever needed.

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1043,6 +1046,21 @@ def print_fields(fields):
...
 
+def extract_assay_expression(args):
+    """
+    Retrieve the selected data or generate SQL to retrieve the data from an expression assay in a dataset or cohort based on provided rules.
+    """
+
+    # Validating input combinations
+    input_validator = ExpressionInputsValidator(args, None)
+    input_validator.run_validations()
+    input_validator.error_exit()
```
ashkan0x, line 1057: This is no longer needed.
```suggestion
    
```

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1043,6 +1046,21 @@ def print_fields(fields):
                 quoting=csv.QUOTE_NONE,
             )
 
+def extract_assay_expression(args):
+    """
+    Retrieve the selected data or generate SQL to retrieve the data from an expression assay in a dataset or cohort based on provided rules.
+    """
+
+    # Validating input combinations
+    input_validator = ExpressionInputsValidator(args, None)
```
ashkan0x, line 1055: We can now just use `err_exit` from `dxpy` here and let it handle the writing of messages to the console.

```suggestion
    input_validator = ExpressionInputsValidator(args, err_exit)
```

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1043,6 +1046,21 @@ def print_fields(fields):
...
+    input_validator.run_validations()
```
jkotrs, line 1056: Or even shorter `.validate()`? :)

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -2,58 +2,55 @@ class ExpressionInputsValidator:
...
-        if self.retrieve_expression_flag and not any([self.input_json, self.json_help]):
-            self.error_handler(
-                'The flag "--retrieve_expression" must be followed by a json input or json help.',
-            )
-
-        if self.input_json == "{}":
-            self.error_handler(
-                'JSON for "--retrieve-expression" does not contain valid filter information.',
-            )
+        invalid_combination_infos = {
```
jkotrs, line 12: @gabriela-reznik I like this declaration style, good job! Maybe just rename "infos" to "definition"?

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -62,6 +59,11 @@ def end_arguments_check(self, end_arg):
                 )
             )
 
+    def invalid_combinations_check(self, invalid_combinations_info):
+        for e in invalid_combinations_info:
+            if invalid_combinations_info[e]["invalid_combination"]:
```
jkotrs, line 64:
```
if e["invalid_combination"]:
```
.. and below ..
```
e["error_message"]
```

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -2,58 +2,55 @@ class ExpressionInputsValidator:
...
+                "error_message": 'The flag "--retrieve_expression" must be followed by a json input or json help.',
+            },
+            "empty_json": {
+                "invalid_combination": self.args["input_json"] == "{}",
+                "error_message": 'JSON for "--retrieve-expression" does not contain valid filter information.',
+            },
+        }
+        self.invalid_combinations_check(invalid_combination_infos)
 
     def end_arguments_check(self, end_arg):
```
jkotrs, line 38: Maybe `exclusive_argument_check`?

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
gabriela-reznik: This doesn't really work cause the for is taking the variable as the strings ("json_help_unique", "retrieve_expression_json" and "empty_json")

src/python/dxpy/dx_extract_utils/ExpressionInputsValidator.py
```diff
@@ -0,0 +1,159 @@
...
+        invalid_combination = self.args[end_arg] and any(args_specific_list)
+
+        if invalid_combination:
+            self.error_handler(
+                '"--{}" cannot be presented with other options.'.format(
+                    end_arg.replace("_", "-")
+                )
+            )
+
+    def invalid_combinations_check(self, invalid_combinations_info):
```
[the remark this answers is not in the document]
gabriela-reznik: yes but I think it makes sense to leave which kind of checking is being done. I created some more general validation methods and with those it might make sense to leave it as is. Let me know your thoughts.

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1043,6 +1046,137 @@ def print_fields(fields):
...
 
+def extract_assay_expression(parser_obj):
+    """
+    Retrieve the selected data or generate SQL to retrieve the data from an expression assay in a dataset or cohort based on provided rules.
+    """
+
+    # Validating input combinations
+    parser_dict = vars(parser_obj)
+
+    schema = {
```
jkotrs: Great! As this is a constant, let's make it one (capital letters, declared at the start of the file).

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -0,0 +1,273 @@
...
+    def validate_schema_conditions(self):
+        # Checking if all conditions exist
+        present_conditions = [
+            value.get("condition")
+            for key, value in self.schema.items()
+            if key != "schema_version" and key != "parser_args"
+        ]
+        not_found = set(present_conditions) - set(self.conditions_funcs)
+        if len(not_found) != 0:
+            self.error_handler("{} schema condition is not defined".format(not_found))
```
jkotrs, line 74: Excellent!

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -0,0 +1,273 @@
...
+    ### Checking general methods ###
+    def interpret_conditions(self):
+        for key, value in self.schema.items():
+            if key != "schema_version" and key != "parser_args":
+                condition = value.get("condition")
+                method_to_call = self.conditions_funcs.get(condition)
+                getattr(self, method_to_call)(key)
+
+    def throw_message(self, check):
+        type = self.schema.get(check).get("error_message").get("type")
```
jkotrs, line 85: You can save the retrieved "check" object and pass that one down. Or better, extract the message into a variable also and pass just the message to the `throw_` method. That provides the lowest scope possible for the downstream method and in turn it's testing and usage becomes easier (str vs object).

jkotrs (approved):
Great job with the Validator and schema!

PR #1152, opened 2023-08-23
APPS-1728 add all nextflow applet inputs explicitly

YuxinShi0423:


YuxinShi0423:
https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/510/

src/python/dxpy/nextflow/nextflow_templates.py
```diff
@@ -66,33 +66,21 @@ def get_nextflow_src(custom_inputs=None, profile=None, resources_dir=None):
...
-            required_runtime_params = required_runtime_params + '''
-            if [ -n "${}" ]; then
-                required_inputs="${{required_inputs}} --{}={}" 
-            fi
-            '''.format(i['name'], i['name'], value)
+
+        # applet_runtime_inputs variable is initialized in the nextflow.sh script template
+        applet_runtime_params = applet_runtime_params + '''
+        if [ -n "${}" ]; then
+            applet_runtime_inputs="${{applet_runtime_inputs}} --{}={}"
```
mhrvol, line 78: What was the reason of having the required parameters separately from the optional ones before? I remember we talked about it a lot due to some edge cases

mhrvol (commented):
Is there a test run for python2 as well?

src/python/dxpy/nextflow/nextflow_templates.py
YuxinShi0423: It was in  APPS-1390 where we tried to handle this case where we try to add this default value parsed from nextflow.config
```
ribo_database_manifest     = "${projectDir}/assets/rrna-db-defaults.txt"
```
Adding this explicitly would not work
```
nextflow run ...
'--ribo_database_manifest=${projectDir}/assets/rrna-db-defaults.txt' 
```
because the projectDir is not resolved when resolving this explicit param. Only when we added it to a runtime config it would work since projectDir is a configuration implicit variable
https://www.nextflow.io/docs/latest/script.html?highlight=projectdir#configuration-implicit-variables

However 
1. passing params via -c is not recommended in DSL 2 
2. this option is no longer needed in the nextflow run command, as we already make all schema inputs as OPTIONAL applet input. Thus when one schema option is not specified by user, its default value would be honored.
@mhrvol

YuxinShi0423:
> Is there a test run for python2 as well?

https://jenkins-vstg.internal.dnanexus.com/job/dx-toolkit/job/dxpy-master-integration-tests-on-staging-16.04-python2/2306/

src/python/dxpy/nextflow/nextflow_templates.py
mhrvol: Sounds good. Thanks for clarification!

r-i-v-a (approved):

PR #1157, opened 2023-08-24
SCIPROD-1664 Input JSON Validation

ashkan0x:


src/python/dxpy/bindings/apollo/ValidateJSONbySchema.py
```diff
@@ -0,0 +1,165 @@
...
+            {"nested_key1": "3", "nested_key2": "4"},
+            {"nested_key1": "5"}
+        ],
+        "id": ["id_1", "id_2"]
+    }
+
+    """
+
+    def __init__(self, schema):
+        self.schema = schema
```
jkotrs, line 77: We need to check for `None` schema as downstream none-None is assumed.

src/python/dxpy/bindings/apollo/ValidateJSONbySchema.py
```diff
@@ -0,0 +1,165 @@
...
+
+        self.check_incompatible_keys(input_json)
+        self.check_dependent_key_combinations(input_json)
+
+    def validate_properties(self, properties, input_dict):
+        for key, value in properties.items():
+            if key not in input_dict and value.get("required"):
+                err_exit(
+                    "Required key '{}' was not found in the input JSON.".format(key)
+                )
```
jkotrs, line 101: For this and following `err_exit` calls I would recommend to utilize the delegator pattern. In the class constructor you would accept a method object which will be called instead of `err_exit`. Then the actual `err_exit` call would be defined outside, hopefully closer to the main flow. Also easily replaceable for the whole class as a behaviour.

For example the usage can get as easy as:
```
validator = JSONValidator(schema, err_exit)
``` 
But would allow for easy modification should the error reporting change:
```
def error_handler(message):
    LOG.error(message)
    err_exit(message)
validator = JSONValidator(schema, error_handler)
```

src/python/dxpy/bindings/apollo/assay_filtering_json_schemas.py
```diff
@@ -0,0 +1,30 @@
...
+        "type": dict,
+        "properties": {
+            "feature_name": {"type": list, "required": False},
+            "feature_id": {"type": list, "required": False},
+        },
+        "conflicting_keys": [["feature_name", "feature_id"]],
+    },
+    "expression": {
+        "type": dict,
+        "properties": {"min_value": {"type": str}, "max_value": {"type": str}},
```
jkotrs, line 12: We might want to consider introducing custom types here (like `numeric`). Also is it possible to receive an actual `int`? In which case a poly-type definition (like `"type": [str, int]`)

jkotrs (commented):
Great work Ashkan! A few comments

src/python/dxpy/bindings/apollo/ValidateJSONbySchema.py
gabriela-reznik: That's a very nice idea. I'll implement to my classes as well

src/python/dxpy/bindings/apollo/assay_filtering_json_schemas.py
ashkan0x: We only need `str` for CLIEXPRESS. Otherwise, while it's possible to use any type with this `class` (beucase implementation is a Python dict), we will just error out with the more basic JSON checks/parsing if all keys and values are not `str` (wrapped in `"`).

src/python/dxpy/bindings/apollo/ValidateJSONbySchema.py
ashkan0x: Done. Please see https://github.com/dnanexus/dx-toolkit/pull/1157/commits/9b336e65ec0c508aaf1619ac09d98e3208fb42a3

src/python/dxpy/bindings/apollo/ValidateJSONbySchema.py
ashkan0x: Addressed here: https://github.com/dnanexus/dx-toolkit/pull/1157/commits/9b336e65ec0c508aaf1619ac09d98e3208fb42a3

jkotrs (approved):

PR #1161, opened 2023-08-25
SCIPROD-1614, SCIPROD-1615, SCIPROD-1679 Added optimization logic to dx extract_assay somatic, bug fix

Niranjanpandeshwar:
Added optimization logic to dx extract_assay somatic

### DIFF
--- a/src/python/dxpy/cli/dataset_utilities.py
+++ b/src/python/dxpy/cli/dataset_utilities.py
@@ -607,14 +607,14 @@ def get_assay_name_info(
             for ga in target_assays:
                 if ga["name"] == assay_name:
                     selected_assay_id = ga["uuid"]
-
-    selected_ref_genome = "GRCh38.92"
     
     if friendly_assay_type == "germline":
+        selected_ref_genome = "GRCh38.92"
         for a in target_assays:
             if a["name"] == selected_assay_name and a["reference_genome"]:
                 selected_ref_genome = a["reference_genome"]["name"]
     elif friendly_assay_type == "somatic":
+        selected_ref_genome = ""
         for a in target_assays:
             if a["name"] == selected_assay_name and a["reference"]:
                 selected_ref_genome = a["reference"]["name"] + "." + a["reference"]["annotation_source_version"]
--- a/src/python/dxpy/cli/dataset_utilities.py
+++ b/src/python/dxpy/cli/dataset_utilities.py
@@ -612,7 +612,7 @@ def get_assay_name_info(
         selected_ref_genome = "GRCh38.92"
         for a in target_assays:
             if a["name"] == selected_assay_name and a["reference_genome"]:
-                selected_ref_genome = a["reference_genome"]["name"]
+                selected_ref_genome = a["reference_genome"]["name"].split(".", 1)[1]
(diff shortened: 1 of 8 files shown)


src/python/dxpy/dx_extract_utils/somatic_filter_payload.py
```diff
@@ -33,6 +32,33 @@
...
+
+def retrieve_geno_bins(list_of_genes, project, genome_reference, friendly_name):
+    """
+    A function for determining appropriate geno bins to attach to a given SYMBOL, Gene or Feature
+    """
+    stage_file = "Homo_sapiens_genes_manifest_staging_vep.json"
+    platform_file = "Homo_sapiens_genes_manifest_vep.json"
+    error_message = "Following symbols, genes or features are invalid"
+    genome_reference = genome_reference + "_" + friendly_name
+    geno_bins = retrieve_bins(list_of_genes, project, genome_reference, extract_utils_basepath,
```
piader1, line 52: Do we support variable bin size here?
I tought that this method returns bins equivalent to the bins in the database, instead this returns list of gene positions like:
`
... ,
{
    "chr": "1",
    "start": "1471765",
    "end": "1497848",
    "strand": "+"
  },
...
`
Right? 
So maybe we can reffer to this as get genomics intervals?

Niranjanpandeshwar: `geno_bins` is a key in the payload for `RAW` API. Named the variable as `geno_bins` to be consistent with that.

keegankelsey (approved):

durae (commented):
LGTM. I can't see any red flag and I assume that the code generates the payload as discussed in the previous ticket.

Niranjanpandeshwar:
> LGTM. I can't see any red flag and I assume that the code generates the payload as discussed in the previous ticket.

Correct. There is no compound object and all filters go into one hash of `filters`.

ksoni (commented):
Looks good to me.

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -607,13 +607,17 @@ def get_assay_name_info(
...
                     selected_assay_id = ga["uuid"]
-
-    selected_ref_genome = "GRCh38.92"
     
     if friendly_assay_type == "germline":
+        selected_ref_genome = "GRCh38.92"
         for a in target_assays:
             if a["name"] == selected_assay_name and a["reference_genome"]:
-                selected_ref_genome = a["reference_genome"]["name"]
+                selected_ref_genome = a["reference_genome"]["name"].split(".", 1)[1]
```
jkotrs, line 615: @Niranjanpandeshwar is it enforced that the `a["reference_genome"]["name"]` will always have the subversion suffix `.xxx`? If not this would fail on non-existent index `1`.

Niranjanpandeshwar: Yes, either `reference_genome` will be empty, if not, then the name will have either GRCh38.92 or GRCh37.87. These are the only possibilities right now.

kpjensen (approved):

PR #1174, opened 2023-09-02
SCIPROD-1585: [CLIBCO] error for non-and logic on primary entity and field

csuloway-dnanexus:
Raise exception and exit with error when a cohort filter has primary entity and field logic that is not "and".

### DIFF
--- a/src/python/dxpy/cli/dataset_utilities.py
+++ b/src/python/dxpy/cli/dataset_utilities.py
@@ -1212,7 +1212,10 @@ def create_cohort(args):
     entity = rec_descriptor.model["global_primary_key"]["entity"]
     field = rec_descriptor.model["global_primary_key"]["field"]
     filters = resp.get("filters", {})
-    payload = cohort_final_payload(samples, entity, field, filters, from_project)
+    try:
+        payload = cohort_final_payload(samples, entity, field, filters, from_project)
+    except Exception as e:
+        err_exit("{}: {}".format(entity_result["id"], e))
     sql = cohort_query_api_call(resp, payload)
 
     ### temporary
--- a/src/python/dxpy/dx_extract_utils/cohort_filter_payload.py
+++ b/src/python/dxpy/dx_extract_utils/cohort_filter_payload.py
@@ -56,7 +56,7 @@ def generate_pheno_filter(values, entity, field, filters):
             if other_entity_field.split("$")[0] != entity:
                 continue
             if "logic" in compound_filter and compound_filter["logic"] != "and":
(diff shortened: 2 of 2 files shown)


src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1212,7 +1212,10 @@ def create_cohort(args):
     entity = rec_descriptor.model["global_primary_key"]["entity"]
     field = rec_descriptor.model["global_primary_key"]["field"]
     filters = resp.get("filters", {})
-    payload = cohort_final_payload(samples, entity, field, filters, from_project)
+    try:
+        payload = cohort_final_payload(samples, entity, field, filters, from_project)
+    except Exception as e:
+        err_exit("{}: {}".format(entity_result["id"], e))
```
piader1, line 1218: Does `format()` transform type `Exception` to `str` by default?

csuloway-dnanexus: Good catch. `str.format` should use the string value of the `Exception`
> ...Returns a copy of the string where each replacement field is replaced with the string value of the corresponding argument.
https://docs.python.org/3/library/stdtypes.html?highlight=str%20format#str.format
(also tested for Python 2 & 3)

src/python/dxpy/dx_extract_utils/cohort_filter_payload.py
```diff
@@ -20,55 +20,57 @@ def generate_pheno_filter(values, entity, field, filters):
         # entity field values cannot be selected by adding a filter to the
         # existing compound pheno_filter. Move the existing pheno_filter into a
         # new pheno_filter level compound filter.
-        print("WARNING: pheno_filter compound filter logic is not \"and\", creating a nested compound filter."
-              " The Cohort Browser may not work as expected with this cohort.")
+        raise ValueError("Invalid input cohort. Cohorts must have “and” logic on the primary entity and field.")
         filters["pheno_filters"] = {"compound": [filters["pheno_filters"]], "logic": "and"}
```
Niranjanpandeshwar, line 24:
```suggestion
```

src/python/dxpy/dx_extract_utils/cohort_filter_payload.py
```diff
@@ -20,55 +20,57 @@ def generate_pheno_filter(values, entity, field, filters):
         # entity field values cannot be selected by adding a filter to the
         # existing compound pheno_filter. Move the existing pheno_filter into a
```
Niranjanpandeshwar, line 21: I guess this comment needs to be updated to reflect new changes.

PR #1180, opened 2023-09-06
SCIPROD-1682: [CLIBCO] Update cohort query route

csuloway-dnanexus:
Replaced `cohort-query` API call with `raw-cohort-query`. Updated functions, variable names, and tests.


Niranjanpandeshwar (approved):
LGTM!

PR #1182, opened 2023-09-06
APPS-2138 parse container image ref

Gvaihir:
For `dx build nextflow` functionality enabling caching of docker images on the platform at compile time.
This PR adds classes to parse docker image reference in the forms of dx file path/ID or docker image/tag/digest and combinations thereof

### DIFF
--- a/src/python/dxpy/nextflow/ImageRefParser.py
+++ b/src/python/dxpy/nextflow/ImageRefParser.py
@@ -73,7 +73,7 @@ class DxPathParser(ImageRefParser):
         self._parse()
 
     def _parse(self):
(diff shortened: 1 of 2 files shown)


r-i-v-a (approved):

YuxinShi0423:
@Gvaihir is there a Jenkins test on the new test suite?

src/python/dxpy/nextflow/ImageRefParser.py
YuxinShi0423, line 1: would that be better to name it as `nextflow_image_ref_parser.py`?

PR #1183, opened 2023-09-07
SCIPROD-1643: [CLIEXPRESS] Path resolution

gabriela-reznik:


src/python/dxpy/bindings/apollo/input_arguments_validation_schemas.py
```diff
@@ -106,7 +108,7 @@
         },
         "condition": "with_at_least_one_required",
         "error_message": {
-            "message": '“--expression-matrix" cannot be passed with any argument other than "--retrieve-expression”'
+            "message": '“--expression-matrix" cannot be passed with any argument other than "--retrieve-expression”.'
```
ashkan0x, line 111: Let's use normal quotation marks as `“` might not get correctly outputted by all terminals.

```suggestion
            "message": '"--expression-matrix" cannot be passed with any argument other than "--retrieve-expression".'
```

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -47,7 +47,7 @@
 from ..dx_extract_utils.input_validation_somatic import validate_somatic_filter
 from ..dx_extract_utils.somatic_filter_payload import somatic_final_payload
 
-from ..dx_extract_utils.InputsValidator import InputsValidator
+from ..dx_extract_utils.InputsValidator import InputsValidator, PathValidator
 from ..bindings.apollo import input_arguments_validation_schemas
```
ashkan0x, line 51:
```suggestion
from ..bindings.apollo import input_arguments_validation_schemas.EXTRACT_ASSAY_EXPRESSION_INPUT_ARGS_SCHEMA
```

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1058,17 +1058,20 @@ def extract_assay_expression(parser_obj):
 
     # Validating input combinations
     parser_dict = vars(parser_obj)
-
     input_validator = InputsValidator(parser_dict=parser_dict, schema=input_arguments_validation_schemas.EXTRACT_ASSAY_EXPRESSION_INPUT_ARGS_SCHEMA, error_handler=err_exit)
```
ashkan0x, line 1061:
```suggestion
    input_validator = InputsValidator(parser_dict=parser_dict, schema=EXTRACT_ASSAY_EXPRESSION_INPUT_ARGS_SCHEMA, error_handler=err_exit)
```

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -174,100 +178,92 @@ def interpret_mutually_exclusive_group(self, check):
...
+                {"project": self.project, "cohortBrowser": False},
+            )
+        except (InvalidInput, InvalidState):
+            self.throw_error(
+                "Invalid cohort or dataset: {}".format(self.entity_result["id"]),
+            )
+        except Exception as details:
+            self.throw_error(str(details))
+
+    def assure_object_found(self):
```
ashkan0x, line 216: Maybe `is_object_in_current_project`?

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -174,100 +178,92 @@ def interpret_mutually_exclusive_group(self, check):
...
+        # for object in a different project:
+        if self.project != self.entity_result["describe"]["project"]:
+            self.throw_error(
+                'Unable to resolve "{}" to a data object or folder name in {}. Please make sure your object is in your selected project.'.format(
+                    self.parser_dict.get("path"), self.project
+                )
+            )
+
+    def assure_cohort_or_dataset(self):
```
ashkan0x, line 225: Perhaps, it can be simply called `is_record`?

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -174,100 +178,92 @@ def interpret_mutually_exclusive_group(self, check):
...
+        self.try_populate_record_http_request_info()
+        if not (
+            ("Dataset" in self.record_http_request_info["recordTypes"])
+            or ("CohortBrowser" in self.record_http_request_info["recordTypes"])
+        ):
+            self.throw_error(
+                "Invalid path. The path must point to a record type of cohort or dataset and not a {} object."
+            ).format(self.record_http_request_info["recordTypes"])
+
+    def assure_dataset_version(self):
```
ashkan0x, line 249: `validate_dataset_version`? Or `assert_dataset_version`?

Additionally, I'd suggest passing the expected dataset version as a parameter.

```suggestion
    def assure_dataset_version(self, expected_dataset_version="3.0"):
```

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -174,100 +178,92 @@ def interpret_mutually_exclusive_group(self, check):
...
+        # checking cohort/dataset version
+        dataset_version = float(self.record_http_request_info["datasetVersion"])
+        if dataset_version < 3.0:
```
ashkan0x, line 252: I'm not sure but this might need to be explicitly `if dataset_version != 3.0`

gabriela-reznik: I also saw it in preview code, as you mentioned but as I was going through cliexpress spec errors and messaging I found "The record id or path is a cohort or dataset but the version is less than 3.0." My error message should in fact match it. Thanks for catching this.

src/python/dxpy/dx_extract_utils/InputsValidator.py
gabriela-reznik: yes but it also checks for cohort or dataset. Not sure if record can be anything else

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1058,17 +1058,20 @@ def extract_assay_expression(parser_obj):
...
+            )
 
-    # path_validator = PathValidator(args.path)
-    # http_request_info = path_validator.get_http_request_info()
-    # input_validator.validate_cohort_list_assay(http_request_info)
+    path_validator = PathValidator(parser_dict=parser_dict, project=project, entity_result=entity_result, error_handler=err_exit)
+    path_validator.is_object_in_current_project()
+    path_validator.is_cohort_or_dataset()
+    path_validator.assert_dataset_version(3.0)
+    path_validator.cohort_list_assays_invalid_combination()
```
jkotrs, line 1074: @gabriela-reznik I like the granularity! Since this is the top-level use, I would recommend wrapping this sequence of 4 methods in yet another method which would be primarily exposed: `path_validator.validate()`

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -81,22 +85,26 @@ def interpret_conditions(self):
                 method_to_call = self.conditions_funcs.get(condition)
                 getattr(self, method_to_call)(key)
 
-    def throw_message(self, check):
+    def throw_schema_message(self, check):
         type = self.schema.get(check).get("error_message").get("type")
```
jkotrs, line 89: You might want to store the message here in a separate variable (like you have below in `throw_exit_error`).

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -171,103 +179,95 @@ def interpret_mutually_exclusive_group(self, check):
...
+            self.throw_error(
+                "Invalid path. The path must point to a record type of cohort or dataset and not a {} object."
+            ).format(self.record_http_request_info["recordTypes"])
+
+    def assert_dataset_version(self, expected_min_dataset_version=3.0):
+        # checking cohort/dataset version
+        dataset_version = float(self.record_http_request_info["datasetVersion"])
+        if dataset_version < expected_min_dataset_version:
+            self.throw_error(
+                "{}: Version of the cohort or dataset is too old. Version must be 3.0.".format(
```
jkotrs, line 258: Minor: but to make sure the message corresponds to the reality, I would suggest replacing the "3.0" string with the `expected_min_dataset_version` variable.

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -171,103 +179,95 @@ def interpret_mutually_exclusive_group(self, check):
...
+                'Unable to resolve "{}" to a data object or folder name in {}. Please make sure your object is in your selected project.'.format(
+                    self.parser_dict.get("path"), self.project
+                )
+            )
+
+    def is_cohort_or_dataset(self):
+        # resolving non record/cohort type
+        if self.entity_result is None:
+            self.throw_error(
+                "The path must point to a record type of cohort or dataset, not a folder."
```
jkotrs, line 233: Does the empty `self.entity_result` always mean the path is a folder? (that is what the message implies)

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -171,103 +179,95 @@ def interpret_mutually_exclusive_group(self, check):
...
-#         return self.http_request
+class PathValidator:
+    """
+    PathValidator class checks for invalid object inputs and its combination with passed arguments.
+    """
+
+    def __init__(self, parser_dict, project, entity_result, error_handler=print):
+        self.parser_dict = parser_dict
+        self.project = project
+        self.entity_result = entity_result
```
jkotrs, line 198: @gabriela-reznik I would highly recommend to work with smaller, clear-scoped objects rather than an API call result directly. How about passing in (to the constructor) the describe object and call it `entity_describe`? That might set the expectations more clearly.

From the usage of the current result object, it's either `result["id"]` or `result["describe"]...`. I believe the `id` should be part of the "describe" too.

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -171,103 +179,95 @@ def interpret_mutually_exclusive_group(self, check):
...
-
-#     def get_http_request_info(self):
-#         return self.http_request
+class PathValidator:
+    """
+    PathValidator class checks for invalid object inputs and its combination with passed arguments.
+    """
+
+    def __init__(self, parser_dict, project, entity_result, error_handler=print):
+        self.parser_dict = parser_dict
```
jkotrs, line 196: Maybe `input_dict` would be more descriptive?

jkotrs (commented):
Great job Gabi! I left a few comments.

src/python/dxpy/dx_extract_utils/InputsValidator.py
gabriela-reznik: you mean just storing or passing it as parameter? I considered the second option but then every method would have some repetitive coding extracting that message

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -81,22 +85,23 @@ def interpret_conditions(self):
                 method_to_call = self.conditions_funcs.get(condition)
                 getattr(self, method_to_call)(key)
 
-    def throw_message(self, check):
+    def throw_schema_message(self, check):
         type = self.schema.get(check).get("error_message").get("type")
```
ashkan0x, line 89: I highly recommend renaming this variable, because `type` is a reserved keyword in Python and should not be overwritten.

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -81,22 +85,23 @@ def interpret_conditions(self):
...
+        message = self.schema.get(check).get("error_message").get("message")
         if type == "warning":
-            self.throw_warning(check)
+            self.throw_warning(message)
         elif (type == None) or (type == "error"):
```
ashkan0x, line 93: A bit safer perhaps using `isinstance`:

```suggestion
        elif (isinstance(type, type(None)) or (type == "error"):
```

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -171,103 +176,94 @@ def interpret_mutually_exclusive_group(self, check):
...
+        if self.project != self.entity_describe["project"]:
+            self.throw_error(
+                'Unable to resolve "{}" to a data object or folder name in {}. Please make sure your object is in your selected project.'.format(
+                    self.input_dict.get("path"), self.project
+                )
+            )
+
+    def is_cohort_or_dataset(self):
+        # resolving non record/cohort type
+        if self.entity_describe["class"] != "record":
```
ashkan0x, line 228: The name of the function implies that we want to check if the `type` of the record is `dataset` or `cohort`, but only the `class` is checked. Is this intentional? I believe we may instead want to check if either `'Cohort'` or `'Dataset'` is in `entity_describe["types"]`.

src/python/dxpy/dx_extract_utils/InputsValidator.py
```diff
@@ -171,103 +176,94 @@ def interpret_mutually_exclusive_group(self, check):
...
+            self.throw_error(
+                "Invalid path. The path must point to a record type of cohort or dataset and not a {} object.".format(
+                    self.entity_describe["class"]
+                )
+            )
```
ashkan0x, line 233: Would it be worth adding one explicit additional check for `Dataset` and `Cohort` types here? I'm not sure if it's needed actually. cc @jkotrs 

e.g.

```
if 'Dataset' not in self.entity_describe["types"] or 'Cohort' not in self.entity_describe["types"]:
    ...
```

ashkan0x (approved):

jkotrs (approved):

PR #1185, opened 2023-09-07
SCIPROD-1702: [CLIBCO] Primary entity/field filtering

csuloway-dnanexus:
Implemented logic for adding the filter on the primary entity and field:
- "in" condition intersects requested cohort
- "not-in" condition is subtracted from requested cohort
- "exists" condition is ignored and removed
- "not-exists" condition raises an error


src/python/dxpy/dx_extract_utils/cohort_filter_payload.py
```diff
@@ -22,18 +22,30 @@ def generate_pheno_filter(values, entity, field, filters):
...
         if "logic" in compound_filter and compound_filter["logic"] != "and":
             raise ValueError("Invalid input cohort. Cohorts must have “and” logic on the primary entity and field.")
-        # The entity field filter is valid for addition of the "in" values condition
-        compound_filter["filters"][entity_field].append(entity_field_filter)
+        primary_filters = []
+        for primary_filter in compound_filter["filters"][entity_field]:
+            if "condition" not in primary_filter or primary_filter["condition"] == "exists":
+                pass
+            elif primary_filter["condition"] == "in":
+                values = [value for value in values if value in primary_filter["values"]]
```
Niranjanpandeshwar, line 37: Will using set.intersect be faster here? Or any other method, since I can imagine the number of IDs can be very large sometimes. Similarly below is it better to use set(A)-set(B)?

csuloway-dnanexus: I was trying to preserve order here so I didn't use set intersection/subtract here. Since there can be a large number of IDs I will update this.

Niranjanpandeshwar (approved):
LGTM.

PR #1186, opened 2023-09-07
Sciprod 1664 test input validation

JoeMulka:
Hi Folks,
Here's all the invalid JSON I could come up with, plus a handful of valid ones.  It tests a few things that aren't currently rejected by the validator, which is why not all the tests are currently passing.  The names of the tests and their test JSON match, and I tried to make the naming of the JSON explain what was being tested.  Please let me know if there is anything that you would like tested that I haven't already made a test for 😄
Note: I should have named this branch after SCIPROD-1665 which is the testing ticket


gabriela-reznik (approved):

src/python/test/test_extract_expression.py
```diff
@@ -0,0 +1,558 @@
...
+from dxpy.bindings.apollo.assay_filtering_json_schemas import (
+    EXTRACT_ASSAY_EXPRESSION_JSON_SCHEMA,
+)
+
+dirname = os.path.dirname(__file__)
+
+python_version = sys.version_info.major
+
+
+class TestDXExtractExpression(unittest.TestCase):
```
jkotrs, line 41: Hi @JoeMulka, I would recommend creating more generic test helper methods so that individual cases won't need to get duplicated code for loading the file and asserting the result.

src/python/dxpy/bindings/apollo/assay_filtering_json_schemas.py
```diff
@@ -9,7 +9,7 @@
     },
     "expression": {
         "type": dict,
-        "properties": {"min_value": {"type": str}, "max_value": {"type": str}},
+        "properties": {"min_value": {"type": str,"required":False}, "max_value": {"type": str,"required":False  }},
```
ashkan0x, line 12: Was the class not working as expected without `"required": False`? If so, this is probably a bug and I can look into it and fix it.

ashkan0x, line 12: Were you seeing any strange behavior when this was not part of the schema? If so, I think there might be a bug somewhere in the class code which I should fix.

JoeMulka: I think I just assumed that those were meant to be required, but I'm realizing that they probably are not meant to be required.  I'll undo this change

src/python/test/expression_test_files/input_dict.py
```diff
@@ -0,0 +1,465 @@
...
+            "expression": {"min_value": "10.2", "max_value": "10000"},
+        },
+        "bad_toplevel_key": {"not_real_key": "1", "sample_id": ["sample1"]},
+        "location_missing_end": {
+            "location": [{"chromosome": "1", "starting_position": "1"}]
+        },
+        "annotation_name_maxitem": {
+            "annotation": {
+                "feature_name": [
+                    "item",
```
jkotrs, line 34:
```
["item"] * 100
```

jkotrs (commented):
The refactoring helped with legibility, thanks Joe! One comment before we can wrap it up - could you please replace the literal "item" arrays with generated (see comment)? Thanks!

PR #1193, opened 2023-09-11
Sciprod 1647 test dataset validation

JoeMulka:
Hi folks,
The path validation function has been implemented on PMUX-1693-cliexpress, so I went ahead and checked out the relevant files from that branch into this one so that the imports would work and tests would run.  From reading the function, it looks like there are 5 possible failure modes, but the ticket only calls out two of them.  Should I create tests for the other possible failure modes, or do we expect other tickets to cover those?
Thanks,
Joe


src/python/dxpy/bindings/apollo/path_validator.py
```diff
@@ -57,12 +57,15 @@ def is_cohort_or_dataset(self):
...
+            ("Dataset" in self.record_http_request_info.get("recordTypes",[]))
+            or ("CohortBrowser" in self.record_http_request_info.get("recordTypes",[]))
+            or ("Dataset" in self.record_http_request_info.get("types",[]))
+            or ("CohortBrowser" in self.record_http_request_info.get("types",[]))
         ):
+            # Check recordTypes first, then types, and if neither have data, put TYPENOTFOUND in the type part of the string
             self.throw_error(
-                "Invalid path. The path must point to a record type of cohort or dataset and not a {} object."
-            ).format(self.record_http_request_info["recordTypes"])
+                "Invalid path. The path must point to a record type of cohort or dataset and not a {} object.".format(self.record_http_request_info.get("recordTypes",self.record_http_request_info.get("types","TYPENOTFOUND")))
```
ashkan0x, line 67: Very minor, but it could be also:

```suggestion
                "Invalid path. The path must point to a record type of cohort or dataset and not a {} object.".format(self.record_http_request_info.get("recordTypes") or self.record_http_request_info.get("types") or "TYPENOTFOUND")
```

src/python/dxpy/bindings/apollo/path_validator.py
```diff
@@ -57,12 +57,15 @@ def is_cohort_or_dataset(self):
         # since object is record:
         self.try_populate_record_http_request_info()
         if not (
-            ("Dataset" in self.record_http_request_info["recordTypes"])
-            or ("CohortBrowser" in self.record_http_request_info["recordTypes"])
+            ("Dataset" in self.record_http_request_info.get("recordTypes",[]))
```
ashkan0x, line 60: Minor, could be:

```
self.try_populate_record_http_request_info()

EXPECTED_TYPES = ["Dataset", "CohortBrowser"]

if any(x not in self.record_http_request_info["recordTypes"] and x not in self.record_http_request_info["types"] for x in EXPECTED_TYPES):
    self.throw_error(.......)
```

src/python/test/test_extract_expression.py
```diff
@@ -78,6 +102,30 @@ def standard_positive_filter_test(self, json_name):
...
+    #
+
+    # 2. Object not of class record
+    def test_bad_dataset_type_unit(self):
+        test_record = self.wrong_type_path_file
+        expected_error_message = "{}: Invalid path. The path must point to a record type of cohort or dataset".format(test_record)
+        self.standard_path_validation_test(test_record,expected_error_message)
+
+
+    # 3. Object not of recordType "Dataset" or "CohortBrowser"
```
ashkan0x, line 122: This comment should point to `4. Object is not of correct version (3.0 at the time of this writing)` instead, right?

ashkan0x (approved):
Left two minor comments, but all in all looks good to me.

Just in case, I'd recommend triggering the Jenkins `Python 2` workflow on this branch before merging.

src/python/dxpy/bindings/apollo/path_validator.py
jkotrs: Hi @JoeMulka I would agree with Ashkan this can get de-repeated slightly. I'd recommend establishing a method to retrieve types first and then use it with benefit to test certain types presence. Something along these lines perhaps?
```
def get_type(self):
    record_info = self.record_http_request_info # <-- we need to rename the field to the shorter version
    return set(record_info.get("record", []) + record_info.get("recordType", []));
```
And then:
```
if self.get_type() in ["Cohort", "Dataset"]:
   ...
```

src/python/test/test_extract_expression.py
```diff
@@ -49,18 +53,38 @@ def setUpClass(cls):
...
+    
+    @classmethod
+    def tearDownClass(cls):
+        shutil.rmtree(cls.general_output_dir)
+    
+    #
+    # Helper functions used by different types of tests
+    #
+
+    def standard_path_validation_test(self,test_record,expected_error_message):
```
jkotrs, line 80: Minor: `standard` does not really add any information. Maybe `common` to signify this is shared to all cases?

jkotrs (commented):
Left a few comments, thanks!

src/python/test/test_extract_expression.py
JoeMulka: yep, updated

PR #1197, opened 2023-09-13
SCIPROD-1658 - `VizPayloadBuilder` class

ashkan0x:


src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1076,6 +1081,33 @@ def extract_assay_expression(parser_obj):
     path_validator = PathValidator(input_dict=parser_dict, project=project, entity_describe=entity_describe, error_handler=err_exit)
     path_validator.validate(check_list_assays_invalid_combination=True)
 
+    # # Validating input JSON
+    # if parser_obj.input_json:
```
piader1, line 1085: This is not a parser obj (name could indicate ArgumentParser) but a Namespace object which is being refferred to as `args` across dx-toolkit

src/python/dxpy/bindings/apollo/vizserver_payload_builder.py
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+        """
+        raw_filters = {
+            "raw_filters": {
+                "assay_filters": {"name": assay_name, "id": assay_id,}
+            }
+        }
+
+        raw_filters["raw_filters"]["assay_filters"].update(filters)
```
piader1, line 118: Should not there be a distinction among `pheno_filters` and `assay_filters`? or is this dedicated to assay queries?

src/python/dxpy/bindings/apollo/vizserver_payload_builder.py
```diff
@@ -0,0 +1,135 @@
...
+        assay_id="a-b1-2c-f-xyz-test",
+        filters={
+            "logic": "or",
+            "compound": [
+                {
+                    "test": "1",
+                },
+                {"exmaple": 3},
+            ],
+        },
```
jkotrs, line 32: For future enh.: I would consider more atomic API to gradually build up the filters. Something like:
```
payload = (
   builder.with_assay("xyz", "a-b1-2c-...")
   .with_any_filter([{"test": "1"}, {"example": 3}])
   ...
   .build()
)
```

jkotrs (approved):

src/python/dxpy/cli/dataset_utilities.py
ashkan0x: Great catch. Thank you! I'll take a note of this and will modify it later. There are many open PRs at the moment, so I don't want to disrupt other devs' work.

src/python/dxpy/bindings/apollo/vizserver_payload_builder.py
ashkan0x: Correct, it's dedicated to `assay_filters`. I'm assuming `raw_filters` are already prepared at this point using the [`InputJSONFiltersValidator` class](https://github.com/dnanexus/dx-toolkit/blob/4e2143b602b2b2f928453081a1ee809fd8754b55/src/python/dxpy/bindings/apollo/ParseFiltersJSONbySchema.py#L4), so only `assay_id` and `assay_name` need to be added to the `dict`.

AFAIK `pheno_filters` doesn't need a special structure, so it can be basically just passed to `filters` within this class. Does that make sense?

PR #1198, opened 2023-09-13
SCIPROD-1648: [CLIEXPRESS] Dataset class

gabriela-reznik:


src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,56 @@
+from dxpy import DXHTTPRequest
+
+
+class Dataset:
+    def __init__(self, record_id, project):
```
jkotrs, line 5: Minor: maybe `project_id` to align with `record_id`?

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,56 @@
...
+        self.record_id = record_id
+        self.project = project
+
+        self.populate_http_request()
+
+    def populate_http_request(self):
```
jkotrs, line 11: The fact that info is drawn from a request object is more of an internal implementation fact. I would call it by what it does: `call_visualize` or `get_visualize_info` or similar. The variable then `vizualise_info`?

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,56 @@
+from dxpy import DXHTTPRequest
+
+
+class Dataset:
+    def __init__(self, record_id, project):
+        self.record_id = record_id
+        self.project = project
+
+        self.populate_http_request()
```
jkotrs, line 9: It is not a good practice to trigger external resources or any heavier computation in the constructor. One of the reasons is to keep our ability to instantiate a new object without necessarily depending on the underlying environment.

Instead I would recommend lazy-loading - check for the request response object only when someone requests info from it:
```
def get_vizualise_info(self):
   if self.vizualise_info is None:
      self.vizualise_info = DXHTTPRequest(
            "/" + self.record_id + "/visualize",
            {"project": self.project, "cohortBrowser": False},
        )
  return self.vizualise_info
```
And then:
```
def get_record_name(self):
   return self.get_vizualise_info().get("recordName")
```

src/python/dxpy/bindings/apollo/dataset.py
gabriela-reznik: sounds good :)

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,48 @@
+from dxpy import DXHTTPRequest
+
+
+class Dataset:
```
gabriela-reznik, line 4: I'm now considering that calling it Dataset class might be confusing as the parameters passed here regard the record (because without visualize internal function all we have is the record-id). Do you guys think it's worth renaming?
Just small recap: 
- visualize gets info from record (regarding record, database and more)
- descriptor gets info from dataset regarding dataset

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1090,7 +1090,35 @@ def extract_assay_expression(args):
...
+    
+    assay_names_list = dataset_handler.get_assay_names("molecular_expression")
+
+    if parser_dict.get("list_assays"):
+        print(assay_names_list)
+        sys.exit(0)
+
+    # possible assay picking
+    if parser_dict.get("assay_name"):
+        assay_name = parser_dict.get("assay_name")
```
ashkan0x, line 1102: Better just to use the argparse attributes themselves, `args.assay_name` instead of `assay_name = parser_dict.get("assay_name")`

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1090,7 +1090,35 @@ def extract_assay_expression(args):
...
+        # print(assay_name)
+        assay_index = dataset_handler.get_assay_indice_in_list(assay_names_list, assay_name)
```
ashkan0x, line 1104: Suggestion for 1101-1106 replacement:

```suggestion
        assay_index = dataset_handler.get_assay_indice_in_list(assay_names_list, args.assay_name) if args.assay_name else 0
```

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1090,7 +1090,35 @@ def extract_assay_expression(args):
     # assay names listing
     descriptor = DXDataset(dataset_handler.dataset, project=dataset_handler.datasetRecordProject).get_descriptor()
     dataset_handler.populate_descriptor(descriptor)
-    print(dataset_handler.list_assay_names("molecular_expression"))
+    
+    assay_names_list = dataset_handler.get_assay_names("molecular_expression")
+
+    if parser_dict.get("list_assays"):
```
ashkan0x, line 1096:
```suggestion
    if args.list_assays:
```

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -1090,7 +1090,35 @@ def extract_assay_expression(args):
...
+        print(assay_names_list)
```
ashkan0x, line 1097: There may be more than one assays, which would be nicer to print on separate lines. 

```suggestion
        print(*assay_names_list, sep="\n")
```

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,120 @@
...
+                {"project": self.project_id, "cohortBrowser": False},
+            )
+        return self.visualize_info
+
+    @property
+    def vizserver_url(self):
+        vis_info = self.get_visualize_info()
+        return vis_info.get("url")
+
+    def get_dx_record(self):
```
ashkan0x, line 28: If you use the initialize the superclass in `__init__` then you can get rid of this

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,120 @@
...
+    @property
+    def dx_record_obj(self):
+        return self.get_dx_record()
+
+    @property
+    def record_details(self):
+        return self.dx_record_obj.get_details()
+
+    @property
+    def record_descriptor(self):
```
ashkan0x, line 42: This could be misleading, because `descriptor` is the `file` that is linked to the dataset. I'd suggest simply calling it `describe` but then again if the superclass is already initialized, then it already contains `describe` info.

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,120 @@
+from dxpy import DXHTTPRequest
+from dxpy.bindings import DXRecord
+
+
+class Dataset(DXRecord):
+    def __init__(self, record_id, project_id):
+        self.record_id = record_id
```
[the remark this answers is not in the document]
ashkan0x: But maybe @jkotrs has a different viewpoint on this. I just think there's no need to lazily grab `describe` or `details`.

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,120 @@
...
+    def record_details(self):
+        return self.dx_record_obj.get_details()
+
+    @property
+    def record_descriptor(self):
+        return self.dx_record_obj.describe()
+
+    @property
+    def is_cohort(self):
+        if "CohortBrowser" in self.record_descriptor.get("types"):
```
ashkan0x, line 47: Minor:

```suggestion
        return True if "CohortBrowser" in self.record_descriptor.get("types") else False
```

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @property
+    def base_sql(self):
+        return self.record_details.get("baseSql")
+
+    @property
+    def filters(self):
+        return self.record_details.get("filters")
+
+    def populate_dx_dataset_descriptor(self, descriptor):
```
ashkan0x, line 75: Better just call it `get_descriptor`, and then this method would need to get the descriptor file ID from `details["descriptor"]` and read it.

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,120 @@
...
+    def get_assay_list(self, assay_type):
+        if self.assays_list is None:
+            selected_type_assays = []
+            for a in self.dx_dataset_descriptor.get("assays"):
+                if a["generalized_assay_model"] == assay_type:
+                    selected_type_assays.append(a)
+            self.assays_list = selected_type_assays
+        return self.assays_list
+
+    def get_assay_names(self, assay_type):
```
ashkan0x, line 87: Might be worth storing all assays in `self.assay_names_list` and instead make a dictionary based on their `type`, `name`, and `uuid`.

Example:

```
{
  "molecular_expression": [
    {
      "name": "assay1",
      "uuid": "abc-123"
    },
    {
      "name": "assay2",
      "uuid": "xyz-567"
    },
    {
      "name": "assay3",
      "uuid": "zzz-123"
    }
  ],
  "somatic": [
    {
      "name": "s1",
      "uuid": "yyy-123"
    },
    {
      "name": "test123",
      "uuid": "aaa-125"
    }
  ]
}
```

src/python/dxpy/bindings/apollo/dataset.py
jkotrs: Why not simply:
```
return "CohortBrowser" in self.record_descriptor.get("types")
```

src/python/dxpy/bindings/apollo/dataset.py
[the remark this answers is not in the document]
jkotrs: Please avoid calling external sources in the constructor - that inhibits the testing and also makes it tricky to manipulate with the class. The lazy load should be the default for external sources.

src/python/dxpy/bindings/apollo/dataset.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assays_list = None
+        self.assay_names_list = None
+
+    def get_visualize_info(self):
+        if self.visualize_info is None:
+            self.visualize_info = DXHTTPRequest(
+                "/" + self.record_id + "/visualize",
+                {"project": self.project_id, "cohortBrowser": False},
+            )
+        return self.visualize_info
```
jkotrs, line 21: This is good, you could also improve this for downstream use and make it a property, while storing the info in a "private" attribute `self._visualize_info`:
```
@property
def visualize_info(self):
    if self._visualize_info is None:
       ...
    return self._visualize_info
```
So that later on, you could use it like this:
```
self.visualize_info["url"]
```

src/python/dxpy/bindings/apollo/dataset.py
jkotrs: `describe` might be confused with the results of `/describe` call as the `descriptor` json contains different info

ashkan0x: @jkotrs Yes, but it's exactly what this class is doing, it's returning the result of the `/describe` call. This is why I suggested calling it describe.

ashkan0x (approved):

jkotrs (approved):
Great job!

PR #1199, opened 2023-09-13
Update to allow latest version of requests

jtratner:
Not sure why there's a pin, but it looks like it's causing some potential version conflicts on my build.


kpjensen:
This should be addressed by https://github.com/dnanexus/dx-toolkit/pull/1277. `requests` will no longer be a requirement of dxpy.

kpjensen:
Released in https://github.com/dnanexus/dx-toolkit/blob/master/CHANGELOG.md#3660---2023121

PR #1207, opened 2023-09-18
Sciprod 1644 test path options

JoeMulka:
Hi folks,
This PR adds the unit tests for the ValidateArgsBySchema validator object.  I'm in the process of updating the Errors and Messaging table to keep track of which errors are covered by which tests, and the code already has the error messages noted in comments above the relevant tests.  There are some errors where I'm not sure which ticket applies, we can discuss this soon.  Also, note that for EM-22, I couldn't come up with a way to test the error message specifically, because the situation described will always have another issue with it, and the other issues are checked first.  Please let me know if there's anything I can improve or make clearer.
Thanks,
Joe


src/python/test/test_extract_expression.py
```diff
@@ -78,8 +116,400 @@ def standard_positive_filter_test(self, json_name):
...
+        actual_err_msg = process.communicate()[1]
+        # print(actual_err_msg)
+
+        self.assertTrue(expected_error_message in actual_err_msg)
+
+    # EM-15
+    # When a --retrieve-expression flag is passed without any value or when an empty JSON (an empty file or just {}) is passed with --retrieve-expression flag
+    # Note: empty JSON is tested in the JSON validation section
+    def test_no_value_retrieve_exp(self):
+        expected_error_message = "No filter json is passed with --retrieve-expression or JSON or --retrieve-expression does not contain valid filter information."
```
ashkan0x, line 387:
```suggestion
        expected_error_message = "No filter json is passed with --retrieve-expression or JSON for --retrieve-expression does not contain valid filter information."
```

src/python/test/test_extract_expression.py
```diff
@@ -78,8 +116,400 @@ def standard_positive_filter_test(self, json_name):
...
+        )
+        actual_err_msg = process.communicate()[1]
+        # print(actual_err_msg)
+
+        self.assertTrue(expected_error_message in actual_err_msg)
+
+    # EM-21
+    # When --json-help is passed with another option from --assay-name, --sql, --additional-fields, --expression-matix, --output
+    def test_json_help_other_option(self):
+        expected_error_message = "--json-help cannot be passed with any of --assay-name, --sql, --additional-fields, --expression-matix, or --output"
```
ashkan0x, line 444:
```suggestion
        expected_error_message = "--json-help cannot be passed with any of --assay-name, --sql, --additional-fields, --expression-matrix, or --output"
```

src/python/test/test_extract_expression.py
```diff
@@ -78,8 +116,400 @@ def standard_positive_filter_test(self, json_name):
...
+        }
+        self.standard_input_args_test(input_dict, expected_error_message)
+
+    # EM-22
+    # When --expression-matrix is passed with other arguments other than, any context other than, --retrieve-expression
+    # It seems that every combination of args that could be passed with this cause a different issue to be caught first
+    # Which is fine but the error message will be for the other error
+    def test_exp_matrix_other_args(self):
+        # expected_error_message = "--expression-matrix cannot be passed with any argument other than --retrieve-expression"
+        expected_error_message = "“--json-help cannot be passed with any of --assay-name, --sql, --additional-fields, --expression-matix, or --output”"
```
ashkan0x, line 458: It's probably fine to just pull the expected message directly from the schema too. @jkotrs any preferences there? Should we hardcode the expected error message or is it fine to grab it from the schema?

```EXTRACT_ASSAY_EXPRESSION_INPUT_ARGS_SCHEMA.get("5_json_help_exclusive_with_exceptions").get("message)```

File: src/python/test/test_extract_expression.py [hunk 4]
```diff
@@ -78,8 +116,400 @@ def standard_positive_filter_test(self, json_name):
...
+            command, stderr=subprocess.PIPE, universal_newlines=True
+        )
+        actual_err_msg = process.communicate()[1]
+        # print(actual_err_msg)
+
+        self.assertTrue(expected_error_message in actual_err_msg)
+
+    # EM-17
+    # When the .json file provided does not exist
+    # Note: this probably needs to be tested with a Popen rather than with the ValidateArgsBySchema function
```
ashkan0x, line 416: Agreed! This should be probably just an E2E test and implemented once the main flow is finished.

src/python/test/test_extract_expression.py
```diff
@@ -78,8 +116,400 @@ def standard_positive_filter_test(self, json_name):
...
+    # When a --retrieve-expression flag is passed without any value or when an empty JSON (an empty file or just {}) is passed with --retrieve-expression flag
+    # Note: empty JSON is tested in the JSON validation section
+    def test_no_value_retrieve_exp(self):
+        expected_error_message = "No filter json is passed with --retrieve-expression or JSON or --retrieve-expression does not contain valid filter information."
+        input_dict = {"path": self.test_record, "retrieve_expression": True}
+        self.standard_input_args_test(input_dict, expected_error_message)
+
+    # EM-16
+    # When the string provided is a malformed JSON
+    def test_malformed_retr_exp_json(self):
```
ashkan0x, line 393: Probably not needed because dxpy will throw an error first via ```json.decoder.JSONDecodeError```

src/python/test/test_extract_expression.py
JoeMulka: added @unittest.skip, we can reactivate this test in the future once the flow logic is in place

src/python/test/test_extract_expression.py
JoeMulka: Added a skip, we can reactivate or remove before release

src/python/test/test_extract_expression.py
JoeMulka: If possible it would be even better to grab it from the spec.  If we grab the error message from the schema, then we won't catch the case where the correct error is being caught, but the message in the schema is wrong

PR #1213, opened 2023-09-22
SCIPROD-1713: fix extract_assay somatic no assay of type error message

csuloway-dnanexus:
Fix `extract_assay somatic` no assay of type error message
- Fixed string format typo
- Use `friendly_assay_type` for both no assay of type error messages


Niranjanpandeshwar (approved):
LGTM!

PR #1221, opened 2023-09-27
Sciprod 1662 vizclient get data

JoeMulka:
Hi Folks,
This PR adds the VizClient class, which is used to get the results of sending the payload to the raw and raw-query endpoints.  I tested it with the following:
`dx extract_assay expression record-GYPg9Jj06j8pp3z41682J23p --retrieve-expression  --input-json '{"expression": {"min_value": 1}, "location": [{"chromosome": "1", "starting_position": "7", "ending_position": "250000000"}]}'  --additional-fields strand --sql`
which works with and without the --sql flag
I borrowed the error handling from the raw_api_call function, please let me know if there's any other type of error that needs to be caught
Thanks,
Joe


src/python/dxpy/bindings/apollo/vizserver_client.py
```diff
@@ -0,0 +1,46 @@
...
+
+class VizClient(object):
+    def __init__(self, url,project_id,record_id, error_handler=print) -> None:
+        self.url = url
+        self.project_id = project_id
+        self.record_id = record_id
+        self.error_handler = error_handler
+
+    def get_data(self,payload):
+        #resource_val = "{}/data/3.0/{}:{}/raw".format(self.url,self.project_id,self.record_id)
```
JoeMulka, line 11:
```suggestion
```

src/python/dxpy/bindings/apollo/vizserver_client.py
```diff
@@ -0,0 +1,45 @@
...
+    def get_data(self,payload):
+        resource_val = "{}/data/3.0/{}/raw".format(self.url,self.record_id)
+        try:    
+            raw_response = dxpy.DXHTTPRequest(
+                resource=resource_val, data=payload, prepend_srv=False
+            )
+            if "error" in raw_response.keys():
+                if raw_response["error"]["type"] == "InvalidInput":
+                    err_message = "Insufficient permissions due to the project policy.\n" + raw_response["error"]["message"]
+                elif raw_response["error"]["type"] == "QueryBuilderError" and raw_response["error"]["details"] == "rsid exists in request filters without rsid entries in rsid_lookup_table.":
```
jkotrs, line 19: Just checking - is it `["error"]["details"]` or  `["error"]["message"]`? I do not see `details` field mentioned in the API: https://github.com/dnanexus/hare/blob/master/docs/API-SpecificationV3.md#common-api-errors However, that does not mean it is not there.

src/python/dxpy/bindings/apollo/vizserver_client.py
```diff
@@ -0,0 +1,45 @@
...
+        self.record_id = record_id
+        self.error_handler = error_handler
+
+    def get_data(self,payload):
+        resource_val = "{}/data/3.0/{}/raw".format(self.url,self.record_id)
+        try:    
+            raw_response = dxpy.DXHTTPRequest(
+                resource=resource_val, data=payload, prepend_srv=False
+            )
+            if "error" in raw_response.keys():
```
jkotrs, line 16: Minor: the `.keys()` is not needed as python would do this for you

src/python/dxpy/bindings/apollo/vizserver_client.py
```diff
@@ -0,0 +1,45 @@
...
+                if raw_response["error"]["type"] == "InvalidInput":
+                    err_message = "Insufficient permissions due to the project policy.\n" + raw_response["error"]["message"]
+                elif raw_response["error"]["type"] == "QueryBuilderError" and raw_response["error"]["details"] == "rsid exists in request filters without rsid entries in rsid_lookup_table.":
+                    err_message = "At least one rsID provided in the filter is not present in the provided dataset or cohort"
+                else:
+                    err_message = raw_response["error"]
+                self.error_handler(str(err_message))
+        except Exception as details:
+            self.error_handler(str(details))
+        return(raw_response)
```
jkotrs, line 26: Why the brackets?

src/python/dxpy/bindings/apollo/vizserver_client.py
```diff
@@ -0,0 +1,45 @@
...
+                if raw_query_response["error"]["type"] == "InvalidInput":
+                    err_message = "Insufficient permissions due to the project policy.\n" + raw_query_response["error"]["message"]
+                elif raw_query_response["error"]["type"] == "QueryBuilderError" and raw_query_response["error"]["details"] == "rsid exists in request filters without rsid entries in rsid_lookup_table.":
+                    err_message = "At least one rsID provided in the filter is not present in the provided dataset or cohort"
+                else:
+                    err_message = raw_query_response["error"]
+                self.error_handler(str(err_message))
+        except Exception as details:
+            self.error_handler(str(details))
+        return(raw_query_response)
```
jkotrs, line 44: same here

src/python/dxpy/bindings/apollo/vizserver_client.py
```diff
@@ -0,0 +1,45 @@
...
+                resource=resource_val, data=payload, prepend_srv=False
+            )
+            if "error" in raw_query_response.keys():
+                if raw_query_response["error"]["type"] == "InvalidInput":
+                    err_message = "Insufficient permissions due to the project policy.\n" + raw_query_response["error"]["message"]
+                elif raw_query_response["error"]["type"] == "QueryBuilderError" and raw_query_response["error"]["details"] == "rsid exists in request filters without rsid entries in rsid_lookup_table.":
+                    err_message = "At least one rsID provided in the filter is not present in the provided dataset or cohort"
+                else:
+                    err_message = raw_query_response["error"]
+                self.error_handler(str(err_message))
```
jkotrs, line 41: Are the error responses really equivalent for both endpoints? If they indeed are, can we please abstract into common `error_from_response(response) -> str` (or similar)?

src/python/dxpy/bindings/apollo/vizserver_client.py
jkotrs, line 1: Please name the file the same way as the class name but lowercase (`vizclient.py`), thanks!

jkotrs (commented):
A few comments (mainly the error messages commonality to be solved), otherwise looks good!

src/python/dxpy/bindings/apollo/vizserver_client.py
ashkan0x, line 19: I'm not fully familiar with vizserver, but I think this is not needed, because there are no `rsid`s anywhere in the expression tables. 

```suggestion
                elif raw_response["error"]["type"] == "QueryBuilderError":
```

src/python/dxpy/bindings/apollo/vizserver_client.py
```diff
@@ -0,0 +1,45 @@
...
+    def get_raw_sql(self,payload):
+        resource_val = "{}/viz-query/3.0/{}/raw-query".format(self.url,self.record_id)
+        try:    
+            raw_query_response = dxpy.DXHTTPRequest(
+                resource=resource_val, data=payload, prepend_srv=False
+            )
+            if "error" in raw_query_response.keys():
+                if raw_query_response["error"]["type"] == "InvalidInput":
+                    err_message = "Insufficient permissions due to the project policy.\n" + raw_query_response["error"]["message"]
+                elif raw_query_response["error"]["type"] == "QueryBuilderError" and raw_query_response["error"]["details"] == "rsid exists in request filters without rsid entries in rsid_lookup_table.":
```
ashkan0x, line 37: Same comment as above here. No need for `rsid` checks I believe.

src/python/dxpy/bindings/apollo/vizserver_client.py
ashkan0x: Minor suggestion:

It would be also nice to have a general method such as `get_response` which would get an `endpoint` argument with two possible values `"sql"` and `"data"`, and then within `dataset_utilities` we would just simply call:

```
VizClient(url,project,record_id).get_response("sql" if args.sql else "data")
```

jkotrs (approved):

PR #1226, opened 2023-09-29
SCIPROD-1570: order extract_assay germline results

csuloway-dnanexus:
`extract_assay germline` results are ordered by chromosome and starting position. Chromosome order is first numeric and then string.

### DIFF
--- a/src/python/dxpy/cli/dataset_utilities.py
+++ b/src/python/dxpy/cli/dataset_utilities.py
@@ -27,6 +27,7 @@ import re
 import csv
 import dxpy
 import codecs
+import math
 import subprocess
 from ..utils.printing import fill
 from ..bindings import DXRecord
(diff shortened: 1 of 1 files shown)


Niranjanpandeshwar (approved):
LGTM!

src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -796,11 +797,18 @@ def extract_assay_germline(args):
                     if r["genotype_type"] == "hom":
                         r["genotype_type"] = "hom-alt"
 
+            def sort_variant(d):
```
durae, line 800: data/raw API supports order_by input if you want to get it sorted from a database/vizserver: https://github.com/dnanexus/hare/blob/master/docs/API-SpecificationV3.md#data30record-xxxxraw

csuloway-dnanexus: Good point, Niranjan mentioned this as well. This might be a special case because we want to sort chromosomes by numeric value if the string represents a number, otherwise by string value. Is this behavior supported by the `order_by` input to the data/raw API?

durae: The supported order is only based on the native sort for that column type. So if it's stored as a string, then it will be always alphabetical ASC or DESC. However, "5" < "6" the same way as 5 < 6, unless there are zeros. You might have special case, so just check the results from API call when you specify order_by and if it's not what you want, then use your sort. Just db sort would be faster.

csuloway-dnanexus: We would prefer to use db sort, but the specific issue here is with zeros unfortunately ("10" < "6").

durae: @csuloway-dnanexus gotcha! That makes sense

csuloway-dnanexus: Thank you, your review is much appreciated.

PR #1227, opened 2023-09-29
SCIPROD-1673: handle duplicate rsids in extract_assay germline

csuloway-dnanexus:
Update `dx extract_assay germline` to handle duplicate rsids
- Remove duplicate rsids prior to API call
- `--json-help` includes note that duplicate rsids are allowed and handled silently


keegankelsey (approved):
lgtm - thanks @csuloway-dnanexus

PR #1228, opened 2023-09-29
Sciprod 1672 test output

JoeMulka:
Hi folks,
This change tests the write_expression_output function.  There are two positive tests that ensure the output is correct with and without the --sql flag, and several negative tests to ensure that the function rejects incorrect configuration.  Please let me know if there are any additional tests that you would like to be part of this ticket.
Thanks,
Joe

### DIFF
--- a/src/python/test/test_extract_expression.py
+++ b/src/python/test/test_extract_expression.py
@@ -28,6 +28,7 @@ import dxpy
 import copy
 import json
 import tempfile
+import csv
 
 import shutil
 from dxpy_testutil import cd, chdir
@@ -44,7 +45,7 @@ from dxpy.bindings.apollo.input_arguments_validation_schemas import (
 )
 from dxpy.bindings.apollo.expression_test_input_dict import CLIEXPRESS_TEST_INPUT
 from dxpy.bindings.apollo.vizserver_client import VizClient
-from dxpy.bindings.apollo.assay_filtering_conditions import write_expression_output
+from dxpy.cli.output_handling import write_expression_output
 
 dirname = os.path.dirname(__file__)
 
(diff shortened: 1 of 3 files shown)


src/python/test/test_extract_expression.py
```diff
@@ -236,6 +268,146 @@ def common_input_args_test(self, input_argument_dict, expected_error_message):
...
+                output_listdict_or_string=self.vizserver_data_mock_response["results"],
+                save_uncommon_delim_to_txt = False,
+                error_handler=self.common_value_error_handler
+            )
+
+        err_msg = str(cm.exception).strip()
+        self.assertEqual(expected_error_message, err_msg)
+
+    def test_incorrect_file_extension(self):
+        expected_error_message = 'File extension ".tsv" does not match delimiter ","'
```
ashkan0x, line 395: This test can be removed, because we don't (per spec) attempt to verify/update suffix if filename is already given.

src/python/test/test_extract_expression.py
```diff
@@ -236,6 +268,146 @@ def common_input_args_test(self, input_argument_dict, expected_error_message):
...
+        }
+        expected_result = "SELECT `expression_1`.`feature_id` AS `feature_id`, `expression_1`.`sample_id` AS `sample_id`, `expression_1`.`value` AS `expression`, `expr_annotation_1`.`strand` AS `strand` FROM `database_gypg8qq06j8kzzp2yybfbzfk__enst_short_multiple_assays2`.`expression` AS `expression_1` LEFT OUTER JOIN `database_gypg8qq06j8kzzp2yybfbzfk__enst_short_multiple_assays2`.`expr_annotation` AS `expr_annotation_1` ON `expression_1`.`feature_id` = `expr_annotation_1`.`feature_id` WHERE `expression_1`.`value` >= 1 AND `expr_annotation_1`.`chr` = '1' AND (`expr_annotation_1`.`end` BETWEEN 7 AND 250000000 OR `expr_annotation_1`.`start` BETWEEN 7 AND 250000000 OR `expr_annotation_1`.`end` >= 250000000 AND `expr_annotation_1`.`start` <= 7)"
+        output_path = os.path.join(
+            self.general_output_dir, "extract_assay_expression_sql.csv"
+        )
+        # Generate the formatted output file
+        write_expression_output(
```
ashkan0x, line 307: Missing 1 required positional argument: 'output_listdict_or_string' (`sql_mock_response` should be passed to `output_listdict_or_string`)

ashkan0x (approved):
LGTM! 👍

PR #1237, opened 2023-10-04
SCIPROD-1741 Update expected sql in extract_dataset tests

Niranjanpandeshwar:
Update expected sql in extract_dataset tests and re-enable the test


durae (commented):
LGTM

PR #1238, opened 2023-10-04
SCIPROD-1731 function created for widening output data table

JoeMulka:
Hi folks,
This PR adds a function which converts the regular vizserver output into a "wider" format, with one column for sample ID, and a column for each feature_id, with the contents of those columns being the expression value.  I'm going to make a few quick unit tests for this, just so I can be sure it works on input with multiple samples, especially if the samples have overlapping feature ids.
Thanks,
Joe

### DIFF
--- a/src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
+++ b/src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
@@ -19,4 +19,13 @@ def expression_transform(results_list):
         for (feature_id,expression) in transformed_dict[sample].items():
             samp_row[feature_id] = expression
         dict_list.append(samp_row)
-    return dict_list
+    
+    # Get the column names that the output writer will use to generate the table
+    colnames = set()
+    for row in results_list:
+        colnames.add(row["feature_id"])
+    colnames = list(colnames)
+    # add sample_id to front
+    colnames.insert(0,"sample_id")
+
+    return (dict_list,colnames)
--- a/src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
+++ b/src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
@@ -21,10 +21,12 @@ def expression_transform(results_list):
         dict_list.append(samp_row)
     
     # Get the column names that the output writer will use to generate the table
-    colnames = set()
+    # Can't use set here because we want the order of the columns to be deterministic
+    colnames = []
     for row in results_list:
-        colnames.add(row["feature_id"])
-    colnames = list(colnames)
+        row_colname = row["feature_id"] 
+        if not row_colname in colnames:
+            colnames.append(row_colname)
     # add sample_id to front
     colnames.insert(0,"sample_id")
 
--- a/src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
+++ b/src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
@@ -52,6 +52,6 @@ def expression_transform(results_list):
     for dict_row in dict_list:
         for colname in colnames:
             if colname not in dict_row:
-                dict_row[colname] = "None"
+                dict_row[colname] = None
 
     return (dict_list,colnames)
--- a/src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
+++ b/src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
@@ -1,4 +1,3 @@
-from copy import deepcopy
 
 def expression_transform(results_list):
     """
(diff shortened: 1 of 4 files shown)


src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
```diff
@@ -0,0 +1,63 @@
...
+        if entry["sample_id"] not in transformed_dict:
+            transformed_dict[entry["sample_id"]] = {entry["feature_id"]:entry["expression"]}
+        else:
+            transformed_dict[entry["sample_id"]][entry["feature_id"]] = entry["expression"]
+
+    # Modify the above dictionary into a list of dictionaries, essentially moving the key of each
+    # dict into the dict itself
+    dict_list = []
+    for sample in transformed_dict:
+        samp_row = deepcopy(transformed_dict[sample])
```
ashkan0x, line 49: I think deepcopy might be too slow for very large dicts.

src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
```diff
@@ -0,0 +1,63 @@
...
+    # create a dict of the form
+    # {
+    #   <sample_id>:{
+    #       <feature_id_1>:<expression_1>,
+    #       <feature_id_2>:<expression_2>,
+    #        etc.
+    #   }
+    # 
+    # }
+    for entry in results_list:
```
ashkan0x, line 39: How about starting with a `set` to store all unique `feature_id`s?

src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
```diff
@@ -0,0 +1,63 @@
...
+    dict_list = []
+    for sample in transformed_dict:
+        samp_row = deepcopy(transformed_dict[sample])
+        samp_row["sample_id"] = sample
+        dict_list.append(samp_row)
+    
+    # Get the column names that the output writer will use to generate the table
+    # Can't use set here because we want the order of the columns to be deterministic
+    colnames = []
+    for row in results_list:
```
ashkan0x, line 56: I think if you start with a `set` of unique `feature_id`s then you might be able to run the second for loop right within the first one, and then `deepcopy` might not be necessary anymore. But this will probably require some changes in the current code in both loops.

src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
```diff
@@ -0,0 +1,63 @@
...
+        row_colname = row["feature_id"] 
+        if not row_colname in colnames:
+            colnames.append(row_colname)
+    # add sample_id to front
+    colnames.insert(0,"sample_id")
```
ashkan0x, line 61: There should be a condition to assign `None` to non-existent `value`s.

Assuming a for loop going over the `set` of unique feature_ids I mentioned above:

```transformed_dict[current_unique_feature_id] = row['value'] if row['feature_id'] == current_unique_feature_id else None```

File: src/python/dxpy/bindings/apollo/expression_matrix_transformation.py (hunk 3 above)
Comment (ashkan0x) [reply src/python/dxpy/bindings/apollo/expression_matrix_transformation.py:56]: One way would be to just start filling in `sample_id`s right here immediately after `for row in results_list`. Then the second nested for loop would just fill in the `value` based on the current, matched `feature_id`.

File: src/python/dxpy/bindings/apollo/expression_matrix_transformation.py (hunk 2 above)
Comment (ashkan0x) [reply src/python/dxpy/bindings/apollo/expression_matrix_transformation.py:39]: That way there would be no need to keep track of `colnames` since all dicts will have the same set of `feature_id`s in the same, exact order. We would just fill in `None` whenever there is no matching `value` for the corresponding sample_id/feature_id pair.

In this case `write_expression_output` will also work without any tweaks (because all dicts will have the same keys in the same order).

File: src/python/dxpy/bindings/apollo/expression_matrix_transformation.py (hunk 1 above)
Comment (JoeMulka) [reply src/python/dxpy/bindings/apollo/expression_matrix_transformation.py:49]: https://stackoverflow.com/questions/57053648/what-is-easier-for-machine-dictionary-comprehension-or-deepcopy-of-existing-di
You're right, I'll replace with a dict comprehension

File: src/python/dxpy/bindings/apollo/expression_matrix_transformation.py (hunk 2 above)
Comment (JoeMulka) [reply src/python/dxpy/bindings/apollo/expression_matrix_transformation.py:39]: For the first point, I can use set in the first loop and get rid of the third loop, I just have to convert to list after the loop and then add in the sample_id key.
For the second point, I don't think I can guarantee that the OrderedDicts will be made with the feature_ids in the same order.  I don't think I can make any assumptions about the order that the vizserver will return the data, so I'm assuming that the loop could encounter any sample_id/feature_id pair at any location in the list.  I could have a loop go through each ordered dict and add the missing feature_id keys with None values, and then sort each OrderedDict in the list.  We basically get this for free in the DictWriter by setting `restval = 0` though.  I could alternatively set `restval = None` there.  I think the decision is basically whether we add in Nones and sort the list of dicts ourselves, or we just pass the column names to the output handler and let DictWriter do the sorting and missing data infill.

File: src/python/dxpy/cli/output_handling.py [hunk 5]
```diff
@@ -140,7 +144,8 @@ def write_expression_output(
             "delimiter": DELIMITER,
             "lineterminator": OS_SPECIFIC_LINE_SEPARATOR,
             "quoting": QUOTING,
-            "quotechar": QUOTE_CHAR,
+            "quotechar": QUOTE_CHAR,\
```
ashkan0x, line 147:
```suggestion
            "quotechar": QUOTE_CHAR,
```

src/python/dxpy/cli/output_handling.py
```diff
@@ -140,7 +144,8 @@ def write_expression_output(
...
+            "restval":"None"
```
ashkan0x, line 148: Might need to check with PL, but I think it's fine if cell is empty, so `None` instead of `"None"`

```suggestion
            "restval":"None"
```

src/python/dxpy/bindings/apollo/expression_matrix_transformation.py
```diff
@@ -0,0 +1,53 @@
...
+    # {
+    #   <sample_id>:{
+    #       "sample_id":<sample_id>,
+    #       <feature_id_1>:<expression_1>,
+    #       <feature_id_2>:<expression_2>,
+    #        etc.
+    #   }
+    # 
+    # }
+    colnames = set()
```
ashkan0x, line 39: How about collecting all colnames already here?

```suggestion
    f_ids = set(row["feature_id"] for row in results_list)
    colnames=["sample_id"]
    colnames.extend(list(f_ids))
```

ashkan0x (approved):
Thanks, LGTM! 👍

PR #1244, opened 2023-10-07
COMPUTE-110 Reduce number of `/file-xxxx/describe` calls in `DXFile.read()`

kpjensen:
The function `object_exists_in_project()` triggers a `/file-xxxx/describe` API call, which is currently called every time the download URL is retrieved via `get_download_url()` even if it is already cached. Rearrange the logic so this is only called if the URL is not cached or has expired to reduce the number of unnecessary API calls.


wormsik (approved):
LGTM 👍

PR #1264, opened 2023-11-03
COMPUTE-185 Set `python_requires`

kpjensen:
Remove `jq` submodule. 
Remove `xenial-deb` build target. 
Set `python_requires` in setup.py for final release with Python 2.7 support.


wormsik (approved):
LGTM 👍

PR #1270, opened 2023-11-13
COMPUTE-191 Drop python 2.7 from dxpy

kpjensen:


build/run_python_integration_tests.py
```diff
@@ -3,6 +3,8 @@
 '''
 Runs Python integration tests
 
+Assumes  `make python src_libs && source build/py_env/bin/activate`
```
wormsik, line 6: does order matter? as in https://github.com/dnanexus/cells/pull/964/files you use `make src_libs python`

src/python/setup.py
```diff
@@ -104,7 +97,7 @@ def make_valid_pypi_version(raw):
     entry_points = {
         "console_scripts": scripts,
     },
-    python_requires = '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
+    python_requires = '>=3.6',
     install_requires = dependencies,
     extras_require={
         'pandas': ["pandas==1.3.5; python_version>='3.7'", "pandas>=0.23.3,<=0.25.3; python_version>='3.5.3' and python_version<'3.7'", "pandas>=0.23.3,< 0.25.0; python_version<'3.5.3'"],
```
wormsik, line 103: maybe we can remove pandas for python < 3.5.3

build/run_traceability_integration_tests.sh
```diff
@@ -8,11 +8,9 @@ export DNANEXUS_INSTALL_PYTHON_TEST_DEPS="yes"
 export DX_USER_CONF_DIR="${TOOLKIT_ROOT_DIR}/dnanexus_config_relocated"
 
 cd $TOOLKIT_ROOT_DIR
-make python
+make src_libs python
 
-source build/py_env2.7/bin/activate
-source environment
+source build/py_env7/bin/activate
```
wormsik, line 13:
```suggestion
source build/py_env/bin/activate
```

src/python/setup.py
```diff
@@ -104,7 +97,7 @@ def make_valid_pypi_version(raw):
     entry_points = {
         "console_scripts": scripts,
```
wormsik, line 98: do we also want to modify hashbang lines in scripts? I don't think they are actually used, so we can perphaps even remove them?

build/run_python_integration_tests.py
```diff
@@ -3,6 +3,8 @@
 '''
```
wormsik, line 3: do we plan to update hashbang line above?

src/python/test/test_dx_bash_helpers.py
```diff
@@ -120,6 +120,7 @@ def update_environ(**kwargs):
 
 
 @unittest.skipUnless(testutil.TEST_RUN_JOBS, 'skipping tests that would run jobs')
+@unittest.skipIf(True, 'Temporarily disabled')
```
wormsik, line 123: maybe use just `@unittest.skip()`?

build/run_python_integration_tests.py
kpjensen: Order does not matter.

wormsik (approved):
LGTM 👍

PR #1272, opened 2023-11-14
HOTFIX - APPS-2043 collect images

Gvaihir:
Remove docker func. To be added in the next pr


YuxinShi0423 (changes requested):
I think you can just comment those lines out, so you can pick them up later in your PR. But LGTM now. 

And if we want to make it in the next release, the changelog needs to be updated as well.

Gvaihir:
@YuxinShi0423 it is in the changelog for release [364](https://github.com/dnanexus/dx-toolkit/blob/master/CHANGELOG.md#3640---20231113).
Pls reapprove

YuxinShi0423:
> @YuxinShi0423 it is in the changelog for release [364](https://github.com/dnanexus/dx-toolkit/blob/master/CHANGELOG.md#3640---20231113). Pls reapprove

Need to move the line `Docker image can be cached on the platform` to 0.365?

YuxinShi0423 (approved):

PR #1277, opened 2023-11-20
COMPUTE-219 Remove dxpy dependencies requests and cryptography

kpjensen:
Remove both `requests` and `cryptography` from dxpy requirements. 
Fixed pytest warnings and removed some python2 specific code.

### DIFF
--- a/src/python/doc/dxpy.rst
+++ b/src/python/doc/dxpy.rst
@@ -16,14 +16,6 @@ This Python 2.7 package includes three key modules:
 
 It has the following external dependencies:
 
-* :mod:`requests`: To install on Linux, use ``sudo pip install
-  requests``. Other installation options can be found at
-  http://docs.python-requests.org/en/latest/user/install
-
-* :mod:`futures`: To install on Linux, use ``sudo pip install futures``.
-  Other installation options can be found at
-  http://code.google.com/p/pythonfutures/
-
 
 Package Configuration
 ---------------------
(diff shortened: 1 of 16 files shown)


src/python/doc/dxpy.rst
```diff
@@ -16,14 +16,6 @@ This Python 2.7 package includes three key modules:
 
 It has the following external dependencies:
```
wormsik, line 18:
```suggestion
```

wormsik: plus I would also remove Python 2.7 on line 4

wormsik (changes requested):

src/python/requirements.txt
```diff
@@ -3,9 +3,8 @@ argcomplete>=1.9.4,<2.0.0; python_version < "3.10"
 websocket-client==0.54.0
 python-dateutil>=2.5
 psutil>=5.9.3
-requests>=2.8.0,<2.28; python_version < "3.7"
-requests>=2.8.0,<2.29; python_version >= "3.7"
-cryptography>=3.4.2,<41; python_version > "3.5"
+certifi
+urllib3>=1.26.18,<2.1.0
```
wormsik, line 7: why don't we pin the most recent version? is it because of SecureTransport TLS deprecation?

kpjensen: I have not done extensive testing with `2.1.0` which has many deprecations, but seems okay so far in testing. `urllib3` deprecations on new minor version updates have caused dxpy errors in the past. 
`v1.26.x` is still maintained and `<2` could be required by other packages a user may have installed.

wormsik (approved):
LGTM 👍

PR #1293, opened 2023-11-30
SCIPROD-1870 cliexpress insufficient permissions traceback suppression

JoeMulka:
Hi folks,
Super simple change that prevents the "variable referenced before assignment" traceback that was being shown to the user when there is insufficient permissions for the dataset.  Moving the return inside the try block means that it will only run if hte response variable has actually been set.
Thanks,
Joe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Unreleased
 
+* Suppressing traceback when extract_assay expression attempts to access a dataset it does not have access to 
+
 ### Fixed
 
 * Improved error messaging for `dx extract_assay expression`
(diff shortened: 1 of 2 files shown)


Niranjanpandeshwar (approved):
LGTM!

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 ### Fixed
 
 * Improved error messaging for `dx extract_assay expression`
+* Suppressing traceback when extract_assay expression attempts to access a dataset it does not have access to
```
ashkan0x, line 12:
```suggestion
* Suppressing traceback when `dx extract_assay expression` attempts to access a dataset it does not have access to 
```

ashkan0x (approved):

PR #1306, opened 2023-12-20
PTFM-37328: Added support for setting monthlyStorageLimit when creating project

ajbieber:
- Added command line option for defining `monthlyStorageLimit` on project creation
- Added float type check for command line input


arpradhan (commented):
Should we add a test to https://github.com/dnanexus/dx-toolkit/blob/master/src/python/test/test_dxpy.py#L121?

PR #1322, opened 2024-02-08
SCIPROD-1892 [GENOMOD] remove hom-alt genotype type

piader1:
Removed references of `hom-alt` type, updated json.help string and added test covering `--retrieve-genotype` flag (runs locally).


csuloway-dnanexus (approved):
LGTM!

PR #1323, opened 2024-02-12
APPS-2464 fix build nextflow with extra_args

YuxinShi0423:


YuxinShi0423:
https://jenkins-vstg.internal.dnanexus.com/view/dx-toolkit%20release%20board/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/705/
https://github.com/dnanexus/nextaur-app/actions/runs/7875442759/job/21487337593

src/python/dxpy/scripts/dx_build_app.py
```diff
@@ -1222,7 +1222,8 @@ def _build_app(args, extra_args):
 
 def build(args):
     try:
-        executable_id = _build_app(args, process_extra_args(args) if args.extra_args else {})
+        process_extra_args(args)
```
YuxinShi0423, line 1225: `process_extra_args` updates args but not returns it. 
other references:
https://github.com/dnanexus/dx-toolkit/blob/master/src/python/dxpy/scripts/dx.py#L3560
 https://github.com/dnanexus/dx-toolkit/blob/master/src/python/scripts/dx-jobutil-new-job#L73

Evanj80 (approved):

src/python/dxpy/scripts/dx_build_app.py
mhrvol: process_extra_args is now called even if args.extra_args does not exist.

YuxinShi0423: > process_extra_args

yes in https://github.com/dnanexus/dx-toolkit/blob/a83f740f986d750caa64a02ad759dc34f578f95e/src/python/dxpy/cli/parsers.py#L219
the parsing happens only when extra args are not None @mhrvol

mhrvol (approved):
LGTM

src/python/test/test_dxclient.py
```diff
@@ -7925,6 +7925,17 @@ def test_build_applet(self):
         self.assertEqual(applet_describe["id"], applet_describe["id"])
         self.assertEqual(applet_describe["name"], "minimal_applet")
 
+    @pytest.mark.TRACEABILITY_MATRIX
```
YuxinShi0423, line 7928: @kpjensen I added this test for generic applet build.

YuxinShi0423:
> What was the NF combination that caused failure? Can you add an integration test to cover this, or was it already covered and failing?

the failure is not limited to NF but to all app/let building with extra args. Just added one test. Thanks for the suggestion!

kpjensen (approved):

PR #1332, opened 2024-02-22
DEVEX-2286 Exit SSH loop if job in terminal state

kpjensen:


wormsik (approved):
small message improvement, otherwise LGTM 👍

PR #1340, opened 2024-03-06
APPS-2510 Fix dx-jobutil-get-identity-token - subject_claims is empty array

r-i-v-a:
Jira --> APPS-2510


src/python/scripts/dx-jobutil-get-identity-token
```diff
@@ -48,10 +48,10 @@ def main(aud, subject_claims):
 
     if subject_claims is not None:
```
wormsik, line 49: if there must be at least one subject claim per request, this condition should be removed and we should require at least one `--subject_claims` argument...

wormsik (approved):
LGTM 👍

bayou-brogrammer (approved):

PR #1348, opened 2024-03-08
COMPUTE-367

AroPho:


src/python/dxpy/scripts/dx.py
```diff
@@ -2084,26 +2085,56 @@ def upload(args, **kwargs):
...
+    # Send the file
+    with open(filename, 'rb') as fd:
+        result = s3.put_object(
+            Bucket=bucket_name,
+            Key=path,
+            Body=fd
+        )
+
+    if result['ResponseMetadata']['HTTPStatusCode'] == 200:
+        response = "https://{0}.s3.us-east-2.amCCazonaws.com/{1}".format(bucket, s3_key)
```
YuxinShi0423, line 2111: is the `amCCazonaws` a typo here?

src/python/dxpy/scripts/dx.py
```diff
@@ -2084,26 +2085,56 @@ def upload(args, **kwargs):
...
+    else:
+       response = False
```
YuxinShi0423, line 2113: better to make this response an empty string?

src/python/dxpy/scripts/dx.py
```diff
@@ -2084,26 +2085,56 @@ def upload(args, **kwargs):
...
+        print(project)
+        print(folder_or_volume)
+        print(name)
+        print(is_v2_path)
+        print(etag)
+        print(args.path)
         if name is None and args.filename != '-':
             name = os.path.basename(args.filename)
+        if is_v2_path:
+            upload_one_v2_file(folder_or_volume, args.path, etag, args)
```
YuxinShi0423, line 2137: We are still gonna call `upload_local_file` after calling `upload_one_v2_file` here, right? 
@kpjensen I am still wondering if it is better to use `upload_local_file` as the common entry point for all uploading (as its name suggests) and add this v2 path uploader inside. Suggestions?

YuxinShi0423 (changes requested):

src/python/dxpy/scripts/dx.py
AroPho: we will not be calling upload_local_file after calling upload_one_v2_file I forgot to put a return

PR #1350, opened 2024-03-13
APPS-2514 awscli nextflow applets

mhrvol:


r-i-v-a (approved):
There will be a separate PR to put the publishing process in nextaur-app and later to add the assets for prod, right?

PR #1355, opened 2024-03-21
SCIPROD-2016: [GENOMOD] `dx extract_assay germline --retrieve-genotype` inference use alleles to determine loci

csuloway-dnanexus:
Updated genotype inference in `dx extract_assay germline --retrieve-genotype` to use loci from the allele table to determine missing samples to infer excluded genotype types.


ashkan0x (approved):

mstevuliak-dx (commented):
Nice addition!

PR #1356, opened 2024-03-21
TIP-3662 Update dx cat to display content for symlink files.

austinlocke:
[TIP-3662](https://jira.internal.dnanexus.com/browse/TIP-3662) Update dx cat to display content for symlink files.


austinlocke:
I also tested dx download with writable symlink files with this change (by disabling the _download_symbolic_link function) and it seemed to work fine since the file had all the parts content. I think if we can successfully migrate all the older readable symlink files we can get rid of that logic all-together. If we want to keep it where older readable symlink files use _download_symbolic_link then we need to modify file-xxxx/describe to also return `uploadId` so we have some way to differentiate between readable and writable symlinks without referring to the project. @ksoni @durae

ksoni (commented):
Looks good to me.

wormsik:
how is this different from #1360 please?

austinlocke:
> how is this different from #1360 please?

It's not. This one can actually be closed and all changes are planned to be merged in the other PR. This change was done first before we tested dx download and realized the same changes was needed there.

PR #1359, opened 2024-03-28
APPS-2532 refresh credentials aws

mhrvol:
To be merged with Nextaur update (possibility for AWS login in headjob)


src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -512,6 +519,42 @@ nf_task_entry() {
...
+EOF
+  else
+    echo "No AWS environment variables available"
+  fi
+}
+
+automatic_aws_relogin() {
+  while true; do
+      sleep 3000 # 50 minutes, first login is done independently, so we wait first
+      if [ -f "$AWS_ENV" ]; then
```
r-i-v-a, line 555: Would expect file not to exist if AWS login not used for this job, right?

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -512,6 +519,42 @@ nf_task_entry() {
...
   dx set_properties ${DX_JOB_ID} nextflow_exit_code=$exit_code
   set -e
 }
+
+aws_login() {
+  if [ -f "$AWS_ENV" ]; then
+    source $AWS_ENV
+    # aws env file example values:
+    # "iamRoleArnToAssume", "roleSessionName", "jobTokenAudience", "jobTokenSubjectClaims", "region"
+    python3 -m pip install --upgrade 'git+https://github.com/dnanexus/dx-toolkit.git@master#egg=dxpy&subdirectory=src/python' # FIXME: Remove after it is available on staging.
```
r-i-v-a, line 532: This can be removed now

r-i-v-a (changes requested):
Just remove dxpy GitHub installation line

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
mhrvol: Correct, basically, this file is present only if it is created in the headjob.

I thought of not creating the aws_relogin process in case it is not created, but I believe it would create too much code and also we would need to check whether it exists or not. (as for example in the headjob we do not know whether it should exist until the headjob creates the aws_env file.).

And as this thread is mostly sleeping, I dont see the problem of having the additional thread.

r-i-v-a (approved):
now lgtm

PR #1360, opened 2024-03-28
TIP-3633 fix dx download symlinks

austinlocke:
TIP-3633 fix dx download symlinks


ksoni (commented):
Looks good to me.

src/python/dxpy/__init__.py
```diff
@@ -609,6 +609,13 @@ def ensure_ascii(i):
 
                 response = pool_manager.request(_method, encoded_url, headers=_headers, body=body,
                                                 timeout=timeout, retries=False, **kwargs)
+                # Handle redirection manually for symlink files
+                if response.status in range(300, 399):
+                    url = response.headers.get('Location')
+                    if url:
+                        # Make a new request to the URL specified in the Location header
+                        response = pool_manager.request(_method, url, headers=_headers, body=body,
```
wormsik, line 617: is there a chance for multiple redirects?

austinlocke: I don't think so, we've tested with very large multipart and single part files and it should only need to redirect on each call to nginx.

src/python/dxpy/__init__.py
```diff
@@ -609,6 +609,16 @@ def ensure_ascii(i):
...
+                # Handle redirection manually for symlink files
+                if response.status in range(300, 399):
+                    url = response.headers.get('Location')
+                    if url:
+                        if try_index > 0:
+                            logger.info("[%s] %s %s: Recovered after %d retries", time.ctime(), method, _url, try_index)
+                        # Make a new request to the URL specified in the Location header
+                        return DXHTTPRequest(url, body, method, headers, auth, timeout, use_compression, jsonify_data, want_full_response,
+                                      decode_response_body, prepend_srv, session_handler,
+                                      max_retries, always_retry, **kwargs)
```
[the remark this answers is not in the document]
wormsik: I would use explicit field names in case we would add any...

src/python/dxpy/__init__.py
```diff
@@ -609,6 +609,16 @@ def ensure_ascii(i):
...
                 response = pool_manager.request(_method, encoded_url, headers=_headers, body=body,
                                                 timeout=timeout, retries=False, **kwargs)
+                # Handle redirection manually for symlink files
+                if response.status in range(300, 399):
+                    url = response.headers.get('Location')
+                    if url:
+                        if try_index > 0:
+                            logger.info("[%s] %s %s: Recovered after %d retries", time.ctime(), method, _url, try_index)
+                        # Make a new request to the URL specified in the Location header
+                        return DXHTTPRequest(url, body, method, headers, auth, timeout, use_compression, jsonify_data, want_full_response,
```
wormsik, line 619: sorry but could you please rewrite to some variable, break this `while` loop and fire the new request once the whole block is done? I've realized that any exception thrown by this new call would trigger exception handling of the original request, which may lead to retry of everything and potentially some weird behaviour...

wormsik (changes requested):

src/python/dxpy/__init__.py
```diff
@@ -609,6 +610,12 @@ def ensure_ascii(i):
 
                 response = pool_manager.request(_method, encoded_url, headers=_headers, body=body,
                                                 timeout=timeout, retries=False, **kwargs)
+                # Handle redirection manually for symlink files
+                if response.status in range(300, 399):
+                    redirect_url = response.headers.get('Location')
+                    if try_index > 0:
+                        logger.info("[%s] %s %s: Recovered after %d retries", time.ctime(), method, _url, try_index)
```
wormsik, line 617:
```suggestion
```

wormsik: this should not be longer necessary as `finally` is processed before new request

src/python/dxpy/__init__.py
```diff
@@ -609,6 +610,12 @@ def ensure_ascii(i):
 
                 response = pool_manager.request(_method, encoded_url, headers=_headers, body=body,
                                                 timeout=timeout, retries=False, **kwargs)
+                # Handle redirection manually for symlink files
+                if response.status in range(300, 399):
+                    redirect_url = response.headers.get('Location')
```
wormsik, line 615: what is expected result when `status in range(300, 399)`, but `Location` header is not provided? so we know we should redirect, but don't know where...

for now, it will throw `AssertionError('Should never reach this line: should never break out of loop')` and it is not retried...can you please raise some better exception? I would say from here so it gets retried, because I'd expect this is faulty behaviour and might be retriable...

src/python/dxpy/__init__.py
```diff
@@ -609,6 +610,12 @@ def ensure_ascii(i):
...
+                    if not redirect_url:
+                        raise exceptions.UrllibInternalError("Location not found in redirect response", response.status)
+                    break
```
wormsik, line 618: sorry just a last minor thing: can you please move this block right below upgrade notify block (line 637 of the updated version)? I've realized it can be handy to set `req_id` first so in case `Location` is missing and exception is thrown so it is reported in error messages and could be used for debugging...

wormsik (changes requested):

wormsik (approved):
one small performance improvement, otherwise LGTM 👍

PR #1362, opened 2024-04-08
PTFM-37872: Add Project Storage Spending Limit Features

ajbieber:
Added the following features in regards to Project Storage Spending Limit

- A new command line argument `--monthly-storage-limit` for setting the project storage limit
- A new validation function for verifying input is a positive number
- New fields on the `describe` call


vtran1022 (approved):
lgtm 🥞

PR #1391, opened 2024-08-15
APPS-2588 verbose error

aogorodnikov-dna:
Added a way to propagate the error msg from the nextaur cli stdout.
Tests described in the ticket

### DIFF
--- a/src/python/dxpy/nextflow/collect_images.py
+++ b/src/python/dxpy/nextflow/collect_images.py
@@ -18,6 +18,7 @@
 #   under the License.
 
 import json
+import os.path
 import subprocess
 from dxpy.nextflow.ImageRefFactory import ImageRefFactory, ImageRefFactoryError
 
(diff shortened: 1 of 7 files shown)


src/python/dxpy/nextflow/collect_images.py
```diff
@@ -66,7 +67,9 @@ def run_nextaur_collect(resources_dir, profile, nextflow_pipeline_params):
     pipeline_params_arg = "pipelineParams={}".format(nextflow_pipeline_params) if nextflow_pipeline_params else ""
     profile_arg = "profile={}".format(profile) if profile else ""
     nextaur_cmd = " ".join([base_cmd, pipeline_params_arg, profile_arg])
-    _ = subprocess.check_output(nextaur_cmd, shell=True)
+    process = subprocess.run(nextaur_cmd, shell=True, capture_output=True, text=True)
```
mhrvol, line 70: Doesnt subprocess.check_output raise an exception in case nextaur_cmd ends with non-zero code, but subprocess.run doesnt? 

Maybe we could add `check=true` to the run arguments?

r-i-v-a (changes requested):
consider Marek's comment

src/python/dxpy/nextflow/collect_images.py
aogorodnikov-dna: it was not raising anything, because exit code from that command is actually 0 even if it exits with error. And the msg is dumped to `stdout`, not `stderr` - this is just an annoying behavior of nextflow calling its own plugin `nf-validation` and not parsing the error correctly. So `subprocess` sees exit code 0 and only stdout.

mhrvol: sounds good, I'd still add true check=true (that should raise an error in case of nonzero code I believe). but that way or the other, LGTM

mhrvol (approved):

r-i-v-a (approved):

PR #1392, opened 2024-08-23
DEVEX-2387 Provide project-id while checking uploaded parts

kpjensen:


src/python/dxpy/bindings/__init__.py
```diff
@@ -660,7 +660,7 @@ def _wait_on_close(self, timeout=3600*24*1, **kwargs):
     def _wait_until_parts_uploaded(self, timeout=255, **kwargs):
         elapsed = 0
         i = 0
-        describe_input = {"fields": {"parts": True, "state": True}}
+        describe_input = {"project": self._proj, "fields": {"parts": True, "state": True}}
```
wormsik, line 663: are we sure that `self._proj` is always set in all possible code paths? in case it would (for any reason) be `null`, the API call will fail with `dxpy.exceptions.InvalidInput: Expected key "project" of input to be a string, code 422.`

wormsik (approved):
if we can rely on project being always set, then LGTM 🚀

src/python/dxpy/bindings/__init__.py
kpjensen: Changed to check `is not None` before adding to the input.

PR #1396, opened 2024-09-19
COMPUTE-963 Provide project-id in `dx get_details` 

kpjensen:


src/python/dxpy/scripts/dx.py
```diff
@@ -1512,14 +1512,14 @@ def set_visibility(args):
...
-    _project, _folderpath, entity_result = try_call(resolve_existing_path,
+    project, _folderpath, entity_result = try_call(resolve_existing_path,
                                                     args.path, expected='entity')
 
     if entity_result is None:
         err_exit(fill('Could not resolve "' + args.path + '" to a name or ID'), 3)
 
     try:
-        print(json.dumps(dxpy.DXHTTPRequest('/' + entity_result['id'] + '/getDetails', {}), indent=4))
+        print(json.dumps(dxpy.DXHTTPRequest('/' + entity_result['id'] + '/getDetails', {'project':project}), indent=4))
```
wormsik, line 1522:
```suggestion
        print(json.dumps(dxpy.DXHTTPRequest('/' + entity_result['id'] + '/getDetails', {'project': project}), indent=4))
```

wormsik (approved):
in case the `project` will always be correctly resolved (I cannot think about any case when it won't, but `getDetails` does not accept `{"project": null}`), then LGTM 🚀

wormsik (approved):
LGTM 🚀

PR #1399, opened 2024-10-01
[COMPUTE-993] nvidiaDriver field support in dx-toolkit

jsitarova-dnanexus:


src/python/test/test_dxpy.py
```diff
@@ -1346,6 +1351,7 @@ def main():
         self.assertEqual(len(jobdesc["properties"]), 1)
         self.assertEqual(jobdesc["properties"]["$dnanexus_link.foo"], "barbaz")
         self.assertEqual(jobdesc["priority"], "normal")
+        self.assertEqual(jobdesc["systemRequirements"]["*"]["nvidiaDriver"], "R535")
```
jsitarova-dnanexus, line 1354: Dont have successful run as the tests are not running regularly, but the test was passing locally.

src/python/dxpy/bindings/dxjob.py
```diff
@@ -38,14 +38,16 @@
...
 #########
 # DXJob #
 #########
 
-def new_dxjob(fn_input, fn_name, name=None, tags=None, properties=None, details=None,
-              instance_type=None, depends_on=None,
-              cluster_spec=None, fpga_driver=None, system_requirements=None, system_requirements_by_executable=None,
-              **kwargs):
+
+def new_dxjob(fn_input, fn_name, name=None, tags=None, properties=None, details=None, instance_type=None,
```
jsitarova-dnanexus, line 48: manually tested using this applet:
https://github.com/dnanexus/cells/pull/1207/files
- which installs dx-toolkit from this branch and call new_dxjob function inside the applet

Job id (default R470 nvidia driver): https://staging.dnanexus.com/panx/projects/GqbGJV80Xqy5Qzz7B9pqFgyQ/monitor/job/GqzvZFQ0XqyKjYQ2ffYQkkg4
Child job id (expected R535 nvidia driver chnaged during the runtime of the child job creation):
https://staging.dnanexus.com/panx/projects/GqbGJV80Xqy5Qzz7B9pqFgyQ/monitor/job/GqzvbX00XqyG3G980QbyZFB9/try/0

src/python/test/test_dxpy.py
```diff
@@ -1346,6 +1351,7 @@ def main():
         self.assertEqual(len(jobdesc["properties"]), 1)
         self.assertEqual(jobdesc["properties"]["$dnanexus_link.foo"], "barbaz")
         self.assertEqual(jobdesc["priority"], "normal")
+        self.assertEqual(jobdesc["systemRequirements"]["*"]["nvidiaDriver"], "R470")
```
jsitarova-dnanexus, line 1354: Dont have successful run as the job tests are not running anywhere right now, but the test was passing locally.

src/python/dxpy/bindings/dxapplet.py
```diff
@@ -195,7 +197,7 @@ def run(self, executable_input, project=None, folder=None, name=None, tags=None,
             depends_on=None, allow_ssh=None, debug=None, delay_workspace_destruction=None, priority=None, head_job_on_demand=None,
             ignore_reuse=None, ignore_reuse_stages=None, detach=None, cost_limit=None, rank=None, max_tree_spot_wait_time=None,
             max_job_spot_wait_time=None, preserve_job_outputs=None, detailed_job_metrics=None, extra_args=None,
-            fpga_driver=None, system_requirements=None, system_requirements_by_executable=None, **kwargs):
+            fpga_driver=None, nvidia_driver=None, system_requirements=None, system_requirements_by_executable=None, **kwargs):
```
jsitarova-dnanexus, line 200: Tested in src/python/test/test_dxpy.py - test_run_dxapplet_and_job_metadata

src/python/dxpy/bindings/dxapplet.py
```diff
@@ -58,11 +58,13 @@ def _get_run_input_common_fields(executable_input, **kwargs):
             if kwargs.get(arg) is not None:
                 run_input[arg] = kwargs[arg]
 
-        if kwargs.get('instance_type') is not None or kwargs.get('cluster_spec') is not None or kwargs.get('fpga_driver') is not None:
+        if (kwargs.get('instance_type') is not None or kwargs.get('cluster_spec') is not None or
```
mhrvol, line 61: We could use 
`if any(kwargs.get(key) for key in ['instance_type', 'cluster_spec', 'fpga_driver', 'nvidia_driver']):`

(note that this checks for "truthy values", not just Nones, previously the value "" would pass the check, now it wouldnt, but that should not really matter)

alternatively : `if any(kwargs.get(key) is not None for key in ['instance_type', 'cluster_spec', 'fpga_driver', 'nvidia_driver']):`

mhrvol (commented):
I see there are some places in tests where we test fpga_driver, but not nvidia_driver (like test_dx_bash_helpers) - does it make sense to put the tests here as well?

Also in test_dxclient.py we have some tests for dx run --clone, could we add one more test to check that this works properly even with --clone

src/python/dxpy/bindings/dxapplet.py
jsitarova-dnanexus: Looks much better - fixed. I used the format with 'is not None' to be sure we dont change behaviour for any possible case.

src/python/dxpy/bindings/dxjob.py
```diff
@@ -38,14 +38,15 @@
...
+              depends_on=None, cluster_spec=None, fpga_driver=None, nvidia_driver=None, system_requirements=None,
```
wormsik, line 48: can you please move `nvidia_driver` argument to the end of the arguments list in all cases? adding it in the middle is actually API breaking change and can cause troubles if anyone is using this method without keyword args...

wormsik (changes requested):

src/python/dxpy/bindings/dxjob.py
jsitarova-dnanexus: sure, fixed.

jsitarova-dnanexus:
> I see there are some places in tests where we test fpga_driver, but not nvidia_driver (like test_dx_bash_helpers) - does it make sense to put the tests here as well?
> 
> Also in test_dxclient.py we have some tests for dx run --clone, could we add one more test to check that this works properly even with --clone

Right, I added 1 more TC to test/test_dx_bash_helpers.py::TestDXJobutilNewJob::test_job_arguments
+
I added 1 new test for dx run --clone - test/test_dxclient.py::TestDXClientRun::test_dx_run_clone_nvidia_driver

src/python/dxpy/bindings/dxjob.py
```diff
@@ -179,11 +184,12 @@ def new(self, fn_input, fn_name, name=None, tags=None, properties=None, details=
                 req_input["tags"] = tags
             if properties is not None:
                 req_input["properties"] = properties
-            if instance_type is not None or cluster_spec is not None or fpga_driver is not None:
+            if instance_type is not None or cluster_spec is not None or fpga_driver is not None or nvidia_driver is not None:
```
wormsik, line 187: I would personally use the same way with `any` as above

src/python/dxpy/bindings/dxapplet.py
```diff
@@ -58,11 +58,13 @@ def _get_run_input_common_fields(executable_input, **kwargs):
             if kwargs.get(arg) is not None:
                 run_input[arg] = kwargs[arg]
 
-        if kwargs.get('instance_type') is not None or kwargs.get('cluster_spec') is not None or kwargs.get('fpga_driver') is not None:
+        if any(kwargs.get(key) is not None for key in
+               ['instance_type', 'cluster_spec', 'fpga_driver', 'nvidia_driver']):
```
wormsik, line 62: please put it to a single line

wormsik (approved):
just a few minor comments, otherwise LGTM 🚀

src/python/dxpy/bindings/dxjob.py
jsitarova-dnanexus: agree, fixed.

src/python/dxpy/bindings/dxapplet.py
jsitarova-dnanexus: Yeah, I think the editor split it as the line is longer than 120chars, but I checked and its 121 so I guess its still OK to have it on 1 line.

mhrvol (approved):
LGTM

PR #1409, opened 2024-10-18
DEVEX-2418

vkuban-dx:
Fix run of globalworkflow when the underlying workflow will be searched in dx env Current workspace instead of its onw container in the given region.

### DIFF
--- a/src/python/dxpy/bindings/dxglobalworkflow.py
+++ b/src/python/dxpy/bindings/dxglobalworkflow.py
@@ -222,9 +222,13 @@ class DXGlobalWorkflow(DXObject, DXExecutable):
             raise DXError('DXGlobalWorkflow: the global workflow {} is not enabled in region {}'.format(
                 self.get_id(), region))
 
-        underlying_workflow_id = describe_output['regionalOptions'][region]['workflow']
-        self._workflow_desc_by_region = dxpy.DXWorkflow(underlying_workflow_id)
-        return dxpy.DXWorkflow(underlying_workflow_id)
+        underlying_workflow = describe_output['regionalOptions'][region]
+        dxworkflow = dxpy.DXWorkflow(
+            dxid=underlying_workflow['workflow'],
+            project=underlying_workflow['resources'],
+        )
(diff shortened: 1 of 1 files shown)


src/python/dxpy/bindings/dxglobalworkflow.py
```diff
@@ -222,9 +222,13 @@ def get_underlying_workflow(self, region, describe_output=None):
...
 
-        underlying_workflow_id = describe_output['regionalOptions'][region]['workflow']
-        self._workflow_desc_by_region = dxpy.DXWorkflow(underlying_workflow_id)
-        return dxpy.DXWorkflow(underlying_workflow_id)
+        underlying_workflow = describe_output['regionalOptions'][region]
+        dxworkflow = dxpy.DXWorkflow(
+            dxid=underlying_workflow['workflow'],
+            project=underlying_workflow['resources'],
+        )
+        self._workflows_by_region[region] = dxworkflow
```
mhrvol, line 230: We previously set `self._workflow_desc_by_region` variable to the DXWorkflow

With this change we set `self._workflows_by_region[region]`

Why are we no longer setting `self._workflow_desc_by_region`?  It is used in `describe_underlying_workflow`.

vkuban-dx: There was a confusion between 2 attributes, see `def __init__()`
```
        # caches for underlying workflow instances and descriptions per region
        # (they are immutable for a given global workflow)
        self._workflows_by_region = {}
        self._workflow_desc_by_region = {}
```

mhrvol: Have you tested that the caching in both cases still work as expected? If so, then LGTM

vkuban-dx: yes, I did.

aogorodnikov-dna: If you are concerned about the behavior - you can wrap those in `try ... except`. Otherwise - could you identify the regression test that monitors this or maybe create one?

PR #1439, opened 2025-02-12
PTFM-39232: dxpy download - verify object integrity with per-part checksum

arpradhan:
https://jira.internal.dnanexus.com/browse/PTFM-39212

Files synced into DNANexus from an external S3 bucket will receive per-part checksums (if specified by customer).

This PR updates `dx download` to perform an integrity check on files that have a per-part checksum

Integration test cannot be added yet, because a file doc has per-part checksum only when a file is synced from an external S3 bucket. So as of now, there's no way to create a file doc with the per-part checksum.

Added a new pip package `crc32c` to compute the crc32c hash


vtran1022 (approved):
lgtm 🦊

haroldna (approved):
lgtm

PR #1445, opened 2025-02-25
[DEVEX-2463] dx-app-wizard should generate template with AEE 24.04 by default

jsitarova-dnanexus:
Tested locally


mhrvol (commented):
Changelog is not updated, otherwise LGTM

jsitarova-dnanexus:
> Changelog is not updated, otherwise LGTM

@mhrvol right, added

mhrvol (approved):
LGTM

PR #1446, opened 2025-03-05
DEVEX-2467 wrong file reads

mhrvol:


src/python/dxpy/bindings/dxfile.py
```diff
@@ -314,39 +314,19 @@ def __del__(self):
...
+        _buffer = ""
+        if self._binary_mode:
+            raise DXFileError("Cannot read lines when file opened in binary mode")
+        while True:
+            more = self.read(self._read_bufsize)
+            if not more:
+                break
+            _buffer += more
+            parts = _buffer.split("\n")
+            for line in parts[:-1]:
```
wormsik, line 326: will it make any time difference to keep `range(len(lines)) - 1`? to avoid copying the list without the last element...

in the worst case scenario when having list of ~8M elements

mhrvol: Good point, no need to create another array with n-1 elements.

wormsik (commented):
LGTM 🚀

wormsik (approved):

PR #1451, opened 2025-03-12
PTFM-39459: Perform per part checksum integrity checks on writable symlinks

arpradhan:
https://jira.internal.dnanexus.com/browse/PTFM-39459

Previous change to implement per-part checksum integrity checks applied only to readable symlinks.

This PR ensures that per-part checksum integrity checks are performed on writable symlinks instead.

This also removes the previous inefficiency of loading the readable symlink into memory after the download. So now the integrity check is performed while the writable symlink file is downloaded.

Also adds unit test for the integrity checks.

### DIFF
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -30,8 +30,6 @@ import traceback
 import warnings
 from collections import defaultdict
 import multiprocessing
-from random import randint
-from time import sleep
 import crc32c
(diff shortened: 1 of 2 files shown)


arpradhan:
Jenkins build succeeded: https://jenkins-vstg.internal.dnanexus.com/view/dx-toolkit%20release%20board/job/dx-toolkit/job/dxpy-branch-integration-tests-on-staging-20.04-python3/78/

PR #1459, opened 2025-03-24
COMPUTE-1113: Increase backoff timeout for throttled requests over attempts

windok:
https://jira.internal.dnanexus.com/browse/COMPUTE-1113

Reduce frequency of retrying throttled requests if apiserver continues to return 429/503 response code over multiple attempts.

[Java sdk](https://github.com/dnanexus/dx-toolkit/blob/master/src/java/src/main/java/com/dnanexus/DXHTTPRequest.java#L412) already has such increased backoff strategy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## [394.0] - beta
 
-* 
 ### Fixed
 
 * Update parsing of nextflow_schema.json file to be compatible with new schema convention
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,23 +6,27 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Unreleased
 
+### Changed
+
+* Increased timeout to retry throttled requests over attempts
+
 ## [394.0] - beta
 
+* 
 ### Fixed
 
 * Update parsing of nextflow_schema.json file to be compatible with new schema convention
 * `DXFile.read()` increase in API calls introduced in 392.0
+
+### Changed
+
+* Retry requests when received 429 response status code in Python, Java SDKs
 
 ## [393.0] - 2025.03.24
 
 ### Fixed
 
 * Preserve newlines in dxpy file iterator for large files
-
-### Changed
-
-* Retry requests when received 429 response status code in Python, Java SDKs
-* Increased timeout to retry throttled requests over attempts 
 
 ## [392.0] - 2025.03.10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 ### Changed
 
 * Retry requests when received 429 response status code in Python, Java SDKs
+* Increased timeout to retry throttled requests over attempts 
 
 ## [392.0] - 2025.03.10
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 ### Changed
 
 * Retry requests when received 429 response status code in Python, Java SDKs
+* Increased timeout to retry throttled requests over attempts
```
wormsik, line 22: shouldn't this be in `Unreleased` section? or do you want to cherry-pick to already released version?

src/python/dxpy/__init__.py
```diff
@@ -353,15 +353,21 @@ def _calculate_retry_delay(response, num_attempts):
     '''
     if response is not None and response.status in (503, 429) and 'retry-after' in response.headers:
         try:
-            return int(response.headers['retry-after'])
+            suggested_delay = int(response.headers['retry-after']) # 20-30 seconds
+
+            # By default, apiserver doesn't track attempts and doesn't provide increased timeout over attempts.
+            # So, increasing backoff for throttled requests up to x5 times from the original one.
+            # After 20th attempt delay will be from 100 to 150 seconds.
```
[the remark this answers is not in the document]
windok: apiserver returns retry after header in range 20-30 seconds what I mentioned above. If apiserver implementation will be changed, then potentially backoff after 20th attempt will be 6-354 seconds.

I agree that maybe these numbers are redundant here, but it gives overview of current values.

CHANGELOG.md
windok: Yeah, it should be there. I wasn't sure if beta version was released or not, so my previous PRs already added info there. I moved all my changes to `Unreleased` part because from what I see in git history no more releases were added and all my changes should be above

src/python/dxpy/__init__.py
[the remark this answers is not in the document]
wormsik:
```suggestion
            # The current implementation of apiserver returns a retry-after header ranging from 20 to 30 seconds.
            # Thus after the 20th attempt the delay will always be between 100 and 150 seconds.
```

src/python/dxpy/__init__.py
```diff
@@ -353,15 +353,21 @@ def _calculate_retry_delay(response, num_attempts):
...
+                else suggested_delay + 0.25 * min(num_attempts - 1, 20) * suggested_delay
         except ValueError:
             # In RFC 2616, retry-after can be formatted as absolute time
             # instead of seconds to wait. We don't bother to parse that,
             # but the apiserver doesn't generate such responses anyway.
             pass
     if num_attempts <= 1:
         return 1
-    num_attempts = min(num_attempts, 7)
+    num_attempts = min(num_attempts, 8) # after 8th attempt delay will be from 64 to 128 seconds
```
wormsik, line 370:
```suggestion
    num_attempts = min(num_attempts, 8)
    # After the 8th attempt the delay will always be between 64 and 128 seconds
```

src/python/dxpy/__init__.py
```diff
@@ -353,15 +353,21 @@ def _calculate_retry_delay(response, num_attempts):
     '''
     if response is not None and response.status in (503, 429) and 'retry-after' in response.headers:
         try:
-            return int(response.headers['retry-after'])
+            suggested_delay = int(response.headers['retry-after']) # 20-30 seconds
```
wormsik, line 356:
```suggestion
            suggested_delay = int(response.headers['retry-after'])
```

src/python/dxpy/__init__.py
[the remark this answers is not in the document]
windok: Applied suggested changes. Thanks for better wording!

wormsik (approved):
LGTM 🚀

PR #1463, opened 2025-03-28
Update dxcpp.cc to retry on 429 http error

rustykampa:
Changes:
- The file `src/cpp/dxcpp/dxcpp.cc` was updated to handle the 429 http error the same as the 503 is handled

### DIFF
--- a/src/cpp/dxcpp/dxcpp.cc
+++ b/src/cpp/dxcpp/dxcpp.cc
@@ -176,10 +176,6 @@ namespace dx {
       // Note: Initial value of "false" is just a dummy value, toRetry will always be re-init before being used.
       //       This dummy initial value is provided, to prevent some spurious warnings from clang
       bool toRetry = false;
-      // True if the request returns with a 503
-      // bool serviceUnavailable = false;
-      // bool servicdThrottled = false; // 429
-
       reqCompleted = true; // will explicitly set it to false in case request couldn't be completed
       try {
         DXLOG(logDEBUG) << "Attempting the actual HTTP request (countTries = " << countTries << ")...";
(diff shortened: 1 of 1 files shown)


src/cpp/dxcpp/dxcpp.cc
```diff
@@ -256,13 +262,13 @@ namespace dx {
         break;
       }
 
-      // 503 with Retry-After-- do not count such responses against the allowed
+      // 503 or 429 with Retry-After-- do not count such responses against the allowed
       // number of retries
-      if (serviceUnavailable) {
-        DXLOG(logWARNING) << "Service unavailable, waiting for " << retryAfterSeconds << " seconds : POST '" << url << "'";
+      if (req.responseCode == 503 || req.responseCode == 429) {
```
windok, line 267: It can be that `req` is not defined and has no `responseCode` property in case of some network error.

src/cpp/dxcpp/dxcpp.cc
```diff
@@ -163,6 +164,8 @@ namespace dx {
     bool reqCompleted; // did last request actually went through, i.e., some response was received)
     bool contentLengthMismatch;
     bool contentLengthMissing;
+    
+    // here...
```
windok, line 168: I think that is redundant comment, because it's useful only during this code review

windok (changes requested):

src/cpp/dxcpp/dxcpp.cc
```diff
@@ -176,7 +177,8 @@ namespace dx {
       //       This dummy initial value is provided, to prevent some spurious warnings from clang
       bool toRetry = false;
       // True if the request returns with a 503
-      bool serviceUnavailable = false;
+      // bool serviceUnavailable = false;
```
r-i-v-a, line 180: commented-out code

src/cpp/dxcpp/dxcpp.cc
windok: I've missed that HttpRequest by default initializes `responseCode` to -1. So, there should be no error when you access `responseCode` of not initialized `req`.

https://github.com/dnanexus/dx-toolkit/blob/master/src/cpp/SimpleHttpLib/SimpleHttp.h#L96

windok:
LGTM, only comments regarding commented code left

windok (approved):

PR #1465, opened 2025-03-31
COMPUTE-1113: Fix having indefinete quantity of retries for throttled requests when always_retry is set to false

windok:


r-i-v-a (approved):

wormsik:
what's the purpose of this change? my understanding of `always_retry` is that it allows you to force usually non-retried exception to be retried...not the other way around...

cc @kpjensen

windok:
Yes, `always_retry` should retry exceptions that are not retryable.

But now with infinite retries of throttling requests there is no possibility to end them. This update will allow to utilize `max_retry` option, so with default `max_retry` equals to 6, it still will be possible to make retries. But if you set `max_retry=0, always_retry=false` then it would mean you should not retry anything.

r-i-v-a (approved):

PR #1467, opened 2025-04-07
[APPS-2949] dx build --nextflow error - Attempting to clone from a RESTRICTED data container

jsitarova-dnanexus:


r-i-v-a:
please add changelog note

src/python/dxpy/nextflow/nextflow_utils.py
```diff
@@ -37,6 +38,23 @@ def get_template_dir():
     return path.join(path.dirname(dxpy.__file__), 'templating', 'templates', 'nextflow')
 
 
+def get_project_with_assets(region):
+    nextflow_basepath = path.join(path.dirname(dxpy.__file__), 'nextflow')
+
+    if 'stagingapi' in dxpy.APISERVER_HOST:
```
mhrvol, line 44: is it possible to use this on workers -- isn't there something like 10.0.0.8 or so in the dxpy.APISERVER_HOST?

r-i-v-a: @mhrvol good point, I think a different method is needed on workers

mhrvol: We can also check the existence of public project - not ideal, but should work in all cases.

r-i-v-a: Since this is dxpy and it can be running locally _or_ on a worker, do both checks.

mhrvol: Why do we need both checks? Existence of public project will work in both environments

jsitarova-dnanexus: Thanks @mhrvol, didnt know it has different format on workers, I rewrote it using public projects

r-i-v-a (approved):

PR #1478, opened 2025-06-19
feat: support multiple types in a nextflow schema

nh13:
Fixes: https://github.com/nf-core/fastquorum/issues/127

Unfortunately I cannot test locally with ` python -m pytest src/python/test -k test_type_with_multiple_values` since I get:

```console
dxpy.exceptions.PermissionDenied: Cannot create a new project as an anonymous user, code 401. Request Time=1750298027.817808, Request ID=1750298028284-940822

src/python/dxpy/__init__.py:670: PermissionDenied
```

### DIFF
--- a/src/python/dxpy/nextflow/nextflow_builder.py
+++ b/src/python/dxpy/nextflow/nextflow_builder.py
@@ -175,7 +175,7 @@ def prepare_custom_inputs(schema_file="./nextflow_schema.json"):
         elif nf_type in types:
             return types[nf_type]
         elif isinstance(nf_type, list):
-            return "array:string"  # default to string if we have a list of types
+            return "string"  # default to string if we have a list of types
         raise Exception("type {} is not supported by DNAnexus".format(nf_type))
 
     inputs = []


src/python/dxpy/nextflow/nextflow_builder.py
```diff
@@ -174,6 +174,8 @@ def get_dx_type(nf_type, nf_format=None):
             return str_types[nf_format]
         elif nf_type in types:
             return types[nf_type]
+        elif isinstance(nf_type, list):
+            return "string"  # default to string if we have a list of types
```
nh13, line 178:
```suggestion
            return "array:string"  # default to string if we have a list of types
```

src/python/dxpy/nextflow/nextflow_builder.py
```diff
@@ -174,6 +174,8 @@ def get_dx_type(nf_type, nf_format=None):
             return str_types[nf_format]
         elif nf_type in types:
             return types[nf_type]
+        elif isinstance(nf_type, list):
+            return "array:string"  # default to string if we have a list of types
```
nh13:
```suggestion
            return "string"  # default to string if we have a list of types
```

PR #1481, opened 2025-07-24
SCIPROD-2917: Bug fix: dx extract_assay expression accepting invalid json filters

gabriela-reznik:
This PR is intended for a bug fix regarding dx extract_assay expression, described on https://jira.internal.dnanexus.com/browse/SCIPROD-3192.

Currently some unexpected behavior has been found:

1. Providing a json filter with only sample_id (which is an invalid filter) fails with the following error message: "{'type': 'QueryBuilderError', 'message': 'Issue trying to craft a query', 'details': "Missing 'geno_bins' in assay filters", 'time': '07/24/2025, 10:16:38'}"

2. Providing a json with sample_id, annotation and location returns data even though filters with both annotation and location are not valid.

Passing E2E: https://github.com/dnanexus/pseudopod/actions/runs/16497641483

### DIFF
--- a/src/python/test/test_extract_expression.py
+++ b/src/python/test/test_extract_expression.py
@@ -794,7 +794,7 @@ class TestDXExtractExpression(unittest.TestCase):
             "sample_id_only",
             "When sample_id is present, one of the following keys must be also present: expression, annotation, location.",
         )
-    
+
     def test_sample_id_location_annotation(self):
         self.common_negative_filter_test(
             "sample_id_location_annotation",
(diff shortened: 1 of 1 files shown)


src/python/test/test_extract_expression.py
```diff
@@ -1425,6 +1438,7 @@ def test_merge_duplicate_filters_schema_1_1_non_optimized(self):
         assert merged_compound_block == expected_merged_compound_block
 
 
+
```
morsza: I suppose this is extra

morsza (approved):

PR #1489, opened 2025-09-03
[PTFM-40426] add new OrgExpired error definition

jcarnogursky-dna:
Adding new OrgExpired error to both python and java source codes.

This is part of new Wallet Lifecycle initiative.


src/python/dxpy/toolkit_version.py
```diff
@@ -1 +1 @@
-version = '0.399.0'
+version = '0.399.0+g92330d19f'
```
kpjensen:
```suggestion
version = '0.399.0'
```

PR #1490, opened 2025-09-04
[PTFM-39210] add support for whole file checksum

honzahovorka:


haroldna (commented):
the changes look good overall.

if you could add a test for each type of checksum, then I think this PR is good to go

PR #1549, opened 2026-04-14
APPS-3923: Default -profile docker for nf-core pipelines

jhuynh-cmd:
## Summary

nf-core pipelines define Docker support inside a `profiles` block, so without `-profile docker` being explicitly passed, Docker is never enabled. On DNAnexus workers, this causes cryptic "command not found" failures because tools aren't installed natively, only inside containers.

**Changes:**

- **`nextflow.sh`**: At job startup, writes `docker.enabled = true` to a temp config (`/tmp/dx_docker.config`) and injects it as the first `-c` flag passed to `nextflow run`. Soft configs supplied by the user are appended after, so they retain higher precedence and can still override docker settings. This makes Docker always active without requiring users to manually specify `-profile docker`.

### DIFF
--- a/src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
+++ b/src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
@@ -85,7 +85,7 @@ main() {
   # Injected at lowest precedence so user soft configs can still override docker settings.
   DX_DOCKER_CONFIG="/tmp/dx_docker.config"
   echo "docker.enabled = true" > "$DX_DOCKER_CONFIG"
-  RUNTIME_CONFIG_CMD="-c $DX_DOCKER_CONFIG "
+  RUNTIME_CONFIG_CMD="-c ${DX_DOCKER_CONFIG} "
   [[ -d "$HOME/in/nextflow_soft_confs/" ]] && RUNTIME_CONFIG_CMD+=$(find "$HOME"/in/nextflow_soft_confs -name "*.config" -type f -printf "-c %p ")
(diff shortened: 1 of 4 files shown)


src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -81,7 +81,12 @@ main() {
   dx-download-all-inputs --parallel @@EXCLUDE_INPUT_DOWNLOAD@@ 2>/dev/null 1>&2
   RUNTIME_CONFIG_CMD=''
   RUNTIME_PARAMS_FILE=''
-  [[ -d "$HOME/in/nextflow_soft_confs/" ]] && RUNTIME_CONFIG_CMD=$(find "$HOME"/in/nextflow_soft_confs -name "*.config" -type f -printf "-c %p ")
+  # Always enable Docker — DNAnexus workers only support Docker as a container runtime.
+  # Injected at lowest precedence so user soft configs can still override docker settings.
+  DX_DOCKER_CONFIG="/tmp/dx_docker.config"
+  echo "docker.enabled = true" > "$DX_DOCKER_CONFIG"
```
windok, line 87: docker profile in nextflow is actually shortcut for this config:

```
docker {
        docker.enabled          = true
        conda.enabled           = false
        singularity.enabled     = false
        podman.enabled          = false
        shifter.enabled         = false
        charliecloud.enabled    = false
        apptainer.enabled       = false
        docker.runOptions       = '-u $(id -u):$(id -g)'
    }
```
So, I think all these commands has to be added by default.

Also, it could make sense to increase qty of parallel subjobs from 4 to some bigger value with this change, if we already are changing default configuration:
```
executor.queueSize = 10
```

PR #1561, opened 2026-06-22
APPS-3832 dx ssh should use only host field

jhuynh-cmd:
### Summary
When running `dx ssh <job-id>`, the CLI resolves the target hostname to open the SSH connection to the job's worker. A job description exposes two relevant fields:

- **`host`** - the direct hostname of the actual worker running the job.
- **`httpsApp.dns.url`** - for HTTPS App jobs.

### Fix
Remove the `httpsApp.dns.url` override so `dx ssh` **always** uses the `host` field, regardless of whether the job is an HTTPS App.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ### Fixed
 
-* `dx ssh` now always connects using the job's `host` field instead of the HTTPS App ALB URL
+* `dx ssh` now always connects using the job's `host` field instead of the HTTPS App domain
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ### Fixed
 
+* `dx ssh` now always connects using the job's `host` field instead of the HTTPS App ALB URL
```
wormsik:
```suggestion
* `dx ssh` now always connects using the job's `host` field instead of the HTTPS App domain
```

jhuynh-cmd: I've updated the comment, thanks Jan 👍

wormsik (approved):
LGTM 🚀

PR #1566, opened 2026-07-17
TITAN-244: Prioritize md5 over per-part checksum on download

haroldna:
### Summary
On download, dxpy verified **both** the per-part `md5` and the per-part `checksum` whenever a symlink/drive file's part carried both. Per team decision with @XingGemini and @ksoni , md5 should be the single integrity check when present. This PR skips per-part `checksum` verification when a part has an `md5`; the checksum verification remains as a fallback for parts without an `md5`.

### Why
`migrateToSymlink`-created files (e.g. UKB migration) carry both `md5` and per-part `checksum`. On dxpy ≥ v0.402.0 (which added CRC64NVME verification), enforcing both causes redundant hashing and download failures (`CRC64NVME checksum not found` / checksum mismatch). This change is being done to unblock testing for the UKB symlink migration. 

### Change
- `src/python/dxpy/bindings/dxfile_functions.py` — gate the per-part `_verify_checksum` call in the main download loop on md5 absence:
  ```python
  if dxfile_desc.get('drive') is not None and "md5" not in parts[cur_part]:
  ```

### DIFF
--- a/src/python/dxpy/bindings/dxfile_functions.py
+++ b/src/python/dxpy/bindings/dxfile_functions.py
@@ -432,7 +432,7 @@ def _download_dxfile(dxid, filename, part_retry_counter,
                 if chunk_part != cur_part:
                     verify_part(cur_part, got_bytes, hasher, e_tag)
                     cur_part, got_bytes, hasher = chunk_part, 0, md5_hasher()
-                    if dxfile_desc.get('drive') is not None:
+                    if dxfile_desc.get('drive') is not None and "md5" not in parts[cur_part]:
                         _verify_checksum(parts, cur_part, chunk_data, checksum_type, dxfile.get_id())
                 got_bytes += len(chunk_data)
                 hasher.update(chunk_data)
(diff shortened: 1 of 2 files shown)


src/python/dxpy/bindings/dxfile_functions.py
```diff
@@ -432,7 +432,7 @@ def verify_part(_part_id, got_bytes, hasher, e_tag = None):
                 if chunk_part != cur_part:
                     verify_part(cur_part, got_bytes, hasher, e_tag)
                     cur_part, got_bytes, hasher = chunk_part, 0, md5_hasher()
-                    if dxfile_desc.get('drive') is not None:
+                    if dxfile_desc.get('drive') is not None and "md5" not in parts[cur_part]:
```
ksoni, line 435: This check looks good to me, but quick question, if md5 present it is skipping this checksum check, but is it performing the md5 checks? do we have some test around it?

ksoni (commented):
Looks good to me, just one comment/ question.

src/python/dxpy/bindings/dxfile_functions.py
haroldna: added `test_md5_verified_when_checksum_skipped` to explicitly cover this test case

kpjensen (approved):

emiloslavsky:
@haroldna @kpjensen - please update the changelog to reflect the user-facing impact of this change.   Also please make clear that this change applies to drive / external storage files only.

PR #1569, opened 2026-07-30
TITAN-272: Fix checksum mismatch on multi-chunk downloads

haroldna:
### Problem
Downloading a symlink/drive file larger than one download chunk (`DEFAULT_BUFFER_SIZE` = 16 MB off-platform, 96 MB in a job) fails with a checksum mismatch even though the downloaded data is fully intact:

dxpy.exceptions.DXChecksumMismatchError: CRC64NVME checksum mismatch in file-xxxx in part 1 (expected ..., got ...)

### Root cause
In `_download_dxfile`, a part is split into `chunksize` chunks by `chunk_requests()`, but the per-part checksum was verified against only the **first** chunk of each part. For any part larger than the chunk size the checksum was computed over partial data and always mismatched.

This is **not** CRC64NVME-specific. `checksumType` is a per-file property threaded through the same code path, so **every** supported type (CRC32, CRC32C, SHA1, SHA256, CRC64NVME) was affected. The md5 path was unaffected because it already accumulates across chunks via `hasher.update()`.

### Fix
- Added `_IncrementalChecksum` to accumulate a whole-part checksum across every chunk (CRC32/CRC32C/CRC64NVME via running value; SHA1/SHA256 via `hashlib`).
- Extracted `_compare_part_checksum` for the expected-vs-got comparison.
- The download loop now updates the accumulator on **every** chunk and verifies over the whole part at the part boundary, mirroring the existing md5 flow.
- `_verify_checksum` delegates to the new helpers, preserving its one-shot contract.

### Tests
- New `TestDownloadMultiChunkChecksum` drives the real chunk-splitting path over an in-memory buffer, **parametrized over all 5 checksum types**. `test_multi_chunk_part_downloads_ok` fails without the fix and passes with it for every type.
- Updated the checksum-gating tests to target the new verification entry point.
- All existing download / `_verify_checksum` unit tests pass (21 total).


emiloslavsky:
please update changelog to reflect the user-facing impact of this change

src/python/dxpy/bindings/dxfile_functions.py
wormsik, line 231: is this going to work? `_verify_checksum` called on the line 440 seems to send only last chunk --> so if you update only with chunk_data, you get only checksum of the last chunk

bluednanexus: @wormsik : I fixed resume-preflight to use full-part incremental accumulation, matching the main download loop behavior.

wormsik (approved):
LGTM 🚀

PR #1570, opened 2026-08-05
APPS-4204: Dx-toolkit sanity checks for nextaur json files

gabriela-reznik:


src/python/dxpy/nextflow/validate_nextflow_assets.py
```diff
@@ -0,0 +1,141 @@
...
+
+                prod_regions = validate_asset_file(
+                    prod_path, RECORD_ID_RE, "record", errors
+                )
+                staging_regions = validate_asset_file(
+                    staging_path, RECORD_ID_RE, "record", errors
+                )
+
+                # Region-set consistency between prod and staging
+                if prod_regions is not None and staging_regions is not None:
```
simunic-cz: This only checks regions parity between staging and production asset files of the same type. Is having a different regions in asset files of different types (`nextflow`, `nextaur`, `awscli`) a valid state?

src/python/dxpy/nextflow/validate_nextflow_assets.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # ------------------------------------------------------------------
+    # 3. Validate app_asset_projects_ids_{prod,staging}.json
+    # ------------------------------------------------------------------
+    for filename in (
+        "app_asset_projects_ids_prod.json",
+        "app_asset_projects_ids_staging.json",
+    ):
+        filepath = os.path.join(NEXTFLOW_DIR, filename)
+        validate_asset_file(filepath, PROJECT_ID_RE, "project", errors)
```
simunic-cz: Related to previous comment - is parity of regions between the records and projects not necessary?

src/python/dxpy/nextflow/validate_nextflow_assets.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # ------------------------------------------------------------------
+    # 4. Report
+    # ------------------------------------------------------------------
+    if errors:
+        for err in errors:
+            print(f"FAIL  {err}", file=sys.stderr)
+        return False
+
+    print(f"OK    All Nextflow asset JSON files passed validation.", file=sys.stderr)
```
simunic-cz: If this is a success message, should it be written to `strerr`?
```suggestion
    print(f"OK    All Nextflow asset JSON files passed validation.", file=sys.stdout)
```

.github/workflows/validate-nextflow-assets.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  validate-assets:
+    name: Validate asset record/project IDs
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
```
simunic-cz: Let's please use action versions pinned down to specific commit hash 🙏
```suggestion
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2

      - uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38 # v5.4.0
```

(Ideally even with the latest [approved versions](https://github.com/dnanexus/standards/blob/main/approved-github-actions/allowed-github-actions.yaml))
```suggestion
      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0

      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
```

src/python/dxpy/nextflow/validate_nextflow_assets.py
gabriela-reznik: Good catch. As far as I know and can observe from the code all assets should be present in a region for nextflow functionalities

src/python/dxpy/nextflow/validate_nextflow_assets.py
gabriela-reznik: in here we are just checking for assets project IDs but yes, they should have parity in stg and prod

simunic-cz (approved):

wormsik (approved):
LGTM 🚀

PR #1571, opened 2026-08-19
APPS-4127: Enable OFFLINE mode for Nextflow applet

bluednanexus:
Nextflow natively allows users to pass a NXF_OFFLINE flag such that the executable will not try to make any calls to the internet for extra info such as update, config, etc. Currently our Nextaur implementation does not have the ability to do this.

Scope:

- Ensure Nextaur support the above flag, and optionally automatically turn it on when in a restricted environment (e.g. a project where jobs are not allowed open internet access).

- Validate that the implementation work with simple nextflow pipeline with docker images already cached in the project. If possible, please test in staging OFH environment in a restricted project.

**Test**
See my test result here:
[TEST_RESULTS.md](https://github.com/user-attachments/files/31206741/TEST_RESULTS.md)


bomquangia:
Nice work overall — the approach is right, the tests are genuinely good (45/45 pass locally, shellcheck is clean), and the docs are honest. A few things to fix before merge.

### 1. Run options after a line break get thrown away — please fix this one

**What happens:** `parse_offline_run_opts` rebuilds the `nextflow_run_opts` string, but the `read` command it uses stops at the first line break. Everything after that line break disappears.

```
User passes:  -resume
              -profile docker
Nextflow gets: -resume
```

**Why it matters:** This happens on **every** run, not just offline ones, because the variable is rewritten unconditionally. The job doesn't fail — it just quietly runs with the wrong options. Before this PR the full string was still passed to Nextflow, so this is new.

**Fix:** Either use `local -a opts=($nextflow_run_opts)` (`set -f` is already on, so this is safe), or only write the variable back when a token was actually removed. The second option is safer — it leaves the string untouched on the common path.

`nextflow.sh:466`, `nextflow.sh:493`

### 2. `-offline` and `-offline=false` together give contradictory results

**What happens:** A bare `-offline` is always kept, but a later `-offline=false` switches the decision to "online" and returns early. So the log says one thing and Nextflow gets another:

```
-offline -offline=false          → log says "running online", but -offline is still on the command line
-offline -offline=false -latest  → log says "running online", then Nextflow aborts with
                                   "-latest and -offline cannot be specified at the same time"
```

**Why it matters:** The job log is the only way support can tell what mode a run was in, so it shouldn't be able to lie. Low probability input, one-line fix.

**Fix:** When the final decision is "off", also drop the bare `-offline` from the kept tokens.

`nextflow.sh:471-474`

### 3. Applets built with `--cache-docker` won't have this feature

`--cache-docker` always builds through the Nextflow Pipeline Importer app, which renders `nextflow.sh` with its own dxpy — so offline mode won't be there until NPI's dxpy is updated. Please file a follow-up JIRA ticket to update the NPI app.

### 4. The CHANGELOG promises automatic detection that has never actually run

The automatic path (`jobOutboundInternet=false`) has only ever been exercised by unit tests with a fake `dx` command. Every one of the 12 staging jobs reports `jobOutboundInternet=true`, so the real branch has never executed on the platform.

`Offline_Mode.md` §5.4 already commits to this: *"if that environment is not ready … the gap is stated in the release notes rather than being left implicit."* That hasn't happened. One sentence in the CHANGELOG closes it.

### Minor

- A quoted `"-offline=true"` isn't matched by the `case` patterns, so it reaches Nextflow and fails as an unknown option.
- Repeated spaces inside a quoted value get collapsed: `--with-report "a  b.html"` becomes `"a b.html"`.
- `-offline=0` and `-offline=no` turn offline **on**. Only `false` turns it off. Consider rejecting values you don't recognise instead of guessing.
- Putting `-offline` in `nextflow_top_level_opts` fails with `Unknown option` (it's a `run` option, not a top-level one). A short check with a "use nextflow_run_opts instead" hint would save a support ticket.
- `import subprocess` is inside `_run_bash` instead of at the top of the test file.

bomquangia (commented):
Withdrawn — please disregard this review. The review comment earlier in the thread stands.

bluednanexus:
> Nice work overall — the approach is right, the tests are genuinely good (45/45 pass locally, shellcheck is clean), and the docs are honest. A few things to fix before merge.
> 
> ### 1. Run options after a line break get thrown away — please fix this one
> **What happens:** `parse_offline_run_opts` rebuilds the `nextflow_run_opts` string, but the `read` command it uses stops at the first line break. Everything after that line break disappears.
> 
> ```
> User passes:  -resume
>               -profile docker
> Nextflow gets: -resume
> ```
> 
> **Why it matters:** This happens on **every** run, not just offline ones, because the variable is rewritten unconditionally. The job doesn't fail — it just quietly runs with the wrong options. Before this PR the full string was still passed to Nextflow, so this is new.
> 
> **Fix:** Either use `local -a opts=($nextflow_run_opts)` (`set -f` is already on, so this is safe), or only write the variable back when a token was actually removed. The second option is safer — it leaves the string untouched on the common path.
> 
> `nextflow.sh:466`, `nextflow.sh:493`
> 
> ### 2. `-offline` and `-offline=false` together give contradictory results
> **What happens:** A bare `-offline` is always kept, but a later `-offline=false` switches the decision to "online" and returns early. So the log says one thing and Nextflow gets another:
> 
> ```
> -offline -offline=false          → log says "running online", but -offline is still on the command line
> -offline -offline=false -latest  → log says "running online", then Nextflow aborts with
>                                    "-latest and -offline cannot be specified at the same time"
> ```
> 
> **Why it matters:** The job log is the only way support can tell what mode a run was in, so it shouldn't be able to lie. Low probability input, one-line fix.
> 
> **Fix:** When the final decision is "off", also drop the bare `-offline` from the kept tokens.
> 
> `nextflow.sh:471-474`
> 
> ### 3. Applets built with `--cache-docker` won't have this feature
> `--cache-docker` always builds through the Nextflow Pipeline Importer app, which renders `nextflow.sh` with its own dxpy — so offline mode won't be there until NPI's dxpy is updated. Please file a follow-up JIRA ticket to update the NPI app.
> 
> ### 4. The CHANGELOG promises automatic detection that has never actually run
> The automatic path (`jobOutboundInternet=false`) has only ever been exercised by unit tests with a fake `dx` command. Every one of the 12 staging jobs reports `jobOutboundInternet=true`, so the real branch has never executed on the platform.
> 
> `Offline_Mode.md` §5.4 already commits to this: _"if that environment is not ready … the gap is stated in the release notes rather than being left implicit."_ That hasn't happened. One sentence in the CHANGELOG closes it.
> 
> ### Minor
> * A quoted `"-offline=true"` isn't matched by the `case` patterns, so it reaches Nextflow and fails as an unknown option.
> * Repeated spaces inside a quoted value get collapsed: `--with-report "a  b.html"` becomes `"a b.html"`.
> * `-offline=0` and `-offline=no` turn offline **on**. Only `false` turns it off. Consider rejecting values you don't recognise instead of guessing.
> * Putting `-offline` in `nextflow_top_level_opts` fails with `Unknown option` (it's a `run` option, not a top-level one). A short check with a "use nextflow_run_opts instead" hint would save a support ticket.
> * `import subprocess` is inside `_run_bash` instead of at the top of the test file.

Resolved

bomquangia (approved):

bomquangia (approved):

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+# =========================================================
+# Helpers: offline mode
+# =========================================================
+
+# Echoes "false" if the platform reports that outbound internet access is disabled for
+# this job, "true" if it is enabled, and nothing if the information is unavailable.
+# `jobOutboundInternet` is excluded from the default describe output, so it has to be
+# requested explicitly; when the field is not set at all, the platform reports `true`.
+# Both the job and its project are checked, and a single `false` is decisive.
+get_job_outbound_internet() {
```
wormsik, line 498: can we plase unify both `get_job_outbound_internet` and `get_job_network_access` unify into a single API call? to not worsen the throttling situation...

wormsik: also isn't some of the information accessible in a file located in `/home/dnanexus/dnanexus-job.json`? so we could avoid API calls at all

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+# Offline mode is used when the user asks for it with the Nextflow `-offline` run option,
+# or when this job runs in a restricted environment without outbound internet access.
+# Every branch records $NXF_OFFLINE_REASON, so log_context_info can state why the run is
+# offline *or* why it is not -- support has to be able to answer that from the job log alone.
+setup_offline_mode() {
+  parse_offline_run_opts
+
+  # explicit opt-out wins over everything: a restricted environment, and a variable that
+  # was already in the job environment -- it is the only way to force a run back online
+  if [[ $NXF_OFFLINE_REQUEST == off ]]; then
```
wormsik: why do we use `on` and `off` if the original variable is `true` or `false`?

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+}
+
+# Offline mode is used when the user asks for it with the Nextflow `-offline` run option,
+# or when this job runs in a restricted environment without outbound internet access.
+# Every branch records $NXF_OFFLINE_REASON, so log_context_info can state why the run is
+# offline *or* why it is not -- support has to be able to answer that from the job log alone.
+setup_offline_mode() {
+  parse_offline_run_opts
+
+  # explicit opt-out wins over everything: a restricted environment, and a variable that
```
wormsik, line 619: when this can be useful? in restricted jobs, it won't have any effect and break everything...or is it possible to have it per-module or per-step - i.e. 90 % of pipeline run is network restricted, but I allow network for a single step

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+
+  # An executable that asks for no network access has no egress regardless of the project
+  if [[ $(get_job_network_access) == false ]]; then
+    enable_offline_mode "this job was granted no outbound network access (networkAccess is empty)"
+    return
+  fi
+
+  case $(get_job_outbound_internet) in
+  false)
+    enable_offline_mode "this job has no outbound internet access (jobOutboundInternet=false)"
```
wormsik:
```suggestion
    enable_offline_mode "this job has no outbound Internet access"
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+    ;;
+  true)
+    NXF_OFFLINE_REASON="this job has outbound internet access (jobOutboundInternet=true)"
```
wormsik:
```suggestion
    NXF_OFFLINE_REASON="this job has outbound Internet access"
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+    ;;
+  *)
+    NXF_OFFLINE_REASON="could not determine whether this job has outbound internet access"
```
wormsik:
```suggestion
    NXF_OFFLINE_REASON="could not determine whether this job has outbound Internet access"
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+  fi
+
+  if [[ $NXF_OFFLINE_REQUEST == on ]]; then
+    enable_offline_mode "requested with -offline in nextflow_run_opts"
+    return
+  fi
+
+  # An executable that asks for no network access has no egress regardless of the project
+  if [[ $(get_job_network_access) == false ]]; then
+    enable_offline_mode "this job was granted no outbound network access (networkAccess is empty)"
```
wormsik:
```suggestion
    enable_offline_mode "this job wasn't granted outbound Internet access (networkAccess is empty)"
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+  case $(get_job_outbound_internet) in
+  false)
+    enable_offline_mode "this job has no outbound internet access (jobOutboundInternet=false)"
+    ;;
+  true)
+    NXF_OFFLINE_REASON="this job has outbound internet access (jobOutboundInternet=true)"
+    ;;
+  *)
+    NXF_OFFLINE_REASON="could not determine whether this job has outbound internet access"
+    echo "Could not determine whether this job has outbound internet access; assuming it has."
```
wormsik:
```suggestion
    echo "Could not determine whether this job has outbound Internet access; assuming it has."
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+  # was already in the job environment -- it is the only way to force a run back online
+  if [[ $NXF_OFFLINE_REQUEST == off ]]; then
+    unset NXF_OFFLINE
+    NXF_OFFLINE_REASON="turned off with -offline=false in nextflow_run_opts"
+    echo "Offline mode turned off with -offline=false in nextflow_run_opts; running online."
+    return
+  fi
+
+  if [[ $NXF_OFFLINE == true ]]; then
+    enable_offline_mode "NXF_OFFLINE was already set in the job environment"
```
wormsik:
```suggestion
    enable_offline_mode "NXF_OFFLINE is set in the job environment"
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+}
+
+# Turns on Nextflow offline mode: no version check, no plugin registry access, no
+# remote pipeline update check. See https://nf-co.re/docs/running/run-pipelines-offline
+enable_offline_mode() {
+  NXF_OFFLINE_REASON="$1"
+  export NXF_OFFLINE=true
+  export NXF_DISABLE_CHECK_LATEST=true
+  drop_latest_run_opt
+  echo "Nextflow offline mode enabled ($NXF_OFFLINE_REASON). Nextflow will not contact the internet"
```
wormsik:
```suggestion
  echo "Nextflow offline mode enabled ($NXF_OFFLINE_REASON). Nextflow will not contact the Internet"
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+  echo "for version checks, plugin downloads or pipeline updates. Pipeline plugins that are not"
```
wormsik:
```suggestion
  echo "for version checks, Docker image downloads, plugin downloads or pipeline updates. Pipeline plugins that are not"
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,211 @@ setup_workdir() {
...
+  echo "bundled in the DNAnexus assets cannot be used, and docker images must be cached in the project."
```
wormsik:
```suggestion
  echo "bundled in the DNAnexus assets cannot be used, and Docker images must be cached in the project."
```

wormsik: can we point user to some public documentation about how to cache docker images?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Unreleased
 
+### Added
+
+* Nextflow applets support offline mode: `nextflow_run_opts="-offline"` exports `NXF_OFFLINE`, or it is enabled automatically when the job has no outbound internet access. `-offline=false` forces a run back online. The automatic path is covered by unit tests only — it has not yet been exercised on a restricted project.
```
wormsik:
```suggestion
* Nextflow applets support offline mode: `nextflow_run_opts="-offline"` exports `NXF_OFFLINE`, or it is enabled automatically when the job has no outbound internet access. `-offline=false` forces a run back online.
```

wormsik: do we want to mention this in Changelog?

wormsik (changes requested):

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
bluednanexus: Yes, we should unify both checks into a single job describe API call to reduce request count. We can additionally try dnanexus-job.json first, but keep API fallback for correctness. This gives best balance: fewer calls in the common path, no behavioral regression in edge cases.

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
bluednanexus: Good question. -offline=false only disables Nextflow offline mode; it does not grant internet access. In truly restricted jobs it may still fail, which is expected, but it is useful as an explicit override in partially restricted/allowlisted environments and for A/B troubleshooting. Per-step/per-module network policy is not supported here because network access is defined at the applet/job level, not per Nextflow process.

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
bluednanexus: Yes. Updated

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
bluednanexus: Updated!

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
bluednanexus: Updated

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,255 @@ setup_workdir() {
...
+      dx-jobutil-report-error "'${opt}' is a Nextflow run option: pass it in nextflow_run_opts, not nextflow_top_level_opts."
+      ;;
+    *) ;;
+    esac
+  done
+
+  # Left as `read` on purpose, unlike parse_offline_run_opts: `read` stops at the first line
+  # break, so options after one are not validated. That is a pre-existing gap and fixing it
+  # here would newly abort runs that pass a misplaced -w or -profile after a line break, which
+  # has nothing to do with offline mode. Worth its own ticket.
```
wormsik: can you please create a ticket and reference it there...so the comment is sth like:
```
# APPS-123 Left as `read` on purpose ....
```

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
```diff
@@ -425,13 +428,255 @@ setup_workdir() {
...
+        local project_outbound
+        project_outbound=$(jq -r 'if has("jobOutboundInternet") then (.jobOutboundInternet|tostring) else "" end' <<<"$project_json" 2>/dev/null || true)
+        if [[ $project_outbound == false ]]; then
+          JOB_OUTBOUND_INTERNET_RESULT=false
+        elif [[ -z $JOB_OUTBOUND_INTERNET_RESULT ]]; then
+          JOB_OUTBOUND_INTERNET_RESULT=$project_outbound
+        fi
+      fi
+    fi
+  fi
```
wormsik, line 490: is this needed at all? IMHO this field is always present when requested by `fields` param

CHANGELOG.md
```diff
@@ -6,7 +6,13 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Unreleased
 
+<<<<<<< HEAD
+### Added
+
+* Nextflow applets support offline mode: `nextflow_run_opts="-offline"` exports `NXF_OFFLINE`, or it is enabled automatically when the job has no outbound internet access. `-offline=false` forces a run back online.
+=======
 ## [414.0] - beta
+>>>>>>> master
```
wormsik: please fix this

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
bluednanexus: Agreed. I removed the has(...) guards and now parse the fields directly as booleans/array, while still failing open on unexpected/null values.

CHANGELOG.md
bluednanexus: @wormsik : fixed

src/python/dxpy/templating/templates/nextflow/src/nextflow.sh
bluednanexus: @wormsik : I created a new ticket : `APPS-4441` and reference it there

wormsik (commented):
LGTM 🚀 will give a final stamp once nextaur assets are updated as well

PR #1574, opened 2026-08-26
APPS-4308: Preserve request-scoped auth across DXFile operations

nanguyen-dnanexus:
## Problem

Several DXFile operations accept request-scoped kwargs such as `auth`, but not every internal request preserved them.

Two gaps were identified:

- `upload_local_file(..., auth=...)` preserved request kwargs through the upload lifecycle, but the upload-parts polling step could fall back to ambient authentication.
- `DXFile.read()` / `get_download_url()` performed project-existence checks without preserving the caller's request kwargs.

This can cause incorrect `PermissionDenied` failures when callers intentionally use request-scoped authentication instead of the global `SECURITY_CONTEXT`.

## Fix

Propagate request kwargs through the affected internal calls:

- forward non-conflicting request kwargs to upload-parts polling;
- forward request kwargs through `DXFile.read()` / `get_download_url()` to the project-existence check and its describe request.

`timeout` is intentionally excluded from upload polling because that method already uses it as its polling deadline.

No global `SECURITY_CONTEXT` mutation is introduced.

## Impact / Risk

The change is localized to preserving request kwargs already supplied by callers. Existing ambient-auth behavior remains unchanged when custom request authentication is not provided.

Related consumer: https://github.com/dnanexus/file-apps/pull/1287

### DIFF
--- a/src/python/dxpy/bindings/dxfile.py
+++ b/src/python/dxpy/bindings/dxfile.py
@@ -760,7 +760,8 @@ class DXFile(DXDataObject):
                 if project is None and 'DX_JOB_ID' not in os.environ:
                     project_from_handler = self.get_proj_id()
                     # object_exists_in_project will call /file-xxxx/describe, which is skipped if the URL is cached
-                    if project_from_handler and object_exists_in_project(self.get_id(), project_from_handler):
+                    if project_from_handler and object_exists_in_project(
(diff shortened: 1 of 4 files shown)


nanguyen-dnanexus:
Closing based on review feedback; handling the authentication context at the consumer/test layer instead

PR #1577, opened 2026-09-03
APOLLO-450: Fetch dx extract_dataset -ddd codings from the codings API

arpradhan:
## Summary

DML now stores dataset codings in each database's `dx_codings` table, and descriptors written since that change carry no `codings` block at all. `dx extract_dataset -ddd` only ever read codings from the descriptor, so it silently produced an empty `codings.csv` for any newly created dataset.

This switches `-ddd` to `POST /viz-meta/3.0/<record>/codings`, which merges `dx_codings` with descriptor codings server-side, so one code path serves both old and new datasets:

- Paginates the codings route via its keyset cursor (`next`/`starting`), reassembling any coding the server had to split across pages.
- For a split **hierarchical** coding, the server sends `code -> parent_code` edges (`codes_to_parent`) instead of a tree, specifically so the caller can rebuild the tree itself once every page is merged. Once pagination completes, `get_codings()` does exactly that — rebuilding a real `display` tree from the coding's complete edge map — so `display_order` is computed from it just like a whole coding's, with siblings ordered by code (the only order the wire format gives).
- For a split **flat** coding, there's no structure to rebuild from — the merged `display` list really is just code order — so `display_order` is left blank for that coding only, rather than fabricated.
- Falls back to descriptor codings, silently, on any API failure — including deployments that predate the route. An *empty* response is treated as authoritative (the dataset genuinely has no codings) and does not trigger the fallback.
- Keeps scope to codings referenced by at least one field, matching today's behavior and `dxdata.Dataset.codings`; `dx_codings` can hold codings no field references (e.g. a shared catalog ingested against a reduced dictionary), and those are intentionally excluded.
- Fixes a latent bug where the non-hierarchical branch re-derived its code list from `codes_to_concepts`, which could silently misalign code/meaning rows if that map differed in order or membership from `codes_to_meanings`.
- Caps pagination at `MAX_CODINGS_PAGES` so a server bug returning a non-advancing cursor raises instead of hanging `-ddd` forever, and copies each stored coding with `copy.deepcopy` rather than a hand-rolled partial copy.

## Pre-existing test failures fixed here

Getting a clean live E2E run required fixing four failures that were **already broken on `master`** and are unrelated to the codings change. They're included here so the suite is green end to end:

- **`test_extract_dataset` field CSVs (`1433151c4`, `a3c49f2e7`)** — the API renders datetimes with an explicit `+00:00` offset while the truth CSVs in `dx-toolkit_test_data` store them naive, so `patient.verified_dtm` diffed on representation rather than value. The `--fields` path writes API values verbatim through `csv_from_json`, so this was a stale fixture, not a client bug. The comparison now drops the redundant offset. Switching from a bare `assertTrue(df1.equals(df2))` to `assert_frame_equal` is what made the cause legible in the first place.
- **`test_extract_assay` rsid TSV (`e6def59b8`)** — expected output updated to match Python list quoting.
- **`test_create_cohort` SQL (`b7389ab8d`)** — expected SQL updated to match Spark's current quoting.
- **`DX_PROJECT_CONTEXT_ID` leakage between tests (`4244f4caa`)** — `dxpy.config` mirrors this into `os.environ`, so every `dx` subprocess started later in the same process inherits it. `TestCreateCohort` pointed it at a temp project and destroyed that project without restoring the previous value, and `TestDXTabCompletion` set it to `''` and *deleted* it in `tearDown` instead of restoring it (one of its tests also called `config.save()`, persisting the removal to `~/.dnanexus_config` for every later `dx` process). Whichever xdist worker ran them was left with an empty or dead project context, so later tests that resolve a path without an explicit project failed — e.g. the `dx cat` in `retrieve_bins` exiting 3 with `"" is not a recognized ID`. Both classes now save and restore the incoming context.

Because that last one only fired when a context-mutating test happened to precede an affected test on the same worker, it presented as flakiness rather than a consistent failure.

## Test plan

- [x] New offline unit tests in `src/python/test/test_extract_dataset_codings.py` (25 tests): pagination/cursor handling (including the real `after_code` cursor for a coding sliced mid-code), the hierarchical split-and-rebuild path across three pages, flat split-coding merge, the pagination cap, error/fallback handling, field-referenced scope filtering, response non-aliasing, and the `codes_to_concepts` misalignment regression — all pass.
- [x] Manual run of `dx extract_dataset -ddd` against a real dataset whose codings live only in `dx_codings` (no descriptor codings block) — previously produced an empty `codings.csv`, now produces the correct, non-empty output with the expected columns.
- [x] **Live E2E suite passed** on staging — [build 530](https://jenkins-vstg.internal.dnanexus.com/view/dx-toolkit%20release%20board/job/dx-toolkit/job/dxpy-master-integration-tests-on-staging-20.04-python3/530/). This covers `end_to_end_ddd`, which compares the generated `codings.csv` against its truth file, so the codings API path is exercised against a real deployment.
- [x] A dataset whose codings the server actually splits mid-coding (hierarchical and flat) has not been observed against a real deployment — the split/reassemble logic is covered by unit tests against canned responses only.

This PR was created with Claude Code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Unreleased
 
+### Changed
+
+* `dx extract_dataset -ddd` now retrieves dataset codings from the vizserver codings route, which serves codings stored in the database's `dx_codings` table, falling back to the codings in the dataset descriptor when the route is unavailable
+
 ## [414.0] - beta
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ### Changed
 
-* `dx extract_dataset -ddd` now retrieves dataset codings from the vizserver codings route, which serves codings stored in the database's `dx_codings` table, falling back to the codings in the dataset descriptor when the route is unavailable
+* `dx extract_dataset -ddd` now retrieves dataset codings from the database's dx_codings table, falling back to the codings in the dataset descriptor when the database codings are unavailable.
 
 ## [414.0] - beta
 
--- a/src/python/dxpy/cli/dataset_utilities.py
+++ b/src/python/dxpy/cli/dataset_utilities.py
@@ -219,8 +219,7 @@ def _hierarchy_from_parents(codes_to_parent, all_codes):
     Rebuilds a nested `display` tree for a hierarchical coding from its complete
     code -> parent_code edges, once every fragment of a coding split across pages has been
     merged. Siblings are ordered by code, since the wire format carries no other order for
-    a coding the server had to slice; see `_slice_page` in
-    hare/src/lib/api_handlers/utils/processor/codings_processor.py.
+    a coding the server had to slice.
     """
     children_of = collections.OrderedDict()
     for code in all_codes:
--- a/src/python/dxpy/cli/dataset_utilities.py
+++ b/src/python/dxpy/cli/dataset_utilities.py
@@ -190,8 +190,11 @@ def raw_api_call(resp, payload, sql_message=True):
     return resp_raw
 
 
-# Set on a coding assembled from more than one page. The server slices an oversized coding in
-# `code` order, so the authored display order of such a coding cannot be recovered.
+# Set on a coding assembled from more than one page. For a flat coding this is permanent:
+# its merged `display` is in `code` order, not the authored order, so display_order cannot
+# be recovered. For a hierarchical coding it is transient -- get_codings() clears it once it
+# rebuilds a real display tree from the coding's fully-merged code -> parent_code edges,
+# since display_order computed from that tree is exactly as legitimate as a whole coding's.
 CODING_PAGINATED_KEY = "_dx_paginated"
 
 # A backstop against a server bug that returns a `next` cursor that never resolves to null,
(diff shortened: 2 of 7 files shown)


src/python/dxpy/cli/dataset_utilities.py
```diff
@@ -188,6 +190,74 @@ def raw_api_call(resp, payload, sql_message=True):
...
+                existing.setdefault("codes_to_parent", {}).update(
+                    coding["codes_to_parent"] or {}
+                )
+                existing.pop("display", None)
+            elif "display" in coding and "display" in existing:
+                existing["display"].extend(coding["display"] or [])
+
+        if not page.get("next"):
+            return codings
+        payload["starting"] = page["next"]
```
austinlocke, line 302: If a page splits in the middle of a code in the case that the individual coding has more codes than the limit, then it will also return a "afterCode" value that the caller should pass back so the cursor can continue in the correct spot of the codings.

austinlocke, line 302: For heirachical codings that get split we also don't return display but instead return `codes_to_parent` so the caller can compute the display themselves after aggregating all the pages. I would recommend looking at `_slice_page` in `hare/src/lib/api_handlers/utils/processor/codings_processor.py` to fully understand the logic there.

arpradhan: I believe afterCode is already passed to the cursor because `page["next"]` always carries a `{"name": ..., "after_code": ...} ` object? 

I thought `page["next"]` always has this structure, whether or not a page splits in the middle of a code.

austinlocke: > I believe afterCode is already passed to the cursor because `page["next"]` always carries a `{"name": ..., "after_code": ...} ` object?
> 
> I thought `page["next"]` always has this structure, whether or not a page splits in the middle of a code.

Oh you're right, that makes sense. I didn't think about it pulling the whole object in.

arpradhan:
## Verified on staging: a coding split mid-code by the server

This closes the open test-plan item for observing a real split against a deployment (previously only covered by unit tests against canned responses).

Created a dataset on staging with a Data Model Loader build that includes APOLLO-411/412, so codings live in the database's `dx_codings` table and **not** in the descriptor — which is the case this PR fixes. Four codings, chosen to cover every combination of {flat, hierarchical} × {whole, split}:

| coding | codes | kind | server behaviour |
|---|---|---|---|
| `icd10_split` | 132626 | hierarchical | **sliced**, sends `codes_to_parent` |
| `lab_code_split` | 100001 | flat | **sliced**, sends partial `display` |
| `icd10_whole` | 5 | hierarchical | whole, nested `display` tree |
| `sex` | 3 | flat | whole, flat `display` list |

hare slices a coding only when that *single* coding exceeds `CODINGS_PAGE_LIMIT` (default 100000), and `-ddd` sends no `limit` override, so the two split codings had to genuinely pass 100k codes. `icd10_split` reuses hare's own oversized-coding fixture shape (`src/tests/conftest.py::icd10_style_rows`): chapter → block → category → subcategory, every `parent_code` itself a code.

### Results

`dx extract_dataset <record> -ddd` produced 232635 data rows, one per `(coding_name, code)`, and:

- **`icd10_split`** — `parent_code` populated, 26 roots (one per A–Z chapter), `display_order` 1..132626. This is the path added in 57695a7cc: the display tree is rebuilt from the coding's fully-merged `codes_to_parent` edges, so a split hierarchical coding gets a real display order rather than a blank column.
- **`lab_code_split`** — `display_order` blank on all 100001 rows, which is the intended outcome: the slice query orders by `code`, so any number here would be invented rather than authored.
- **`icd10_whole` / `sex`** — arrive whole and keep their authored order. `icd10_whole` deliberately orders `N01` after `N02`, which is *not* code order, so this check would fail if anything in the whole-coding path fell back to code sorting.

### One bug this found

Rehearsing hare's pagination locally against these inputs (reimplementing `process`/`_slice_page`/`_whole_codings` and feeding the pages to the real `get_codings()`) surfaced a defect the unit tests missed: a split coding contributes no `display_order` column at all, so `pd.concat` filled it with NaN and upcast the whole column to float — writing every order as `1.0` instead of `1`. That changes a column consumers parse as an integer, and would also break the e2e truth-file comparison for any dataset containing a split coding. Fixed in d1b0442f5 with a nullable `Int64` dtype plus a regression test.

### Note for anyone reproducing this

The DML dictionary CSV must **not** contain an `is_hierarchical` column. It is not a DXDM input — a field is hierarchical iff its coding has `parent_code` values. Any unrecognised dictionary column is pass-through `setattr`'d onto the `Field`, and `is_hierarchical` is a read-only property, so including it fails the job with `property 'is_hierarchical' of 'Field' object has no setter`. Also don't use `skip_ingestion=true`: that path calls `create_dataset()` without `stage_data()`, so `sync_codings()` never runs and the dataset ends up with no codings at all.

### Still outstanding

The live e2e suite (`test/test_extract_dataset.py -k ddd`) has not been run — it needs an environment with access to the `dx-toolkit_test_data` project.

austinlocke (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ Categories for each release: Added, Changed, Deprecated, Removed, Fixed, Securit
 
 ## Unreleased
 
+### Changed
+
+* `dx extract_dataset -ddd` now retrieves dataset codings from the vizserver codings route, which serves codings stored in the database's `dx_codings` table, falling back to the codings in the dataset descriptor when the route is unavailable
```
wormsik: can you please rephrase a bit? since this is customer facing, it should ideally contain only relevant things...e.g.

* `dx extract_dataset -ddd` now retrieves dataset codings from the database's `dx_codings` table, falling back to the codings in the dataset descriptor when the database codings are unavailable

wormsik (commented):
just a small changelog suggestion, otherwise LGTM

wormsik (approved):
LGTM 🚀

PR #1578, opened 2026-09-14
APPS-4127: Enable OFFLINE mode for Nextflow applet

bluednanexus:
- Release version 1.17.0 with feature Offline Mode for Nextflow
- Fix Nextflow Offline Mode Test Failures in Jenkins (cache state + JSON parsing fallback): 

1. Fix cache issue

- Fixed offline-mode evaluation to avoid subshell side effects when reading policy helpers.
- Preserved shared cache/state across checks so policy decisions are consistent within one run.
- Eliminated redundant API calls and stabilized call-order behavior in tests.

2. Fix variables and add Jenkins-safe fallback
- Standardized offline reason strings so tests can assert explicit jobOutboundInternet=true/false outcomes.
- Added resilient JSON parsing fallback in offline helpers when jq is unavailable or unreliable on Jenkins runners.
- Switched policy field extraction (networkAccess, jobOutboundInternet, project) to use fallback-aware helpers instead of relying only on jq.


wormsik (approved):
LGTM 🚀

wormsik:
Supersedes #1571

PR #1580, opened 2026-09-14
TITAN-310 When creating an app the project should always be internal storage

RealP:
## Problem

Building an app fails when the billing entity's default storage is set to a drive. `dx build --create-app` creates a temporary project to build in, and because `/project/new` was called without a `drive` field, the new project inherited the billTo's default storage rather than using DNAnexus storage.

## Change
Temporary build projects now pass `"drive": null` to `/project/new`, which keeps app build artifacts on DNAnexus storage regardless of the billTo's default storage setting.


wormsik (approved):
LGTM 🚀

